PR #3, opened 2015-07-09
Robotaxi

Robotaxi:
Hi guys,
I present a patch for #907 (Driving instructions in odd numbers of feet when using imperial measurements). This is a complete rewrite of the generation of miles/feet and kilometer/meter strings for the length announcements including consistent formatting of float numbers.
Have fun!

Robotaxi

### DIFF
--- a/navit/navigation.c
+++ b/navit/navigation.c
@@ -193,6 +193,9 @@ struct navigation {
 	int flags;
 };
 
+/** @brief Set of simplified distance values that are easy to be pronounced.
+*          Used for the 'vocabulary_distances' configuration.
+*/
 int distances[]={1,2,3,4,5,10,25,50,75,100,150,200,250,300,400,500,750,-1};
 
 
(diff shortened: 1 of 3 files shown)


pgrandin:
Thanks Robotaxi.
I cannot merge your PR directly into master as it would break our current workflow, but I tested it and merged it into SVN and it will make its way to master in the next hour.

I did not see any regressions on the metric announces. One possible improvement would be to use factional values for some announcements, example :
"Turn right in 0.3 miles into South Lake Shore Drive US 41" would usually be announced as "Turn right in a quarter of a mile into South Lake Shore Drive US 41"

Thanks Robotaxi!

pgrandin:
Merged into 69b41cc757e6af96bb20d2c7fe54655d9d25cd28

Robotaxi:
Set #907 to 'fixed'.
Thanks for your effort, pgrandin!

PR #5, opened 2015-07-16
Mostly UI Related

greg42:
These are my recent changes to the UI. I've been working on the key highlighting code and also added basic support for scrolling through the menus using rotary encoders. Comments and feedback are always welcome.


sleske:
Thanks for your fixes. I'll have a look to get them merged.

sleske:
Sorry, took me a while to look over the commits. The basic ideas look
good, however there are a number of issues.

First, some general remarks:

* Your pull request mixes several unrelated commits - that is more
  difficult to review (and makes it less likely the request can be
  pulled as-is). Please try to make pull-requests only about a single
  feature (as reasonably practical).
* The commit comments should make it clear what the commit does, even
  without reading the commit. Commits like "Fix some highlighting
  issues" are not ideal (could be worse, though).
  Also, if a commit is just a fix for an earlier commit (like "Fix
  windows builds"), just combine it with the commit it fixes.
* Finally, watch out for correct indendation.

About the individual commits (I omitted some commits that have already
been cherry-picked individually):

**Prevent back/prev buttons from staying highlighted**
What does this do? I could not find any change in behavior; for me
back/prev buttons do not stay highlighted even without the commit.
Could you provide a testcase?

**Instead of highlighting possible keys for completing a word, disable all other keys**
I'm a bit torn on the idea of just hiding keys that do not make sense
currently. On the one hand, this makes it clear that the key is
unavailable; on the other hand, hiding part of the UI always risks
confusing the user. I'd probably rather gray out the keys - that is 
the usual way to indicate something is disabled.

In addition to that, we currently have a number of bugs when deciding on
"possible" keys. In particular:
 - The search is case-insensitive and "accent insensitive" (i.e. it
   finds results with é if you enter e, or with ü if you enter u). The
   "possible keys" do not follow this behaviour, so if for example you
   enter "Mul", the search will find "Mülheim", but the "h" key will be
   wrongly hidden. Same problem if you search for "mül".
 - If you delete a letter, you will typically get more available keys
   (as you get more results). However, these keys only "come back" when
   the search has finished, which can take several seconds. This makes it
   difficult to correct a typo, especially if you must go back to one or
   two characters. 

I don't think the "hiding" feature makes sense before these bugs are
resolved.
I'll probably open a bug report for these problems, and attach your
patch there, to merge when the bugs are resolved.

**Support for scrolling through menus using rotary encoders**
The code looks ok. However, what is the purpose of this feature? If I
understand correctly, it allows operating the menus with just
PgUp/PgDown. What use is that? And what is a "rotary encoder"? Googling
only showed me some electronic switches.

**Fix style issues**
Ok

**Fix some highlighting issues**
If I understand correctly, this will switch the onscreen keyboard back to upper-case if all search input is deleted. However, this only works  using the backspace key on the physical keyboard, not with the onscreen  keyboard.
 

Could you address these issues (and rebase your pull request on trunk)?

pohlinkzei:
Rotary encoders are some electrical devices. If you use navit without a touch screen in an embedded environment, you might want to use rotary encoders. It only can switch left, right (and sometimes return/enter). So you need a way to access all menu items with the same button. Here with pgup/pgdown.

sleske:
@pohlinkzei : Thanks for the information. So basically the patch would allow people to operate navit on a system where mainly/only a rotary encoder is available for input? And for this to work, the rotary encoder would need to be set up such that it generates PgUp/PgDown key events?

pohlinkzei:
Yes, thats the main use case. It could probably be other keys as well. But in my opinon pgup / pgdown seem suitable for this purpose. In this case it is needed to convert the rotary signals to keypresses outside navit. But that makes testing easier because you wouldn't need the hardware. The highlighting part is strongly related to this rotary code because if you select the keys on the internal keyboard (while search) with the rotary or arrow-keys, the highlighting of possible keys and the highlightig of the currently selected key will mess up and confuse the user.

greg42:
> What does this do? I could not find any change in behavior; for me
> back/prev buttons do not stay highlighted even without the commit.
> Could you provide a testcase?

Sure. Start navit, and perform the following actions
* From the main menu select "Actions" -> "Town"
* Select Germany as country in case it's not your default
* Select Munich as city (any other city with a large number of streets will also do)
* When navit asks for a street name, enter one character (e.g., "a"). This will make navit generate a long list of possible streets so that the "Prev" and "Next" buttons will be needed to navigate in that list.
* Close the on-screen keyboard by hitting the triangle button in the bottom row
* Wait until the list of streets is generated completely (might take a while, but during the list is updated, the buttons on the screen will be re-drawn as well so that the bug won't be visible)
* Using *only* the arrow keys on your keyboard, now select the "Next" button. Then try to select any other element on the screen, again only using your keyboard. The "Next" button should stay highlighted.

> on the other hand, hiding part of the UI always risks
confusing the user

I understand this concern. However, if the user only has a rotary encoder to interact with navit, a large number of available buttons on the screen makes entering street names quite a tedious task. That was my reason to just hide the keys that wouldn't make sense anyway. If you believe that from a UI perspective graying them out is the better approach, I won't argue with you. Otherwise, one could also consider making the behavior optional (e.g., as a configuration setting or so).

> I don't think the "hiding" feature makes sense before these bugs are
> resolved.
> I'll probably open a bug report for these problems, 
> and attach your patch there, to merge when the bugs are resolved.

Yeah, why not.

sleske:
Ok, I now have the commits cleaned up and rebased on top of trunk. See branch https://github.com/navit-gps/navit/compare/trunk...pr-5-cleanedup for the result. I squashed and reordered some commits, and expanded the commit messages. There should be no code changes, apart from some renaming. I have kept you as the author (it's your code after all).

Please have a look at the changes. If you have no objections, I will merge all commits but the last (hiding keys during search). For the last commit, I will open a trac ticket, as discussed above.
Sound fair?

greg42:
Hey, cool. Makes me (and I believe @pohlinkzei as well) happy to have it in the official version :) The code looks good to me. Thanks a lot for cleaning this up and for keeping me as the author. And sorry for the mess - I'll (try to) provide cleaner patches next time ;)

sleske:
Thanks for your feedback. One more (hopefully last) thing.
I noticed that gui_internal_highlight_possible_keys has a small bug: Go to Town search, start a search so the Prev/Next buttons appear, then hide the keyboard (triangle button). Now keep pressing PgUp to go through all fields backwards. The highlighting will skip the result list and the triangle button. With PgDown this works correctly.
After I saw this, I tried to understand and fix gui_internal_highlight_possible_keys. Unfortunately, I did not manage to understand it completely. So I wrote a replacement, which both seems simpler (to me) and works correctly. I (force-)pushed it to branch pr-5-cleanup.

I don't like fixing code I don't understand - so could you have a look whether I missed some special case in my fix? Could you explain why the special cases in your  gui_internal_highlight_possible_keys are necessary? Thanks!

If you don't see any remaining problems, I'll merge in a few days.

sleske:
Okay, I just merged most of the changes with commit 8bdcd53d6 and preceding. For the highlighting code I opened [Trac 1339](http://trac.navit-project.org/ticket/1339), and pushed the code to branch pr-5-hide-keys (commit 1388a93 ).
I'm closing this pull request, as all the code has been taken care of :-). Thanks again for the contribution!

sleske:
I also documented the use of a rotary encoder in the Wiki: http://wiki.navit-project.org/index.php/Internal_GUI#Using_the_Internal_GUI . Feel free to edit or expand - I don't know much about rotary encoders.

PR #7, opened 2015-07-21
Try to read all available gps data in order to prevent lagging behind

greg42:
After driving for a while, navit started to lag behind several minutes on my system. This patch addresses the issue by skipping all but the most recent information received from gpsd.

### DIFF
--- a/navit/vehicle/gpsd/vehicle_gpsd.c
+++ b/navit/vehicle/gpsd/vehicle_gpsd.c
@@ -310,11 +310,17 @@ vehicle_gpsd_close(struct vehicle_priv *priv)
 static void
 vehicle_gpsd_io(struct vehicle_priv *priv)
 {
+	int result = -1;
+
 	dbg(lvl_debug, "enter\n");
 	if (priv->gps) {
 	 	vehicle_last = priv;
 #if GPSD_API_MAJOR_VERSION >= 5
-                if(gps_read(priv->gps)==-1) {
+		/* Try to catch up, in case we're lagging behind. */
+	 	while (gps_waiting (priv->gps, 500)) {
+	 		result = gps_read(priv->gps);
+	 	}
+                if(result == -1) {
                   dbg(lvl_error,"gps_poll failed\n");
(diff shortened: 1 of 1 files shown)


sleske:
Thank you for alerting us to the problem, and for the pull request. I'll have a look, and hope to commit/merge shortly.

sleske:
Unfortunately, I have been unable to replicate your problem. Looking at Navit's code, I also do not understand how this problem can arise.  Navit reads from gpsd by calling gps_read in vehicle_gpsd_io. vehicle_gpsd_io in turn is called by the event system whenever the gpsd socket becomes ready. So if messages accumulate on the socket,  vehicle_gpsd_io is called more quickly until the backlog is processed.

I tested this on Linux (with gtk_drawing_area graphics) using gpsfake, letting it send its test data very quickly (-c 0.03 ), and using ioctl(gps->gps_fd, FIONREAD, &count) to check the size of the backlog (data waiting on the socket). I saw the backlog increase when Navit was busy (zooming the map, calculating a route), but the backlog quickly went away afterwards.

So I suspect you are seeing some other problem. I'd like to have a reproducible problem before changing any code.

Therefore I am closing this pull request. Could you please file a bug in Trac ( http://trac.navit-project.org/ ) with the details of the problem you are seeing? Then we can hopefully figure out what's going wrong.
Thanks!

greg42:
Thanks for your investigation. The problem occurred on a raspberry pi running Navit, while doing other stuff in the background. When the problem occurred, the system was under a permanent high load. From what I can tell (please correct me if I'm wrong), Navit is inherently single-threaded. Therefore, I believe `vehicle_gpsd_io` is called when the gpsd socket becomes ready *and* Navit is idle. Do you think it make make sense to repeat your test while stressing Navit a bit (say, by repeatedly zooming the map via the dbus interface or so)?

sleske:
I opened http://trac.navit-project.org/ticket/1332 for this problem, and hope to have a patch ready soon. Let's continue discussion there. You can login using OpenID or your login for wiki.navit-project.org.

PR #18, opened 2015-08-14
Concatenation of announcements now working

Robotaxi:
Up to now the concatenation of turn announcements isn't implemented correctly.

Currently two announcements are only connected if the announcements are such close in a timely manner that they tend to get overlapped. In practice this hardly ever occurs.
In my view this is not the idea of concatenating turns. They shall be concatenated if the turns are close in a distance manner.

Hence, for me the criteria shall be like this:
- If two turns have a distance to each other that is less than the "level 2"-distance configured for the specific route item, (which stands for the "soon"-announcement), they shall be concatenated.
- The "soon" announcement for the second turn shall be suppressed in that case.
- Any announcement for the second turn from that concatenation down to "level 1" shall be suppressed, if (and only if) it would come too close to the preceding one.

Here is an example:
1.: "turn right soon"
2.: "in 250 meters turn right into the Riedlstraße, then turn left into the Marxstraße"
3.: "turn right into the Riedlstraße now, then turn left into the Marxstraße"
4.: ("turn left into the Marxstraße soon" will be suppressed, as well as "follow the road for the next xx meters")
5.: "in 250 meters turn left into the Marxstraße"  -> This may be suppressed, if the announcement no. 6 would come a few meters later. Navit would interrupt itself otherwise.
6.: "turn left into the Marxstraße now"

In such situations, in which many turns come close after the other this behavior sounds much less mechanic and for me it saves a bunch of nerves of the users because of annoying announcements.
Have fun!

### DIFF
--- a/navit/navigation.c
+++ b/navit/navigation.c
@@ -127,6 +127,18 @@ static int sharp_turn_limit = 110;
  * Note that, depending on other conditions, even maneuvers whose delta exceeds the threshold may still be announced as (sharp) turns. */
 static int u_turn_limit = 165;
 
+/** Enum that defines the different states in announcing a turn.
+ *  These states are dependent on the current distance to the turn;
+ *  the distances are configured in navit.xml for every type of highway. */
+enum announcement_level {
+	level_connect = -2,
+	level_error   = -1,
+	level_now     = 0,
+	level_meters  = 1,
+	level_soon    = 2,
+	level_follow  = 3
+};
+
 enum gender {unknown, masculine, feminine, neuter};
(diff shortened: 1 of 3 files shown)


pgrandin:
Robotaxi, I'd be really happy to merge this PR, but can you please look at solving the conflict?

Thanks!

pgrandin:
Thanks Robotaxi!
Just a heads-up for the next time : let's avoid merging translation directly in the trunk as it will mess up our launchpad workflow.

Great job otherwise!

PR #19, opened 2015-08-24
Circle.yml update

jandegr:
just some tweak to get the circleCI builds going again


pgrandin:
Thanks Jan. I'm concerned that removing libgtk will disable gui_internal, but let's see.

jandegr:
in a testbuild I get this from cmake in circle after the changes

--->>> gui
Enabled   gtk ( GTK libs found )
Enabled   internal ( Default )
Disabled  qml ( Qt Declarative not found )

PR #20, opened 2015-08-27
Support for reloading CSV maps

greg42:
Navit contains some speed camera warning code. In order to identify the locations of speed cameras, the mapset is queried. The [Wiki suggests](http://wiki.navit-project.org/index.php/OSD#speed_cam) to make use of CSV files in the mapset in order to provide navit with the required information. This approach works well as long as the data in the CSV file doesn't change. If however the speed camera information is updated while navit is running, there was no way to reload the CSV file. This pull request addresses that issue by adding a function `map_reload()`, which is exposed via DBUS.

### DIFF
--- a/navit/binding/dbus/binding_dbus.c
+++ b/navit/binding/dbus/binding_dbus.c
@@ -1934,6 +1934,18 @@ request_vehicleprofile_attr_iter_destroy(DBusConnection *connection, DBusMessage
 	return request_attr_iter_destroy(connection, message, "vehicleprofile", (void (*)(struct attr_iter *))vehicleprofile_attr_iter_destroy);
 }
 
+static DBusHandlerResult
+request_map_reload(DBusConnection *connection, DBusMessage *message)
+{
+	struct map *map;
+	map=object_get_from_message(message, "map");
+	if (! map)
+		return dbus_error_invalid_object_path(connection, message);
+
+	map_reload(map);
+	return empty_reply(connection, message);
+}
+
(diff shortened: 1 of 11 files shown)


mdankov:
greg42, thank you for your contribution.

As we discussed on the irc channel, I prefer not to change map API for this feature, to keep its simplicity and stability.

Unless there's a reason to update the map file but keep navit running with an older snapshot, I think it would be best to monitor the file for a change, and reload it in that case.

On-demand reloads (if you have reasons) could be done on switching the map active attribute, or on setting "data" attribute to the same (or a different) value.

I have not verified your code for the following case yet, but please note that you should not drop all the map contents immediately. At the time you're doing reload, there might be some suspended activity on the map. Some component may have allocated a map_rect, fetched a few items and gone idle. It should not crash on an attempt to get currently fetched item attributes, or fetch another item, or just close the map_rect. It's acceptable after map reload to return no more items for the map_rect created before map reload. IIRC, current map_csv map_rec implementation survives situation when any or all map items are deleted (by setting their type to type_none), and/or when new items are inserted. I think it's possible to maintain that behaviour on map reload.

pgrandin:
@greg42 would you like to provide an updated PR to address @mdankov's concerns?

greg42:
Sorry for the late response. Right now I don't think I have the time to properly address @mdankov's concerns. I've found another way for announcing POIs while driving and therefore I believe this can be closed. Sorry for bugging you.

PR #21, opened 2015-08-28
Importing 215433680/pt.po from launchpad

pgrandin:


pgrandin:
This was a test. Closing.

PR #35, opened 2015-11-08
Merge:Status osd.trac1324

mvglasow:
Adds a new OSD for routing engine status (busy, routing, no route, waiting for position) and hides OSDs which are relevant only for routing when no route is active


mvglasow:
Note that the implementation currently suffers from the issue described at http://trac.navit-project.org/ticket/1334#ticket. This is a separate issue, affecting multiple (if not all) OSD types, and the effects are minor: in the configuration supplied, the status OSD will not show until the route status changes for the first time. Since the OSD is hidden anyway when no destination is set, this will have an impact only when starting Navit with a destination already set: in this case, the OSD will go from hidden to "calculating" rather than from "waiting for destination" to "calculating".

Thus I don't see the issue as an impediment to merging this one, as the above is not a regression.

mvglasow:
The issue mentioned above is currently being fixed (tests were successful, just waiting for CircleCI to finish before I merge the fix).

mvglasow:
I'm closing this pull request as it got polluted with a bunch of commits which are not really related to this functionality (they fix a bug common to all OSDs and have since been cherry-picked into trunk). As a replacement I've opened #40.

PR #38, opened 2015-11-12
Fix:gui_internal:Format coordinates properly on Android Lollipop

mvglasow:

### DIFF
--- a/navit/gui/internal/gui_internal_command.c
+++ b/navit/gui/internal/gui_internal_command.c
@@ -39,6 +39,15 @@
 
 extern char *version;
 
+/**
+ * @brief Converts a WGS84 coordinate pair to its string representation.
+ *
+ * This function takes a coordinate pair with latitude and longitude in degrees and converts them to a
+ * string of the form {@code 45°28'0" N 9°11'26" E}.
+ *
+ * @param gc A WGS84 coordinate pair
+ * @param sep The separator character to insert between latitude and longitude
+ */
 static char *
 coordinates_geo(const struct coord_geo *gc, char sep)
 {
@@ -46,6 +55,7 @@ coordinates_geo(const struct coord_geo *gc, char sep)
 	int lat_deg,lat_min,lat_sec;
 	int lng_deg,lng_min,lng_sec;
 	struct coord_geo g=*gc;
+	char *ret;
 
 	if (g.lat < 0) {
 		g.lat=-g.lat;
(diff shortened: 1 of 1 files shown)


mdankov:
g_strjoin does not free strings passed to it. But it's required to free the memory allocated by g_strdup_printf and g_strnfill.

I think the current (memory leaking) state is better than it was before (values being cut) but we anyway should fix the leak asap.

I'm quite surprised with g_strdup_printf' inability to proceed with degree char in the format string but I guess following code should work well.
```C
return g_strdup_printf("%d%s%d'%d\" %c%c%d%s %d'%d\" %c",lat_deg,"°",lat_min,lat_sec,latc,sep,lng_deg,"°",lng_min,lng_sec,lngc);
```

mvglasow:
Passing the degree sign as an argument was the first thing I tried (not sure if I did that as a string or char argument). In log output it would cause mojibake to appear instead of the degree sign, and the GUI would crash on rendering the string. I've tried a few ways of persuading `g_strdup_printf()` to render the degree sign properly.

As for the memory leak, thanks for the heads up! I'll take care of it.

mdankov:
We can't store a generic Unicode code point inside the 8-bit char. So %c won't work.
All strings we work with in C code are stored in UTF-8, and all 7-bit ASCII characters are encoded exactly the same way in UTF-8 and in ASCII. All other code points are encoded as 2-4 byte sequences, and no 0x0 bytes are used in this encoding, except for 0x0 character itself.

So I still do not see why our code did fail on Lollipop, but I was able to reproduce it on the emulator, and it actually terminates the string not only on degree char but also on Cyrillic letters.

I had success passing degree symbol as a string argument, tested it on 5.1.1. emulator and have committed my change: https://github.com/navit-gps/navit/commit/3dcf4ba91e33a327d920cedfe29455569c3de747

It looks like Lollipop has changed behaviour of vsnprintf(), and it it's used by both g_strdup_printf() and different gettext macros.

So now I have tested Russian navigational instructions on Lollipop emulator. **They are broken too,** and there seems to be no such a simple way to solve them.

Will open a ticket for the issue.

mvglasow:
Obviously %c won't work for non-Latin characters, but the degree sign is 0xb0, thus it should fit in an 8-bit char. Anyways, passing it as a string is probably the better solution.

As far as I can tell, any character beyond 0x7f in the format string will truncate the output. I haven't tested the impact on translations (as I use Navit in English), but I fear it's going to be pretty bad. Non-Latin result strings are likely going to be completely empty, and Latin strings will get truncated at the first diacritic – which is going to affect pretty much any language other than English.

If I get to it, I'll test this behavior with a sample app to see if the issue is specific to Navit's software stack or if all native Android apps are affected.

PR #40, opened 2015-11-12
Status OSD

mvglasow:
Adds a new OSD for routing engine status (busy, routing, no route, waiting for position) and hides route/tracking-related OSDs when they are not relevant.

This is essentially a cleaned-up version of #35, which I am going to close because it got polluted with some unrelated commits.


sleske:
Hey, it's cool we have a new OSD element for routing status, which bundles everything.
Could you also update the Wiki docs under http://wiki.navit-project.org/index.php/OSD ? Otherwise noone will know about that nice feature :-).

mvglasow:
Merged and documented :-)

sleske:
Thanks, both for the cool feature (which I now realize was sorely missing), and for actually documenting it :+1: .

pohlinkzei:
There seems to be an Issue with this OSD. It does't 'listen' to the osd_configuration. In my case it should switch from white to black as the layout switches from night to day but for the StatusOSD it wont work. Both are displayed.

sleske:
@pohlinkzei: This pull request is closed, so it's probably not a good idea to discuss problems here. Could you file a bug on http://trac.navit-project.org/ , with reproduction instructions? Then we can hopefully solve it.

mvglasow:
@pohlinkzei: In the default configuration the status OSD comes with an `enable_expression`. If both `osd_configuration` and `enable_expression` are set, only `enable_expression` will take effect and `osd_configuration` will be ignored. This behavior is common to all OSDs.

As a workaround, you can either remove the `enable_expression` (note that this will

mvglasow:
@pohlinkzei: In the default configuration the status OSD comes with an `enable_expression`. If both `osd_configuration` and `enable_expression` are set, only `enable_expression` will take effect and `osd_configuration` will be ignored. This behavior is common to all OSDs, but since the status OSD is the first to come with an `enable_expression` by default, it may have gone mostly unnoticed so far.

You can remove the `enable_expression`, which will cause `osd_configuration` to take effect. Note that this will prevent the status OSD from being automatically hidden when route guidance is active or no destination is set – if the `osd_configuration` matches, it will show.

Alternatively, you can take advantage of the feature that every `osd_configuration` can be rewritten as a `enable_expression`. Setting `osd_configuration="42"` is the same as writing `enable_expression="osd_configuration=42"` (I'm not 100% sure on the syntax, it might also be `enable_expression="gui.osd_configuration=42"`). All you need to do is extend your `enable_expression` with an appropriate condition.

If this doesn't work, please file a bug on trac, as sleske suggested.

pohlinkzei:
Ok, that really works. I think its worth to be in the documentation. :D

mvglasow:
@pohlinkzei Definitely – I'll update it. Can you give me the exact syntax you used? That'll make it easier for me :-)

pohlinkzei:
For the StatusOSD I this worked: enable_expression="osd_configuration==2 &amp;&amp; (navigation.nav_status==-1 || navigation.nav_status==1 || navigation.nav_status==2)"

mvglasow:
Thanks – I see there is already a hint in the wiki (under http://wiki.navit-project.org/index.php/Configuration/Full_list_of_options#osd). Unfortunately the server's out of disk space so I can't edit anything at the moment...

mvglasow:
Documentation is updated, at http://wiki.navit-project.org/index.php/OSD#enable_expression and http://wiki.navit-project.org/index.php/Configuration/Full_list_of_options#osd.

PR #42, opened 2015-11-13
Refactor:gui_internal:More documentation

mvglasow:


sleske:
I know it does not really matter in the grand scheme of things, but I don't think it's necessary to have pull requests for doc updates. Just push to trunk right away (like I do :wink: ).

PR #49, opened 2015-12-01
fix:osd: Make osd "Button" accept h and w props

metalstrolch:
Now you can nicely scale osd buttons.
This helps if using pure .svg images, as .svg images don't have a size.

### DIFF
--- a/navit/osd/core/osd_core.c
+++ b/navit/osd/core/osd_core.c
@@ -1497,7 +1497,12 @@ osd_button_init(struct osd_priv_common *opc, struct navit *nav)
 
 	/* translate properties to real size */
 	osd_std_calculate_sizes(&opc->osd_item, navit_get_width(nav), navit_get_height(nav));
-
+	/* most graphics plugins cannot accept w=0 or h=0. They require special w=-1 or h=-1 for "no size"*/
+	if((opc->osd_item.w <= 0) || (opc->osd_item.h <=0))
+	{
+		opc->osd_item.w = -1;
+		opc->osd_item.h = -1;
+	}
 	dbg(lvl_debug, "enter\n");
 	dbg(lvl_debug, "Get: %s, %d, %d, %d, %d\n", this->src, opc->osd_item.rel_w, opc->osd_item.rel_h, opc->osd_item.w, opc->osd_item.h);
 	this->img = graphics_image_new_scaled(gra, this->src, opc->osd_item.w, opc->osd_item.h);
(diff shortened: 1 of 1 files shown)


sleske:
Thanks for the pull request. I'll have a look, to see if we can merge this shortly.

sleske:
There are some problems to be resolved before this can be merged (tested with PNG icons, may or may not apply to SVGs):
* This makes the w and h attributes mandatory for buttons (the button does not show if they are missing). The change must not break existing configs, so buttons without the attributes must work as before.
* It seems w and h do not support percentages (w="50%"). This should be added for consistency.
* If w and h have different values (w=30, h=50), the image's aspect ration does not change (as expected); instead the smaller value seems to apply. This should change the aspect ratio, or at least an error should be logged.

Could you fix these problems?

mdankov:
> If w and h have different values (w=30, h=50), the image's aspect ration does not change

This is probably graphics module dependent. I guess some support  changing aspect ratio, while others do not.

metalstrolch:
Hmm. a lot more seems to be graphics dependant.
* I don't see w and h props becoming mandatory on my qt5 port of qt_painter graphics, but i'll check this. Requested size seems to be -1 if no prop is given. To be honest I didn't check this with other graphics. What graphics are you using?

* Honestly I didn't care about percentage values, as I (wrongly) assumed that osd framweork already translates this to points. As you see I didn't change anything in getting pops from xml. Is this done in osd_std_calculate_sizes?

* Is there any documentation on struct osd_priv_common?

metalstrolch:
OK, tried to do this right. Now it should support percentages and such things. On my qt5 graphics there is no need to have h and w props for Buttons. Hopefully this works on other graphics too.

metalstrolch:
Rebased branch, and cleaned up commits. Confirmed to work with GTK graphics. Any interest?

sleske:
Thanks! I'm already looking at it. Sorry, I'm a bit busy so it may take a week or two, but I'll get back to you and hopefully merge soon.

pgrandin:
metalstrolch, I'll be interested in setting up some CI tests for your qt5 graphics. Can you please contact me?

sleske:
Hm, GTK graphics does not quite work for me - buttons don't show if w/h are missing. I used Navit trunk with default config (on Linux), I just enabled the OSD buttons in navit.xml. The buttons don't show, and there are assert messages on stdout - apparently the code does not handle missing w/h. I'll look into this. How did you test the GTK graphics?

metalstrolch:
Oops. Sorry my fault. I forgot to push one more change. Graphics seem to require .w and .h prop to be -1 if no size is given, but osd_std_calculate_sizes sets them to 0 instead.

sleske:
Thanks. I'm already looking at it.

sleske:
Merged. Thank you for your contribution.

metalstrolch:
@pgrandin: Did you get my e-mail? If not tell me how to contact you. Greets.

PR #51, opened 2015-12-16
fix:build Cope with build fron non git tree but git present

metalstrolch:
Tried to cmake navit (for rpm building) from tarball created from my fork. Since this is no git repository then, but git executable is present on the machine, getting git revision fails fatal. In detail the regex detecting "R" fails if VERSION is unset.
This small fix prevents it from failing and instead resorts to invalid (0000) revision.


sleske:
Thanks for the fix; merged.

PR #53, opened 2015-12-24
Automatic translation import

pgrandin:


pgrandin:
The translation content should be in the file, not the queue. There's an issue with the importer.

PR #60, opened 2015-12-27
Enable OSD color settings for non-text-OSD

pohlinkzei:
Change OSD-items-color for custom OSD-Layouts. I wanted to switch from night to daylayout with 100% transparent OSD backgrounds. This lead to the unreadable combination of white OSD on light grey day layout. Therefore I made some changes to set the foreground_color of an OSD item and replace the hardcoded graphics_fg_white. 

To set a custom foreground color, just add an foreground_color="#AAABBB" to the osd tag. It might be used with osd_configuration for the current layout. Default is white of course.

### DIFF
--- a/navit/attr_def.h
+++ b/navit/attr_def.h
@@ -427,6 +427,7 @@ ATTR2(0x00070000,type_color_begin)
 ATTR(color)
 ATTR_UNUSED
 ATTR(background_color)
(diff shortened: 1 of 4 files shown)


pgrandin:
Nice, thanks @pohlinkzei

PR #66, opened 2016-01-04
Default settings tweaks

pgrandin:
The default settings are not producing the most appealing map when starting Navit on Android, for example.

This is how the default map is currently rendered on the CI : default

The default zoom level renders an area that is too large, leading to the streets of lesser importance not being displayed. Most of the time when using Navit you mostly care about your next turn, not seeing the full city.

The default pitch value is too high, displaying a part of the "sky" which is currently not even blue, loosing in my opinion valuable space that could be used to display the map instead.

Default zoom with a suggested pitch of 10 : z=256,p=10

Suggested zoom of 32 with the default pitch of 20 : z=32,p=20

Suggested zoom of 32 with a suggested pitch of 10 : z=32,p=10

### DIFF
--- a/ci/build_linux.sh
+++ b/ci/build_linux.sh
@@ -3,7 +3,7 @@ if [[ "${CIRCLE_PROJECT_USERNAME}" == "navit-gps" && "${CIRCLE_BRANCH}" == "trun
 	export PATH=~/navit/cov-analysis-linux64-7.6.0/bin:$PATH
 	
 	mkdir bin && cd bin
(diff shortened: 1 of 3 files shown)


pgrandin:
APK for test with z=32 and p=10 : https://circle-artifacts.com/gh/navit-gps/navit/858/artifacts/0/tmp/circle-artifacts.960RNxV/navit-669fb11cb22e7a55056471da525b9c607232ba2c-debug.apk

pohlinkzei:
The usability of the pitch setting depends on the screen size. On my 320x234px screen the pitch of 20 doesn't seem to have an effect. So for another screensize these changes arent tweaks anymore.

pgrandin:
Good point @pohlinkzei thanks.
I will build another set of screenshots at various resolution to get a better idea of the results. The default screenshot method isn't optimal anyway, as we have some wasted/unused space in the window.

PR #69, opened 2016-01-20
Add:gui_internal:Added Googleplaces search from the internal gui

pgrandin:
This PR adds a new menu to gui_internal that allows users to search from googleplaces.
This has proven to be really efficient during my tests, especially in areas where house numbers are not tagged correctly, or when you know the name of a place but not its address.

This currenly adds two new dependencies :
- libcurl to post and get from the API endpoint
- jansson to process the resulting JSON

I tried to put all the network dependent code in a new file, network.c, and from there we could merge the other network related code (map downloading) and/or get rid of the libcurl implementation.

The feature is disabled by default ( use -D USE_GOOGLEPLACES:BOOL=TRUE to enable ) for now. Also, I am not satisfied with the following CMakeLists.txt code: 
```
list(APPEND NAVIT_LIBS "-ljansson -lcurl")
SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -u fetch_url_to_string ")
```

maybe @sleske or @mdankov will be able to figure out a better way of doing this :)


sleske:
Interesting feature. However, I don't think this is ready to merge as-is.

The CMake stuff is probably solvable. The Google API key is hardcoded in the source; that should be configurable via navit.xml or similar. Most importantly, IMHO the **Google icon SVG definitely has to go** - we really don't want to get into hot water with Google for infringing their copyright on the logo, or their trademark (or have you already cleared that with Google?).

Finally, and that is rather a personal opinion, I'm not sure fits well with Navit's purpose / direction (if there is one). To me, the main selling point of Navit is that it works "everywhere", with minimal requirements - it works completely off-line, does not require the use of proprietary services, works on different OSes, on desktop, mobile and embedded systems.

A feature that only works with an active internet connection (and currently only on Linux) does not fit that well. After all, if you are on a desktop system with a good net connection, you can just fire up any web-based system, which is probably more powerful than Navit.

That said, the feature is still cool and could work well as some kind of plugin; I just feel uncomfortable integrating it into Navit - particularly seeing that we already have so much dead and partially maintained code.

ernesto75:
I Agree. Especially using Openstreetmap + Googleplaces doesn't sounds good

(maybe other more open services like geocoder.opencagedata.com/api.html 
can be better...or foursquare api?)

ps: the 'we already have so much dead and partially maintained code' 
reminds me a question:
where can i find documentation about navit reading position from udp 
server? it can be used for friends tracking?

Il 21/01/2016 09:16, Sebastian Leske ha scritto:

pgrandin:
Thanks guys, I do appreciate the feedback.

The main purpose of this PR was not to be merged as-is, but indeed to start a discussion on the implementation itself, and get some help fixing the cmake uglinesses :)

My take on this feature (which I was not foreseeing controversial, but that's OK :) ) :
- regarding the Google icon : using branded icons is actually a requirements when using a Google API service. I double checked the terms of use, and actually we should only be missing a statement that complies with https://developers.google.com/maps/terms#section_9_3 
- I definitely agree with Navit should work offline and everywhere. This is the roots of the project. That being said, most devices and setups have Internet connectivity now, so I think that enabling the users to leverage it is a good thing. Using this feature, I usually found my destination with only 4 or 5 keypresses total. In terms of user experience, this is much better than our current search implementation.
- making it a plugin is a good idea. I am not aware of a search-only plugin as of now, so integrating it in the GUI was easier (at least for me)
- I actually have a similar code for Foursquare, but it would bring the same concerns
- This code is not enabled by default (the user needs to enable it, and configure his own API token) so it would not change the current behavior of Navit. But, I figured that I should try to get it merged instead of continuing to maintain it in my own fork (which I'd like to keep for only highly experimental features)
- There are still people using a carputer, so a full featured linux system could be something else than a desktop, where Navit could be the only application available ;) 
- I did not know about opencagedata, but it's really interesting. I will definitely check it out. Since these services ( GooglePlaces, Foursquare, OpenCage ) are basically technically the same, it should be quite easy to build a generic search backend which could be used to query any location search API.

Thanks guys for your comments. Triggering a discussion was exactly the purpose of this PR (along with getting help to fix the cmake code).

mvglasow:
I largely agree with sleske and ernesto75:

* Navit is an offline navigation system. That doesn’t mean we can’t integrate some features which rely on an Internet service, but these should be add-ons and Navit should continue to work without them.
* I am worried about integrating a dependency on a non-free service into Navit. If this is (or is going to be) a plugin, it could be kept separate from Navit and installed via an extra package. That way, users would be able to choose whether or not they want to install an add-on that depends on a non-free service. Of course, using a free backend (free data, even better free data and software) is an even better option.
* In terms of OSM – I fear we are going to do OSM a disservice if we fall back to Google wherever OSM is missing data, because that will remove the incentive to add missing data to the map.
* Regarding “this is much better than our current search implementation”: this sounds like a UX issue. Granted – search has a lot of room for improvement. For instance, we could pre-populate the search lists with entries that the user is likely to search for (country/town/street we are currently in, list of past searches...), or we could sort search results based on some ranking (the majority of users searching for “Milano” are looking for the capital of Lombardia, not the district of Milano in Spoleto, Perugia), or maybe we want to add freeform search? That is actually stuff for a separate ticket, and we might want to discuss it in more detail.

ernesto75:
Pardon me for the OT, but speeking about integrating features which rely 
on an Internet service....i'm too curious about the "friends tracking" 
experimental feature, aka the possibility for Navit to read gps position 
from udp server. What is the current status of that feature?

Navit README says:

Navit read the current vehicle position :
- directly from a file
- from gpsd (local or remote)
- from udp server (friends tracking) (experimental)


Is it similar to the friends server used by GpsDrive? As you can read at 
http://www.gpsdrive.de/documentation/manual.shtml

You can define a friends  server to exchange position information with
other people.You can start your own friends server with the program friendsd2,which is included.

There is a server program, called friendsd which acts as server for the
        position  of  your friends. If you enable it in the settings menu, then
        you can see the position of all gpsdrive connected with this server.

        You will see the position of your friends as a car symbol on  the  map,
        including the name, time, day of week and the speed of his last connec
        tion.  The blue arrow shows the last reported direction of your friend.

        The  time is transmitted as UTC, but shown on the display as your local
        time, so it is also correct if your friend lives in another time  zone.

        The  server uses port 50123 (UDP), so be sure that you open the port in
        your firewall. The server needs no root privileges and  should  run  as
        normal  user  or  a special user with no privileges. The server was NOT
        tested for security.

        There is a friends server running on friends.gpsdrive.de, you  can  try
        it if you enable it in the settings menu.

        You  can  also  send  messages to other mobile targets (Misc. Menu/Mes
        sages)

stephan75:
About better online search engine in navit:

Did you ever think of alternative engines like pelias or photon?

See the osm wiki about "search engines" for more information.

pgrandin:
I'm closing this PR and will reopen later with a more generic approach.

PR #70, opened 2016-01-21
[BugFix] Modified route_path_add_item_from_graph function in route.c:

koheiotsuka0213:
Now it doesn't cache the route_path_segment if it was the first one.
    This prevents the route_path_segment from having incorrect coordinates
    when updating it after current GPS position jumped back or
    played back GPS recording again without resetting Destination.

### DIFF
--- a/navit/route.c
+++ b/navit/route.c
@@ -1931,8 +1931,14 @@ route_path_add_item_from_graph(struct route_path *this, struct route_path *oldpa
 linkold:
 	segment->data->len=len;
 	segment->next=NULL;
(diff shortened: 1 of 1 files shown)


pgrandin:
@koheiotsuka0213 please submit the PR against the trunk branch, not master.
We push code in master only from trunk after a successful CI build ( as explained in http://wiki.navit-project.org/index.php/Testing_and_CI_strategy )

koheiotsuka0213:
Hi, 
I have just committed it to the trunk branch.

pgrandin:
Closing this since followup will be done in #84 
Thanks @koheiotsuka0213 for the patch still!

PR #71, opened 2016-01-22
fix:build Get Git in cmake to read git version

metalstrolch:
On my box, GIT_EXECUTABLE is never set during cmake build regardless if git is present or not.
In order to get GIT_EXECUTABLE set if the package is built, I propose including cmake Git module in version.cmake.

### DIFF
--- a/cmake/version.cmake
+++ b/cmake/version.cmake
@@ -1,4 +1,4 @@
-FIND_PROGRAM(GIT_EXECUTABLE NAMES git git.exe DOC "git command line client")
+find_program(GIT_EXECUTABLE NAMES git git.exe DOC "git command line client")
 
 get_filename_component(SOURCE_DIR ${SRC} PATH)
 


sleske:
Hm, strange that "FIND_PROGRAM" does not find your git install. Is your git binary in your path?

Using "find_package(Git)" is a nice idea; however, it looks like it was only added in CMake 2.8.2. We currently support CMake 2.6+, so we'd need to check this does not break old CMakes, or raise our CMake minimum (which may cause other problems).
So I'd prefer to stay with "FIND_PROGRAM". Maybe we can find out why it does not work in some cases.

metalstrolch:
OK, I'll try to check what's wrong with FIND_PROGRAM since you can confirm this works.. What's your cmake version?

I have two different systems to compile on: 1 some Texas Instruments sitara board with recent Gentoo linux on it and Qt5.5. Compilation for armhf native on this device.

The other is the MER build VM brought by the SailfishOS SDK. Cross compilation to armhf on x86 virtualbox.
Both have the same problem: git is not found without my patch. Both use cmake 3.3.1
But the results even with this patch seem to be odd in some way. May version says: 0.5.1-git:5efe5a8-
The trailing dash looks quite odd.

sleske:
@metalstrolch 
> OK, I'll try to check what's wrong with FIND_PROGRAM since you can confirm this works.. What's your cmake version?

CMake 3.2.2

>But the results even with this patch seem to be odd in some way.
> May version says: 0.5.1-git:5efe5a8-
> The trailing dash looks quite odd.

Yes, it does, but that's normal :smile: . We append the dash (in start_real.c , looks like some historical leftover). If you get a git hash in your version number, then git was found and run successfully.

To debug the problem, it's probably best to whip up a small CMake project that just checks for git.

Try this CMakeLists.txt:
> cmake_minimum_required(VERSION 2.6)
> project (Gittest C)
> find_program(GIT_BINARY NAMES "git.exe" "git")
> message(STATUS "Git binary found at: -- ${GIT_BINARY} --")

Just run "cmake ." - then it should print where it found git - or not.

metalstrolch:
The problem seems to be that version.cmake has FIND_PROGRAM call in capitals instead of small letters.
If I replace FIND_PROGRAM by find_program it seems to work.
Wondering how this ever worked.

metalstrolch:
Oops. I was wrong, indeed. cmake syntax is case insensitive. Stumbled over a copy paste err. Navit does not use GIT_BINARY as output var. So if you just copy paste the line from version.cmake you get a false negative.
Seems your example always finds git. Why doesn't the navit build.
Sorry for the fuzz. Still searching for the cause.

pgrandin:
@metalstrolch did you manage to find a solution to your problem? The patch attached to this PR does not seem useful to me.

pgrandin:
Bumping for comments. If no updates/review on the patch, I'll be tempted to close this PR.

metalstrolch:
Am 11.07.2016 um 22:39 schrieb Pierre GRANDIN:
>
> Bumping for comments. If no updates/review on the patch, I'll be 
> tempted to close this PR.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub 
> <https://github.com/navit-gps/navit/pull/71#issuecomment-231857572>, 
> or mute the thread 
> <https://github.com/notifications/unsubscribe/ADnyxSvfd8oZmfSzqE2bfRiVmBplHaj-ks5qUqoLgaJpZM4HKhqX>.
>
Oops, almost forgot about this.
Since I was unable to check this out any further until now, feel free to 
close the request.
I'll reopen once I found the real reason for the oddity on building on 
Sailfish OS.

pgrandin:
Closing, as discussed. Feel free to reopen for further investigation.

PR #74, opened 2016-02-22
Add 10s timeout before speed warner talks to you

pohlinkzei:
When max_speed changes while driving, the speed warner often announces unless I am decreasing speed. So I'd want to navit wait a little before complaining about my speed.

Another point is, that if navit places me on a road next to the e.g. highway/Autobahn I am currently on, it complains a lot while I am not breaking the speed limit on the road I am really on.

The timeout is based on position update. So if the position is updated once a second, navit will wait 10s before complaining about speed.


pohlinkzei:
The speed_warner can now be configured with the (already existing) attribute timeout. Default is 10.

pgrandin:
Looks good to me. Thanks for making this configurable!

PR #83, opened 2016-02-27
Switch layout - switch from day to night layout on gpio change

pohlinkzei:
Hi,

here another PR for navit on an embedded raspberry pi:

If GPIO 6  is high, the layout switches to night layout and vice versa - if there is one associated.
It uses WiringPi's GPIO command line utility to check the GPIO without root permissions, so WiringPi must be installed to use the plugin

The plugin must be activated by commandline during cmake prebuild with -DUSE_SWITCH_LAYOUT=true.


jandegr:
As mentioned on IRC I propose a layout switch command with wider usage and that does not crash on an inconsistent navit.xml

made a seperate PR for the switch command

https://github.com/navit-gps/navit/pull/90

pohlinkzei:
It makes half the PR obsolete if PR #90 gets merged.

pohlinkzei:
Restart. :D

PR #84, opened 2016-02-28
Modified route_path_add_item_from_graph function in route.c:

koheiotsuka0213:
```
Now it doesn't cache the route_path_segment if it was the first one.
This prevents the route_path_segment from having incorrect coordinates
when updating it after current GPS position jumed back or
played back GPS recording again without resetting Destination.
```

### DIFF
--- a/navit/route.c
+++ b/navit/route.c
@@ -1931,8 +1931,14 @@ route_path_add_item_from_graph(struct route_path *this, struct route_path *oldpa
 linkold:
 	segment->data->len=len;
 	segment->next=NULL;
-	item_hash_insert(this->path_hash,  &rgs->data.item, segment);
-
+        if (!pos) {
+	        item_hash_insert(this->path_hash,  &rgs->data.item, segment);
+	} else {
+	        /* Don't cache the route_path_segment if it was the first one.
+                 * This prevents the route_path_segment from having incorrect coordinates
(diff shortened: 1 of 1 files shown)


jandegr:
Hi,
AFAIK this makes navigation.c reprocess the entire route at each position update,

koheiotsuka0213:
Hi,
I believe that this make it reprocess only the first segment.

pgrandin:
@mdankov any opinion on this?

jandegr:
A little test shows route status gets stuck at 17, plan a 1000km or so route on mediocre hardware and the thing bricks, confirming my suspicion.

pgrandin:
@jandegr any advice on improving this?

pgrandin:
@mdankov @jandegr what's your take on this?

sleske:
Hm, the commit comment seems to hint at a bug:

> This prevents the route_path_segment from having incorrect coordinates
> when updating it after current GPS position jumed

I think we first need to understand how to reproduce that problem. Then we can see whether the patch addresses it.
@koheiotsuka0213 : Could you give us steps how to reproduce the problem you observed?

sleske:
Since there has been no feedback from the submitter, @koheiotsuka0213 , I'm closing this PR. Feel free to reopen once there is more information.

PR #85, opened 2016-03-03
Android: disable pedestrian default build

jandegr:
disable the building of the pedestrian plugin by default for the Android port


pgrandin:
Actually we could maybe even disable it completely. I am not sure that this plugin is useful anymore.

jandegr:
Feel free to remove it all together, but in either case I did not see any point in still distributing it in the apk's

PR #88, opened 2016-03-04
add destination forward/backward

jandegr:
navigation.c contains code to handle destination: backward and forward, but the changes needed in maptool to provide those did not get merged yet.


pgrandin:
Interesting. What is a backward destination?

jandegr:
for oneway roads OSM has a destination tag, but for non oneway roads there are the forward and backward variation to hold the destination for when you are driving in or against the direction of the OSM way.

PR #89, opened 2016-03-04
Change format_distance to more proper values

pohlinkzei:
In metric distance, all values between 10km and 1km were printed in meters. After this change all values over 100km are printed without the fractional part, between 10km and 1km in kilometers with fractional part and below that in meters as it was before.


jandegr:
looks indeed better this way

PR #90, opened 2016-03-06
Layoutswitcher

jandegr:
A wider usage version of the layout switch command in https://github.com/navit-gps/navit/pull/83

sample usage as a command in navit.xml 

    command="switch_layout_day_night(&quot;manual_day&quot;)"

see the code for more usage cases

### DIFF
--- a/navit/navit.c
+++ b/navit/navit.c
@@ -772,7 +772,7 @@ static GHashTable *cmd_attr_var_hash = NULL;
  * Store key value pair for the  command system (for int typed values)
  *
  * @param navit The navit instance
- * @param function unused (needed to match command function signiture)
+ * @param function unused (needed to match command function signature)
  * @param in input attributes in[0] is the key string, in[1] is the integer value to store
  * @param out output attributes, unused 
  * @param valid unused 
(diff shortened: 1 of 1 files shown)


pohlinkzei:
Would you mind making the auto_switch parameter int? I'd need a third state for this. :)

That would make it useful for the GPIO-Switcher: 
auto_switch == 0 -> off, no switching at all
auto_switch == 1 -> normal, switch on sunrise and sunset.
auto_switch == 2 -> gpio, switch to night_layout if GPIO is HIGH
auto_switch == 3 -> manual gpio*

*) That is my master use-case: The GPIO swticher plugin cares about the switching unless the user presses the toggle_layout osd button. If so, the layout will change once and on the next gpio change (the layout has now the according gpio (e.g. gpio=1 and night layout)) the gpio switcher takes control back.

jandegr:
Hi, I used the same signature as the other commands, that seemed safest to me.
So you'll have to call it with an attr. anyway and then switching to an int attr. would be nothing more than introducing a few more of Navit's famous magic numbers. As for your GPIO logic itself, I think that belongs in your GPIO plugin instead of in navit.c.

pohlinkzei:
Sure. The simple solution woild be to define "switch on sunset/sunrise" as 0 and manual to 1. Other states can be safely controlled by plugins without affecting navit.c

See [this](https://github.com/pohlinkzei/navit/blob/990f31088cd20dde46bcdc5811937af4c99ca19b/navit/navit.h#L33-L35), [this](https://github.com/pohlinkzei/navit/blob/990f31088cd20dde46bcdc5811937af4c99ca19b/navit/navit.c#L3729) and of course [this](https://github.com/pohlinkzei/navit/blob/990f31088cd20dde46bcdc5811937af4c99ca19b/navit/navit.c#L3797-L3846) for explanation. Finally three states for the modes are enough.

pohlinkzei:
I've been testting the layoutswitcher the last days as is. And it works like charm!

PR #95, opened 2016-04-27
Pause the autozoom feature when zoom_in() or zoom_out() is called

pohlinkzei:
This PR adds the feature to zoom out, look around and get back to autozoom after 10 gps updates. This will work with zoom in, too.


pgrandin:
Looks good to me. Thank you @pohlinkzei !

PR #101, opened 2016-06-13
Add an Uninstall target to remove navit from the (linux) system

pohlinkzei:
Hi,

i was annoyed from the need to remove the navit dir after sudo make install, so I googled an uninstall target for cmake and added it to our cmake.

I have no possibility to test this on other systems than linux, so I reduced it to that.

To test: just pull, make, sudo make install and then sudo make uninstall and be happy to have a clean install directory for the test or development of another branch.


pohlinkzei:
This is an approach to get the requested changes. But it does not work! If noone else can fix this issue, I will drop the PR. It is not important enough for me.

PR #111, opened 2016-09-01
Gui internal heightprofile

jandegr:
some fixes for the heightprofile in gui internal.
the name of the map with heightlines must be 
*.heightlines.bin


pgrandin:
Nice work fixing that @jandegr 

Attaching an example :
heightprofile

jandegr:
For completeness a view with heightlines on the map.
screenshot_2016-08-31-17-45-32

PR #112, opened 2016-09-02
Importing audio_framework code from my own repo

pgrandin:
This PR imports the audio_framework code from my own repo.

The purpose is to add media playback and volume control capabilities, in a modular way.

This PR comports 3 parts :
- some code to manage a new type of plugin ("audio") in navit
- a generic, very basic, playlist/track menu in gui_internal
- two new "audio" plugins : one to manage volume via alsa, and one to add playback capabilities via libspotify

Currently this code only works under Linux, as it requires alsa. Features are simply disabled if alsa (or libspotify) are not available.

To test the output-alsa plugin, you wil need:
- alsa
- an entry in your navit.xml such as 
`<audio type="output-alsa" audio_device_mixer="Headphone" audio_device="hw:1" />` (in the future we can add audio_device_mixer and audio_device detection and selection from the UI
- a way to trigger a volume change, for example 
`<osd x="-128" y="-64" w="64"  h="64" type="button" src="..." command='volume_down()'/>`

To test the player-spotify plugin, you will need :
- alsa
- libspotify
- a spotify premium account (other types of account won't work)
- a libspotify API key ( https://devaccount.spotify.com/my-account/keys/ )
- an entry in your navit.xml such as :
`<audio type="player-spotify" audio_playback_pcm="sysdefault:CARD=PCH" spotify_login="mylogin" spotify_password="mypassword" spotify_playlist="Default playlist" />`

Comments and patches are welcome. I am not satisfied with this code in CMakeLists.txt for example:
` list(APPEND NAVIT_LIBS "-lasound -lspotify -lpthread") # FIXME : this is a bad way to do it as maptool also gets linked`

I have been using this feature for about a year now, it's really convenient for carputer users.


sleske:
Hm, I'm really torn on this. On the one hand, the code looks decent, and works. On the other hand, IMHO the one thing Navit suffers from most is a lack of clear focus - we have loads of old code, unmaintained plugins, ports that barely work. So in that respect the last thing we need is more code that is only tangentially related to map display and navigation. I think I'll abstain from voting :-).

jandegr:
I only have one request, make it disabled by default, even if the libs are found.

pgrandin:
> On the one hand, the code looks decent, and works. 
That's a good start :)

> On the other hand, IMHO the one thing Navit suffers from most is a lack of clear focus - we have loads of old code, unmaintained plugins, ports that barely work. So in that respect the last thing we need is more code that is only tangentially related to map display and navigation. I think I'll abstain from voting :-).
I think that there is two things to consider here. We need to get better at preventing code obsolescence. One way to do it is to add build tests and have a better CI/CD to detect breakage earlier. The second thing is needs from our userbase. I think that the userbase interested in Navit is probably a bit different today from what it was a few years ago. In the past, Navit was the only free and opensource navigation software. Today, a lot of people are just using an app on their smartphone (usually only free). On the opposite, I think that cheap and accessible platforms like the raspberry pi brought a new wave of carputer enthusiasts. There is a few pre-bundled solutions integrating Navit, but in my opinion all lack a seemless integration of the applications they are composed of. Switching from an app to another will cause a flicker on your screen, your applications can't really talk with each other, etc. A good example is the one that uses Kodi for the media part. 

I do understand your concern. If we diversify, there is a risk that we don't make our core functionalities better. But in my opinion, if we are able to reach a new and broader userbase, it can be good and it could also bring in new contributors, willing to help us improve ( @pohlinkzei is a good example : https://github.com/pohlinkzei )

I take your concerns seriously @sleske and I'd really like you to be onboard before merging this. Let me know how we can address your concerns.

> I only have one request, make it disabled by default, even if the libs are found.
Good idea. I will do that and update the PR.

sleske:
@pgrandin : I may be over-cautious. If one of our goals is to become a better solution for onboard entertainment, then the code is quite on-topic. That's not how I use Navit, but I'm not everyone.
I think my worries will be reduced if we make the code neatly decoupled from the rest of Navit, to the extent practical. As far as I can tell, this is the case (I didn't read everything).
Then we can just try integrating it - and worst case we just rip it out again later. If we can manage that, count me in :-).

sleske:
I'm mainly worried about repeating something like the "pedestrian" plugin - there is quite a lot of code in core Navit that is only used by that plugin (and was added just for it), yet the plugin is unmaintained, and doesn't even have clear docs in the Wiki, so even if you wanted to use it, you won't find out how to. I think we really should scrap all that code - I just didn't feel like finding out if someone does in fact use it, and then ripping out all the dependencies. Ok, enough ranting...

asanoaozora:
Hello Navit team,
first of all I'd like to thank you for keep on maintaining and improving Navit.
In the GitHub there are tons of code, and I won't be surprised if you never heard about GENIVI, the navigation APIs and the use of Navit as a navicore for proof of concepts and demonstrator. So, I take the liberty of introducing a little bit what I'm doing with your code.
Six years ago, we (GENIVI) had to choose an open source solution for POC, and, because of its flexible mechanism of plugins we decided to work with Martin Schaller. He developed several plugins (funded by GENIVI), and after his 'reconversion' as a farmer somewhere in Germany, 
I've continued to maintain the code, to adapt it to new versions of APIs...
About 3 years ago, I proposed to develop a demonstrator, based on navit and plugins, and, even if it's not really a 'navigation product', it runs (on desktop under Linux and on board, with a  Yocto build).
So, this quick introduction is just to say that, each time I make presentation of the code, I always mention and promote Navit as a flexible navigation core.
I'm aligned with the latest version of Navit (I'm not using the HMI, I'm implemented another one), except one patch.
I've no contact with Navit team since Martin left, and, because my knowledge of Navit architecture and code is limited, some improvement I have in mind can't be achieved.
All the code pushed by GENIVI is MPLV2, documentation and APIs are CC by 4.0
In my opinion, Navit is a navigation solution, and not really an IVI solution, so audio player plugin doesn't make really sense for me.
I'd make more sense to add dead reckoning, HMI with cross road zoom, or to add a plugin to connect to POI providers, to extend the embedded database.
And of course, improving documentation would be helpful ;-) 
I don't know how you're using Navit (I guess it's more the Android version), but, maybe having a look at what is done in GENIVI could be interesting for you.
On my side, I have a watch on what you're doing.
Regards.
Philippe

sleske:
@asanoaozora : Welcome, and thank you for the interesting feedback. I had never heard of Genivi. Is this the repository? https://github.com/GENIVI/navigation

pgrandin:
@sleske thanks for the update. This PR only adds a very basic code in navit's core, which is inspired from the way we manage vehicles. Regarding the pedestrian plugin, you are right and we should probably create a branch and use something like gcov/lcov to do some cleanup. 

@asanoaozora welcome and thanks for contributing. I actually know about Genivi since some time, so it's nice to have a direct contact with someone from the project. Martin leaving was definitely a big loss for the project, but we have to continue without him. If you have needs concerning Navit features, please let us know, and we'll definitely do our best to help you. The easiest way is probably to discuss it first on IRC. I'm always happy when Navit is used in another application or setup.

Now back to this PR : I read your answer yesterday, and I thought "damn, he might be right and maybe this PR should rather be against Genivi". So I took some time to freshen up my knowledge of Genivi and see if 1) I could push my spotify plugin as part of Genivi and 2) how things have changed since the last time I looked at Genevi a few months ago (I do check in periodically).

I was happy to find that there is a fresh release with a build for raspberry pi. Perfect, that's my target! I downloaded it, flashed it, and booted the board. Unfortunately, gdp-11-rc1-rpi is really a demo platform. The navigation isn't even Navit in this build, just a demo, and the media player is "xeyes". No big deal I though, let's have a look at the code. I scoured your wiki, looked at various repositories, and honestly your codebase is quite complicated. I understand that given the number of actors involved in this project some decoupling is needed. The architecture described in your wiki makes sense. But if we look at this from the point of view of a hobbyist, this is really, really complex. Where do I start if I just want Navit and a media player for example? Building this project seems like a daunting task...

Please do no take this reply as a negative rant, because it is really not my point. I just don't really know where to start if I wanted to add features. I do hope that we can collaborate to make this better, and I'll start to hang out in the #automotive channel. Feel free also to ping me in #navit (my nickname is KaZeR) to move this discussion forward.

asanoaozora:
For sure, I forgot to give you some more details about the code.... 
For architectural reasons, in GENIVI, Navigation is shared between three repositories:
- [positioning ](https://github.com/GENIVI/positioning)(interfaces for sensors, GNSS and also enhanced position (dead recknoning)). This repository also contains a log replayer, to simulate inputs by using a log file with tags
- [navigation ](https://github.com/GENIVI/navigation) core (interfaces with location input, route calculation, guidance, map viewer, POI search)
- [navigation application](https://github.com/GENIVI/navigation-application) (a simple navigation application, HMI in Qt QML)
There's also some stuff into this [wiki](https://at.projects.genivi.org/wiki/display/NAV/IVI+Navigation+Home)
I'll come back to you guys, by end of October. Until this date, I'm going to have a look at your Android version (I'm not familiar with Android, so I need some time to train me a little bit).
Cheers !

pohlinkzei:
I think navit's modularity is its key feature. If we only maintain the core, the modules will stay behind. For Android, Windows and iOS builds this framework is completely useless, but for the masses of ARM-based, linux driving micromashines like the RaspberryPi, Beaglebone and many others this feature is great. And I want this stuff available and maintained.  

I'd recommend to switch from the lot of functions and commands that are inside navit.c to using attrs for all the communication between the parts.

pgrandin:
Thanks everyone who participated in the thread, this was definitely useful.
I am merging the PR (to its own branch) and we will work with @pohlinkzei on improving it.

The code will live in its own branch for some time, and we should keep the discussion open about how to improve plugins maintainability.

@asanoaozora : i would really like to work with you on improving the collaboration between the two projects.

mvglasow:
@pgrandin @asanoaozora Interesting to hear there is some degree of interoperation between GENIVI and Navit. I'd had a look at GENIVI in the past but didn't dig deeply into it.

Inertial navigation is also an interesting point. I'd started working on something in that direction (Android-based) a while back but don't yet have anything usable – the topic is quite complex and errors accumulate quickly. Plus, currently the needs of my day job are kicking in big time, leaving me with little time to work on this topic.

The main design question for me at the moment is where inertial navigation code should go.

Should it be kept separate from Navit?
- Avoids adding more platform-specific code to Navit (needed to interface with GPS and sensors)
- Other apps can benefit from it as well, even on devices without Navit

Or should it be integrated into Navit?
- High-level processing (on top of an abstraction layer) is probably reusable across platforms
- Allows for tighter integration, such as using map/route data to improve inertial navigation accuracy

Or something in between (however that may be implemented) which would give us the best of both worlds?

Last but not least: what would be a good place to continue this discussion? Probably not this merge request...

pgrandin:
> Last but not least: what would be a good place to continue this discussion? Probably not this merge request...

When we moved from sourceforge's svn to github, we decided to keep trac for this kind of purpose.
We can discuss it, see if it still makes sense, and hear what people are thinking about it. Maybe discuss it in the forums, or trac itself to avoid using this PR :)

PR #115, opened 2016-09-08
Gui internal heightprofile

jandegr:
The last of the basic fixes for heightprofile / heightlines.
- fixes an occasional crash
- add a new icon
- rework xml for heightlines in the map view

Toggle the layer heightlines in gui_internal->settings->display->layers to toggle visibility
of the heightlines in the map view.

If I ever work on this again, the first job would be to move most of this code out of gui_internal and make it usable for other gui's as well and leave only the part that renders it in gui_internal, but now it was just fixing the existing code a bit.


pgrandin:
Nice work, Jan!

PR #117, opened 2016-09-14
autotools

jandegr:
try to fix a few builds on the old server still using autotools now breaking on removed NavitAndroidOverlay.java


jandegr:
if this works then the next problem would be to fix 

In file included from /work/compile/navit/src/navit/start_real.c:29:
/work/compile/navit/src/navit/config_.h:16: warning: parameter has incomplete type
/work/compile/navit/src/navit/start_real.c:54: error: syntax error before "GIT_VERSION"

as can be seen in many of the build logs after aug. 30 2015

PR #118, opened 2016-09-21
Update Makefile.am

jandegr:
This will not properly set the version for release builds, but it should be another step forward in getting the autotools builds on the Navit server going again.


sleske:
@jandegr : Which build still needs autotools? The idea used to be to get rid of autotools in favor of CMake. See http://trac.navit-project.org/ticket/1341 . If there are still builds/platforms where CMake does not work, it would be good to note that in trac ticket 1341, so we can fix it.

pgrandin:
If I remember correctly, this is used for the Tomtom build.
@jandegr is working on it.

PR #119, opened 2016-10-19
Tomtom2

jandegr:
Builds a package for Tomtom similar to those @xenos1984  used to post
[http://wiki.navit-project.org/index.php/TomTom](url) and as a continuation of
[http://trac.navit-project.org/ticket/1144](url)

- uses Cmake for Navit instead of autotools
- has translations and espeak + espeakdsp from Gernot
- does not (yet) provide a config for 320x240 screens
- once finished HAVE_API_TOMTOM was not needed anymore but it is preserved for future changes
- tested on a Tomtom go 710


circle.yml
```diff
@@ -6,9 +6,11 @@ dependencies:
     - ~/.android
     - ~/android
     - "~/assets/"
+    - ~/tomtom_assets
   pre:
     - "[ -d ~/assets ] || mkdir ~/assets"
     - "[ -d ~/android ] || mkdir ~/android"
+    - "[ -d ~/tomtom_assets ] || mkdir ~/tomtom_assets"
```
pgrandin: We should probably take care of that in the tomtom specific script.

Eventually we should simplify the circle.yml file so that it only triggers sub jobs, but it should not really care about the jobs dependencies.

ci/build_tomtom.sh
```diff
@@ -0,0 +1,525 @@
...
+		ioctl(fh, SNDCTL_DSP_SPEED , &rate);
+		ioctl(fh, SNDCTL_DSP_SYNC, 0);
+		while(l)
+		{
+			for(co=0,wp=0;co<IBUFFERLEN;co++)
+			{
+				bufo[wp++]=bufi[co]; /* mono->stereo */
+				bufo[wp++]=bufi[co];
+			}
+			write (fh,bufo,wp);
```
pgrandin: Should we include this file in the repo instead?

ci/build_tomtom.sh
```diff
@@ -0,0 +1,525 @@
...
+# tar jxvf flite-2.0.0-release.tar.bz2
+# cd flite-2.0.0-release
+# ./configure --prefix=$PREFIX --host=$ARCH
+# make -j$JOBS
+# make install
+
+# in the end we only want Navit locale
+rm -r $PREFIX/share/locale
+
+cat > ~/navit/navit/xpm/tomtom_minus.svg << EOF
```
pgrandin: Maybe we should add the files to the repo instead?

jandegr:
I did some cleanup and switched to build_tomtom_minimal.sh that does only the minimum to build for tomtom and does nothing of the funny stuff. The original script is renamed into build_tomtom_plugin.sh for in case somebody wants to build a new navitom package as mentioned in the wiki

jandegr:
a link to the plugin build (circleci 1146)
https://1146-30791823-gh.circle-artifacts.com/0/tmp/circle-artifacts.QLIgxrQ/navitom.zip

gefin:
This version (now merged) route real faster. Also it work now well with 64M RAM. Navit work on the TT730  well in the opentom environment, but in tomtom mode it don't bind the navit button into the menu. 
maybe the newer firmware makes problems .

jandegr:
The plugin works on older tomtom navcores, I don't konw the exact numbers but from what I know it does not work on navcores having lanes assist or junction view, but found it working on older versions, at some point tomtom disabled plugins. Most important is it works properly in opentom.

PR #122, opened 2016-11-30
Upgrade our CircleCI image to Trusty

pgrandin:
We were currently running our tests on Precise.
Moved most of the dependencies installation in each build script
Also, enabled parallelization of builds which reduce the total build time by ~5 minutes (we can do better by distributing the jobs differently)


jandegr:
it now fetches the android ndk for each of the android builds, I think this leads to a problem when both android builds are done in the same container.

`Starting build script #6 : ci/build_android.sh
replace /home/ubuntu/android-ndk-r13/toolchains/x86-4.9/prebuilt/linux-x86_64/COPYING? [y]es, [n]o, [A]ll, [N]one, [r]ename: 
command bash ci/build.sh took more than 10 minutes since last output`

[https://circleci.com/gh/navit-gps/navit/1367](url)

pgrandin:
Good catch @jandegr 
I restored the branch for further work, the fix should be simple.

pgrandin:
Should be fixed in https://github.com/navit-gps/navit/commit/53a5f48ab18810625d48911e38d85cff67ffd6d9

PR #183, opened 2016-12-03
Update:i18n:Updated Turkish (tr)  translation from launchpad

pgrandin:


sleske:
Hi,

it's good that translations from launchpad are imported regularly now.
However, is there maybe a way that causes less commits and merges? Right now, the recent commit history is rather messy and hard to read, because of all those commits and merges. Out of the last 200 commits, over 100 are I18N commits - one commit and one merge commit for each I18N file.

Could you not do a bulk import for all languages in one commit? And can we maybe do it without pull requests (this also makes the list of PRs on Github quite long)? Can't we just commit changes to trunk directly (or rebase the merges before pushing, or something like that)?

Greetings

Sebastian

xenos1984:
I fully agree. It would improve the readability of the history a lot to have just one single commit for all translation updates at once instead of this huge list.

pgrandin:
The fact that we ended up with so many PR and merge is a bug, indeed.

Usually we import translation updates one by one, when someone notice that a given translation has been updated on launchpad. This is a manual process that relies on a helper script.

My intent was to move forward with automation, by downloading all translations from launchpad maybe once a day, and import the updates automatically.

I should have tested the import script on a clone of the repo, and I apologize for the noise. Next time I will work on something as potentially disruptive for the history and readability I will do that.

Some of the noise has been caused by a trivial change in the translation where a block of text has been moved somewhere else. Technically the translation did not change, but the content of the file changed from git's perspective.

We need to test each translation separately, because sometime an update will break the compilation. This is why using branches is convenient.

The next step for this import process was also to have auto merge when tests in a branch succeed, as we do for the trunk -> master merge (so we still get two commits for each update, but no more PR). I was planning on doing that this week-end.

Would a rebase avoid the need for these two commits?

sleske:
HI,
and thanks for the prompt reply.

> I should have tested the import script on a clone of the repo, and I apologize for the noise. Next time I will work on something as potentially disruptive for the history and readability I will do that.

It's not a big deal, the main thing is we find a solution that works in the longer run.

> We need to test each translation separately, because sometime an update will break the compilation. This is why using branches is convenient.

Yes, if updates break the compilation, we will need a branch. However - is it unavoidable that updates break things? As long as the format of the .po file is ok, nothing should break. So maybe we can avoid these breakages by making them import script stricter?

> The next step for this import process was also to have auto merge when tests in a branch succeed, as we do for the trunk -> master merge (so we still get two commits for each update, but no more PR). I was planning on doing that this week-end.
> Would a rebase avoid the need for these two commits?

Yes, a rebase would mean we only need a single commit. The only thing "lost" would be the information where the "import branch" was branched off master - but that is an arbitrary point, as the branches are created automatically. So I think here a rebase would make more sense.

pgrandin:
> It's not a big deal, the main thing is we find a solution that works in the longer run.

Definitely.

> However - is it unavoidable that updates break things? As long as the format of the .po file is ok, nothing should break. So maybe we can avoid these breakages by making them import script stricter?

Since we import the files from launchpad, it is a bit hard to control how they are edited. If a translator adds an extra %s in a string, for example, compilation can break.

> Yes, a rebase would mean we only need a single commit. The only thing "lost" would be the information where the "import branch" was branched off master - but that is an arbitrary point, as the branches are created automatically. So I think here a rebase would make more sense.

Sounds good. Let's test it on this PR!

pgrandin:
Here is a good example of a translation breaking the build:
https://circleci.com/gh/navit-gps/navit/1463

`[ 99%] Generating bg.mo
/home/ubuntu/navit/bin/po/bg.po:12: nplurals = 2...
/home/ubuntu/navit/bin/po/bg.po:96: ...but some messages have only one plural form
/usr/bin/msgfmt: found 1 fatal error
/home/ubuntu/navit/bin/po/bg.po: 430 translated messages, 315 untranslated messages.
make[2]: *** [po/bg.mo] Error 1
make[1]: *** [po/CMakeFiles/locales.dir/all] Error 2
make: *** [all] Error 2`

PR #197, opened 2017-01-02
Add:CI:Add basic UI tests

pgrandin:
This PR adds some basic UI testing and generates a .gif with the results:

Example UI tests result


pgrandin:
@jandegr this is correct for now.
There is several possible ways to work around this : 
- monitor dbus for events : e.g. at the end of a specific test, we should have a route computed
- use OpenCV to detect if we are in the correct sub-menu, or if we get a specific item on screen. I have some proof of concept code to do that

PR #198, opened 2017-01-04
Reduce verbosity and reenable Rotary Encoder to zoom in and out.

pohlinkzei:
Hi,

I used to use a rotary encoder to control navit. The rotary triggers a virtual keypress on page up and page down. This keys should be used for zoom in and zoom out in map view, since all attempts to solve this with configuration in navit.xml failed.

Addionally some false-error messages will be removed by this PR.


navit/audio.c
```diff
@@ -207,16 +207,16 @@ audio_new(struct attr *parent, struct attr **attrs)
 		g_free(this_);
 		return NULL;
 	}
-	dbg(lvl_error, "Attrs: %p\n", attrs);
+	dbg(lvl_info, "Attrs: %p\n", attrs);
 	/*
 	if(attrs != NULL){
```
pgrandin: We should avoid dead code and probably just remove that block.

pohlinkzei:
I found a way to access the ZoomIn and ZoomOut from the OSD overlay, so the the keys aren't hardcoded anymore. But we should think about why we need 8bit broad key values. We should use utf8 instead.

pgrandin:
I think that we're going to need to rebase this on current trunk, or submit a new PR, this one is getting difficult to review.

pohlinkzei:
Oh, I made a bit of a mess. I must admit that I forgot this PR was open and continued developing.

So closing is the best option.

PR #199, opened 2017-01-13
Updated: F-Droid download link

zintor:


pgrandin:
Thanks for catching it @zintor !

PR #200, opened 2017-01-21
fix buffer usage

gefin:
let espeakdsp don't send more bytes then received.
This solve the noise problem speaking empty strings


pgrandin:
Thanks!

PR #204, opened 2017-01-26
autotools lefotover

jandegr:
unless I am mistaken, this is an autotools leftover


pgrandin:
Good catch, thank you!

PR #205, opened 2017-01-31
Fix:port/android:Fix "invalid DT_NEEDED" warnings on API 23+, fixes #1348

### LINKED ISSUE #1348: recalculate route on traffic distortions
currently, entering a traffic distortion does not cause the route to be recalculated. In my tests, even stopping navigation and starting it again did not cause routing to go around the blocked road.
So there may be two issues here:
* ensure manually-entered distortions are considered when routing
* re-calculate the route when distortions are entered

mvglasow:
I'll merge this as soon as CI finishes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -600,9 +600,6 @@ if (SAMPLE_MAP)
 endif(SAMPLE_MAP)
 
 if(ANDROID)
-   # for android API 3 compatiblity
-   SET(CMAKE_SHARED_LIBRARY_SONAME_C_FLAG "-Wl,-soname,/data/data/org.navitproject.navit/lib/")
-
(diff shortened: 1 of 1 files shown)


mvglasow:
CI fails due to apparently unrelated issues—I reverted this change on a separate branch, and I get the same error without my changes.

The latest error I’ve been getting, both with and without my fix, is:

    ci/build_tomtom_minimal.sh: line 96: tomtom.cache: Permission denied

I have successfully built an APK, and the only change is in an Android-specific section, so it shouldn’t affect other builds. If it does break something, feel free to revert (a heads-up would be appreciated).

mvglasow:
CI finished successfully after merging—the issue seems to be with my own setup. I accidentally merged into master instead of trunk—fixed this by merging master into trunk, so everything should be OK now.

PR #206, opened 2017-02-17
sunrise near poles simplification

jandegr:
Remove redundant sunrise/set test.
If the first test shows we are near a pole, it will never get to the second test.
If we are not near a pole, a second test will result in the same.


pgrandin:
Thanks @jandegr !

PR #207, opened 2017-02-20
Add Qt5 graphics and vehicle

metalstrolch:
This pull request is about adding basic qt5 support to navit.
It brings a qt5 graphics plug in.
It brings a qt5 vehicle plug in.


pohlinkzei:
Hi, can you provide some information how to get it working? It doesn't run out of the box on Ubuntu, so I could need some input to find out why.

Thanks.

metalstrolch:
Hmm, never thought that the configuration of this is non standard. It's just graphics="qt5" usually.

What kinds of tests does CircleCI do to navit when the plugin is enabled? And where to provide the navit.xml? Within the navit/graphics/qt5 directory?
Unfortunately I'm not an CircleCI expert.

metalstrolch:
You can have vehicle="qt5" as well, if you have set up QWhereabouts correctly. It's the only way to get GPS fix on Jolla phones as there is no gpsd.

pohlinkzei:
CI tests the navit compilation and starts navit, does some nasty, magic stuff i forgot and closes navit afterwards. 

It is easy as can be: I'd add the to be tested configuration into navit-sipped.xml, because this config will be used as standard. It does make the merge a bit more complicated, but tests the configuration.

If the graphics fail on startup of navit, ci wont get it if the plugin is inactive and has compiled fine. If the qt5 graphics do a segfault on startup, CI only gets it if the graphics were activated. Maybe we should add tests for multiple configs at one compilation. Or so.

metalstrolch:
Do we really want to "globally" change the default configuration of this? Because the up to now standard gtk graphics will get disabled? Wouldn't a change in default configuration confuse users?

pohlinkzei:
Not if there is another way to test the plugin. For a qt5 related branch it makes sense to change the global config for CI. But we can handle the global config during merge. It is right: someone has to take care of it.

metalstrolch:
OK. I tried to keep the work as clean as possible, just changing things i have to.
But I can provide a changed default config and have a look what CircleCI results.
Then you cannot just merge this anymore, but so what.
And what about that topic after a merge? No more CircleCI for Qt5 at all?

pohlinkzei:
We have to add tests for all maintained GUI/Graphics. And that might be the point to drop changes of the navit-shipped.xml config. But as I am neither a CI expert, we should discuss this on iRC, too. ;)

pgrandin:
Nice!

I am really happy to see work happening on QT5, this opens the door for a much better UI using QML2.

I think that there is a few stuffs that we can improve in this PR:
- we should drop navit/graphics/qt5/Makefile.am as we intend to support cmake only
- in the ci/run_linux_tests.sh script, we should maybe split the tests in two : run once against gtk + internal, then using xslt (or even sed, as we do in "ci/run_linux_tests.sh", but a xslt would allow for a group of settings to be applied at once) we can switch to the qt5 graphics_driver and run another set of tests. This will avoid the need to modify navit_shipped.xml and we can revert that change too. 

Otherwise nice job @metalstrolch !

I can take care of these changes a bit later if nobody can do it before.

navit/graphics/qt5/Makefile.am
```diff
@@ -0,0 +1,9 @@
+include $(top_srcdir)/Makefile.inc
```
pgrandin: We should drop autotools support, as per a general effort to switch to cmake

navit/navit_shipped.xml
```diff
@@ -38,7 +38,7 @@
 	<navit center="11.5666 48.1333" zoom="256" tracking="1" orientation="-1" recent_dest="250" drag_bitmap="0">
 		<!-- Use one of gtk_drawing_area, qt_qpainter or sdl. 
 		     On windows systems, use win32 -->
-		<graphics type="gtk_drawing_area"/>
+		<graphics type="qt5" w="792" h="547"/>
```
pgrandin: We should revert this change and leverage sed or xslt from ci/run_linux.sh instead

navit/vehicle/qt5/Makefile.am
```diff
@@ -0,0 +1,9 @@
+include $(top_srcdir)/Makefile.inc
```
pgrandin: We want to support cmake only

navit/graphics/qt5/Makefile.am
metalstrolch: Was already in work before autotools removal. Removed.

navit/vehicle/qt5/Makefile.am
metalstrolch: Was already in work before autotools removal. Removed.

navit/navit_shipped.xml
metalstrolch: Damn it. Will wait for somebody fixing CircleCI. Never mind.

metalstrolch: Reverted it.

metalstrolch:
@pgrandin 
Would be nice of you to take care about the CircleCI stuff, as I did not fully understand the CircleCI setup used here. - thanks -

pgrandin:
@metalstrolch nice work on https://github.com/navit-gps/navit/commit/bacca401ffaabd078d31000ff0ea201cd3a6a92b 
Exactly what I had in mind :)

PR #210, opened 2017-02-28
Hide impossible keys at the internal keyboard instead of highlighting the possible ones (leftover from PR5)

pohlinkzei:
Hi,

If we use the internal gui for searching a country/town/street, at a point are the possible next keys highlighted to give a hint to the user which keys are possible and which are not. This PR goes the other way around and disables the impossible next keys.

The main reason for this is if you access the gui through a keyboard (or buttons or rotary encoder) the selected and highlighted key are highlighted the same way, so one can hardly divide highlighted from selected.

The town search was tested with the sample map and munich, so I expect it is working with (any) umlauts.

The feature can be activated in navit.xml: Add the attr 'hide_next_possible_keys="1"' to <gui type="internal" [...] />


sleske:
Thanks for the fix. I have closed the corresponding trac ticket - http://trac.navit-project.org/ticket/1339 .

sleske:
I did some cleanup work:

I renamed the new option to "hide_impossible_next_keys", because it hides _impossible_ keys, not possible keys.

So the right syntax to use in navit.xml is:

`
gui type="internal" hide_impossible_next_keys="1" ...
`

I also documented this in the WIki, on http://wiki.navit-project.org/index.php/Configuration/Full_list_of_options#gui and http://wiki.navit-project.org/index.php/Internal_GUI#Using_the_Internal_GUI .
That way people will actually know about this :-).

sleske:
Finally, I noted a few remaining problems with hiding the keys:

If you enter some characters, then delete the last one, the number of possible keys typically increases. However, the once again possible keys only appear slowly as the search progresses (especially on slow devices, search can take a while). On deleting a character, we should probably immediately show all keys, then hide them later when search is finished. Probably not a major problem, as deleting characters is not so typical.

In some cases, the hiding feature fails to find the possible keys, and falls back to showing all:
    * If the search input uses incorrect casing (search for "MU" with the sample map)
    * If the search input uses special characters such as umlauts (search for "Mü")
    * If the search input is for the second or later word in a multi-word name (search for "Farch", finds  "Wolfratshausen (Farchet)").

The problem of failing to find possible keys is old, however, it also occurs when highlighting keys instead of hiding them, so is not caused by this PR.

Maybe we can get to this later...

PR #211, opened 2017-03-02
Postcodesearch_essentials

jandegr:
a tweak in binfile.c so a gui can be made to search for attr_town_postal in a similar way as a search is done for attr_town_or_district_name.


jandegr:
just some usage sample

screenshot_2017-03-01-14-03-31small

PR #212, opened 2017-03-25
No libcrypto test trac 1350

jandegr:
see the issue in http://trac.navit-project.org/ticket/1350

I found no usage of libcrypto in the code and my best guess is this has at some point
been used for some non-disclosed or removed code.
If somebody could please test it on linux and then remove the lines I commented-out, I can't test it on linux right now.


gefin:
I was able to build  working navit from no_libcrypto_test_trac_1350 without errors.
OS : debian 7

cmake summary:
-- SVG2PNG-Converter: /usr/bin/convert
-- Use plugins: MODULE

Summary:


--->>> autoload
Disabled  osso ( Default )

--->>> binding
Disabled  dbus ( dbus-glib-1 not found )
Disabled  python ( python libraries not found )
Disabled  win32 ( Default )

--->>> font
Enabled   freetype ( freetype found )

--->>> graphics
Disabled  android ( FreeType library not found )
Disabled  cocoa ( Default )
Disabled  egl ( Required library not found )
Disabled  gd ( FreeType library not found )
Enabled   gtk_drawing_area ( GTK libs found )
Enabled   null ( Default )
Disabled  opengl ( FreeType library not found )
Disabled  qt5 ( Qt5 libraries not found )
Disabled  qt_qpainter ( Qt libraries not found )
Disabled  sdl ( Required library not found )

--->>> gui
Enabled   gtk ( GTK libs found )
Enabled   internal ( Default )
Disabled  qml ( Qt Declarative not found )

--->>> map
Enabled   binfile ( Default )
Enabled   csv ( Default )
Enabled   filter ( Default )
Disabled  garmin ( Garmin library not found )
Enabled   mg ( Default )
Enabled   shapefile ( Default )
Enabled   textfile ( Default )

--->>> osd
Enabled   core ( Default )

--->>> plugin
Disabled  j1850 ( Default )
Disabled  pedestrian ( Default )

--->>> speech
Disabled  android ( Default )
Enabled   cmdline ( system() call is available )
Disabled  dbus ( dbus-glib-1 not found )
Disabled  espeak ( Default )
Disabled  iphone ( Default )
Disabled  speech_dispatcher ( speech_dispatcher lib not found )

--->>> support
Disabled  espeak ( Default )
Disabled  ezxml ( native Glib found )
Disabled  gettext_intl ( native libintl found )
Disabled  glib ( native Glib found )
Disabled  libpng ( native libpng found )
Enabled   shapefile ( Default )
Disabled  wordexp ( native wordexp found )
Disabled  zlib ( native zlib found )

--->>> vehicle
Disabled  android ( Default )
Enabled   demo ( Default )
Enabled   file ( Default )
Disabled  gpsd ( gpsd lib not found )
Disabled  gpsd_dbus ( dbus-glib-1 not found )
Disabled  gypsy ( gypsy lib not found )
Disabled  iphone ( Default )
Disabled  maemo ( Default )
Disabled  null ( Default )
Disabled  qt5 ( Qt5 libraries not found )
Disabled  wince ( Default )

--->>> Features
Disabled  BUILD_MAPTOOL ( maptool works only on 64 bit architectures )
Disabled  DBUS_USE_SYSTEM_BUS ( default )
Enabled   SAMPLE_MAP ( downloading binary map because maptool compilation is disabled )
Enabled   SVG2PNG ( default )
Enabled   USE_NATIVE_LANGUAGE_SUPPORT ( Gettext found )
Enabled   USE_PLUGINS ( default )
Enabled   USE_ROUTING ( default )
Enabled   USE_SVG ( default )
Disabled  XSL_PROCESSING ( Saxon missing )

CMakeLists.txt
```diff
@@ -257,11 +257,11 @@ else(ZLIB_FOUND)
    message(STATUS "using internal zlib")
    set_with_reason(support/zlib "native zlib missing" TRUE)
 endif(ZLIB_FOUND)
-if(OPENSSL_CRYPTO_LIBRARIES)
-   set(HAVE_LIBCRYPTO 1)
-   include_directories(${OPENSSL_INCLUDE_DIR})
-   list(APPEND NAVIT_LIBS ${OPENSSL_CRYPTO_LIBRARIES})
-endif()
+#if(OPENSSL_CRYPTO_LIBRARIES)
```
pgrandin: We should probably just remove the block instead of commenting it

jandegr: indeed

mdankov:
Hi!

We actually have a few uses for libcrypto. I think all of them could be found by analyzing HAVE_LIBCRYPTO macro usage in the code.

As far as I know, the one most important to us is ability to compute md5 sum of map file by maptool option -5 (--md5): it's used on our map server. Though I think we could easily switch there to an external command (with a little timing penalty).

Another option is ability of maptool to produce and ability of navit binfile plugin to use encrypted map files. I think it was used by cp15 at some point to protect some non-osm data. As for the maptool, it seems to be currently impossible to fill in the passwd member of struct zip_info in maptool/zip.c, so no encrypted file could be produced by maptool. On the other hand, binfile plugin of navit accepts map_pass attribute and it could be possible to use encrypted binfiles.

I'm unsure if encrypted binfiles are used these days by anybody.

If we are going to skip openssl dependency, we should drop all its references in the C code, not only at CMakeLists.txt.

jandegr:
Interesting comment from mdankov, thx.
I did a search in binfile.c for libcrypto before I submitted this PR and did not find it, the actual usage related to map reading is in file.c but I had not searched there.
[https://github.com/navit-gps/navit/blob/trunk/navit/file.c#L49](url)
And here is one in maptool, I initially neglected to search in maptool completely
https://github.com/navit-gps/navit/blob/trunk/navit/maptool/maptool.h#L26

It looks like libcrypto usage could be turned off by default and let it a compile option for specialist usage.

mdankov:
I had a chat with cp15. He asks us to keep support for encrypted binfiles. His suggestion is:

> For a quick solution I would add an option --enable-license-violation to cmake so who wants to use it knows it has a problem.

> Maybe in the long term we should include something public domain like https://github.com/ctz/cifra/tree/master/src except for the random stuff in maptool it seems it includes everything we use.

I already have disabled usage of maptool internal md5 calculation, so our map build process should not depend on maptool being linked to openssl. Being tested now.

jandegr:
hi,

>  For a quick solution I would add an option --enable-license-violation to cmake so who wants to use it 
> knows it has a problem.

that seems a good solution, I don't see anybody using that option soon but it would remain possible.

jkoan:
Probably we can also use this public license code for the md5 thing:
https://github.com/george-hopkins/opentom/blob/master/src/tools/ttimage_tools/md5.c

sleske:
> I had a chat with cp15. He asks us to keep support for encrypted binfiles. His suggestion is:
> For a quick solution I would add an option --enable-license-violation to cmake so who wants to use it knows it has a problem.

While I deeply respect cp15's work on Navit, I disagree here.
In my experience it's a bad idea to keep code around that is not actively used and cannot be tested.
I think we should rip out support for encrypted binfiles. If the support is needed later, we can always re-introduce it (that's why we use version control). Navit already has a lot of code that is unused and/or buggy, and that should be reduced.

> Maybe in the long term we should include something public domain like https://github.com/ctz/cifra/tree/master/src except for the random stuff in maptool it seems it includes everything we use.

Yes, _if_ we need crypto stuff, we will certainly find some implementation. The library [Nettle](https://en.wikipedia.org/wiki/Nettle_(cryptographic_library)) also looks good, and its license is compatible ( GPL2 and LGPL3).

> I already have disabled usage of maptool internal md5 calculation, so our map build process should not depend on maptool being linked to openssl. Being tested now.

If maptool does not need MD5 anymore, I'd take that part out, too (see above :-) ). Then we can drop OpenSSL right away, because AFAIK OpenSSL is not used elsewhere in Navit.

jandegr:
I agree for 100% with @sleske but the way the branches were archived on the old svn repo made it easier to search in legacy or old code, or I don't fully know how to use github yet :)

sleske:
> the way the branches were archived on the old svn repo made it easier to search in legacy or old code, or I don't fully know how to use github yet :)

That has nothing to do with GutHub really - just standard Git. If we remove the code, we will of course use a good commit message ("Remove OpenSSL dependency; this also removes MD5 checksumming and support for encrypted binfiles..."). You can easily full text search commit messages (and commit content), so you'll find the right commit in no time :-).

sleske:
Ok, if there are no other objections, I'll close this PR, and open a new one to remove OpenSSL as explained above.

pgrandin:
@sleske works for me, thanks for taking care of this!

sleske:
OpenSSL dependency removed with commit 0b6027ff53.

PR #213, opened 2017-04-03
Qt5 Graphics: Add QML support

metalstrolch:
This pull request is about adding QML (QtQuick) Support to graphics_qt5 plug in.
A QML based graphics back plug in allows:
 a) to easily add a QML based GUI
 b) to integrate navit to Jolla Sailfish OS once, since QWidgets library is banned in Sailfish's Harbour shop.
This adds two build flags:
USE_QWIDGET:BOOL --> Enable QWidgets support (autodetected)
USE_QML:BOOL --> Enable QML support (autodetected)
Both can be set at the same time.

This adds one attr to graphics section in navit.xml:
qt5_widget -> set to "qml" for QML display, set to "qwidget" for qwidget display. Must only be set if both methods are present at the same time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,15 +320,21 @@ if (QT_FOUND)
    endif()
 endif(QT_FOUND)
 if (Qt5Widgets_FOUND OR Qt5Quick_FOUND)
+   set(Qt5_ADDITIONAL_LIBRARIES "")
    if(Qt5Widgets_FOUND)
 	   set_with_reason(USE_QWIDGET "Qt5Widgets found" TRUE)
+	   if(USE_QWIDGET)
+	   	set(Qt5_ADDITIONAL_LIBRARIES ${Qt5_ADDITIONAL_LIBRARIES} ${Qt5Widgets_LIBRARIES})
+           endif()
    endif()
    if(Qt5Quick_FOUND)
 	   set_with_reason(USE_QML "Qt5Quick found" TRUE)
+	   if(USE_QML)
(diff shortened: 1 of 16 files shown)


metalstrolch:
What's the official documentation for attrs, as we should document graphics_qt5's attrs once this is merged?
Do we have a Navit coding style? Or is it OK to just process the Qt5 plugins with clang-format.

pgrandin:
Thanks @metalstrolch nice work!
I am really happy to see tangible progress towards QT5 and a new QML UI!

Your PR looks good to me, but maybe we would want to address this before merging :
https://1713-30791823-gh.circle-artifacts.com/1/tmp/circle-artifacts.UuE0TvX/logs_qt5_qml/stderr.txt
```
error:navit:convert_to_attrs:failed to create attribute 'v' with value '792'
QQmlApplicationEngine failed to load component 
qrc:///graphics_qt5.qml:-1 File not found
 ```
Regarding your questions about attributes, I think that this is the kind of thing that we should document in both doxygen and the wiki. Unfortunately the wiki is down right now...

metalstrolch:
This is fixed now. It builds (and runs) in CI. The Gifs look silly for both the QT5 GUIs test runs because qt5 seems to miss keyboard support. Since this was not introduced with this change I'll fix this in another PR.

metalstrolch:
Picked the wrong button... Of course still open.

pgrandin:
Thanks @metalstrolch really nice work here.
Looks good to me!

PR #214, opened 2017-04-05
graphics Qt5 keyboard

metalstrolch:
This pull request adds keyboard input for graphics_qt5 plugin


metalstrolch:
CI runs now and produces gifs for all tested graphics.

Btw: the key sequence in xdotools.sh seems to be outdated, as it produces strange output inside the gif for all graphics.

metalstrolch:
Closed (and merged) due to be a blocker for future work and has no effects outside graphics_qt5

PR #217, opened 2017-04-19
keep vehicle profile when deactivating vehicle

mcapdeville:
When deactivating vehicle, profile is choosen from "car" vehicle_profile or by
default the first vehicle_profile in list. So, by example,
car_avoid_tolls never work when manually selecting current position (and
deactivate vehicle).

Now, if there is a selected vehicle profile, we keep it as
vehicle_profile (not changing it).

Aditionnaly, in internal gui, selecting a vehicle profile when wehicle
is deactivated, in gui_internal_cmd_set_active_profile(), call navit_set_vehicleprofile_name() with new
vehicle_profile (by name) before setting vehicle_attr to NULL (like in
gui_internal_cmd_set_position()).

Now, we can switch profile when vehicle is deactivatied and route is
update correctly.

TO DO : Same in other gui
TO TEST : with more than one vehicle enable

### DIFF
--- a/navit/gui/internal/gui_internal.c
+++ b/navit/gui/internal/gui_internal.c
@@ -2042,6 +2042,7 @@ gui_internal_cmd_set_active_profile(struct gui_priv *this, struct
 	struct attr vehicle_name_attr;
 	char *vehicle_name = NULL;
 	struct attr profilename_attr;
+	struct attr vehicle;
 
 	// Get the vehicle name
 	vehicle_get_attr(v, attr_name, &vehicle_name_attr, NULL);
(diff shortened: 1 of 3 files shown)


navit/vehicleprofile.c
```diff
@@ -198,7 +198,9 @@ vehicleprofile_update(struct vehicleprofile *this_)
 {
 	struct attr_iter *iter=vehicleprofile_attr_iter_new();
 	struct attr profile_option;
-	dbg(lvl_debug,"enter\n");
```
metalstrolch: Hmm, do we really want those messages on error level? Suppose to change them to debug before merge.

navit/vehicleprofile.c
```diff
@@ -211,8 +213,8 @@ vehicleprofile_update(struct vehicleprofile *this_)
 			vehicleprofile_apply_attrs(this_, profile_option.u.navit_object, 1);
 	}
 	vehicleprofile_attr_iter_destroy(iter);
-	dbg(lvl_debug,"result l %d w %d h %d wg %d awg %d pen %d\n",this_->length,this_->width,this_->height,this_->weight,this_->axle_weight,this_->through_traffic_penalty);
```
metalstrolch: Hmm, do we really want those messages on error level? Suppose to change them to debug before merge.

pohlinkzei:
Hi, you might want to set the debug messages back to debug level.

pgrandin:
Nice, thanks @mcapdeville !

PR #218, opened 2017-04-19
:qnavigation_analyze_roundabout : central_angle may lead to division by 0

mcapdeville:
On the path from orange (france) to valence (france) by highway,
the third roundabout after highway exit lead to central_angle=0, leading
to divide by 0 (near line 2488).

from osm_id : 160627295 to osm_id : 4277896

(delta1 + delta2) seams to be 360, so I just set roundabout_length to len in that
case.

### DIFF
--- a/navit/navigation.c
+++ b/navit/navigation.c
@@ -2486,7 +2486,12 @@ void navigation_analyze_roundabout(struct navigation *this_, struct navigation_c
 			 * The central angle is approximated using the unweighted average of delta1 and delta2,
 			 * which is somewhat crude but sufficient for our purposes. */
 			central_angle = abs((delta1 + delta2) / 2 + ((cmd->delta < dtsir) ? 180 : -180));
-			roundabout_length = len * 360 / central_angle;
+			if (central_angle)
+				roundabout_length = len * 360 / central_angle;
+			else {
+				dbg(lvl_error,"central_angle in roundabout_length computation lead to divide by zero (delta1 = %d, delta2 = %d, cmd->delta  = %d, dtsir = %d, len = %d)",delta1,delta2,cmd->delta,dtsir,len);
+				roundabout_length = len;
+			}
 			dbg(lvl_debug,"roundabout_length = %dm (for central_angle = %d degrees)\n", roundabout_length, central_angle);
 
 			/* in the case of separate carriageways, approach roads become hard to identify, thus we keep a cap on distance.


pgrandin:
This one is a bit more tricky.
Before merging we need to check the results of this change against our routing test cases : https://github.com/navit-gps/routing-qa/tree/roundabout-exit

We have a stub of reporting here : http://navit-gps.github.io/routing-qa/ automatically built from the CI that we should to improve to make it more usable.

I'll look into altering our CI code so that it can run tests from another repo : currently we only tested different branches of our own repo, and here we'd need to check against your fork.

Thanks for the PR!

metalstrolch:
I agree that this should be checked against a automatic qa system. But sice this seems to reqire care before it can be run i propose stepping back to good old code review.

In my opinion, this change cannot degrade any behaviour, because everything it does is preventing div by zero crash. I would opt for adding a err log in case div by zero would have happened and then merge it.

pgrandin:
Good call @metalstrolch this works for me.

mcapdeville:
I have added error log.
In commit text,I have also added coordinate for test.

jandegr:
Is this a correct path to test this case ?
https://297-40190699-gh.circle-artifacts.com/0/tmp/circle-artifacts.0s5S6vA/Valence_roundabout.yamlmetric.png

mcapdeville:
Yes, this is correct. But the problem is only on the second roundabout after highway exit. Coordinate are in the last commit.

jandegr:
Strange, I fail to duplicate the failure.
Did anybody else succeed in duplicating it ?
Last OSM edits are 4 to 6 years ago, so even if we're using a different map version that would not explain the difference.

metalstrolch:
Was not able to replicate this yet. What brings me to two conclusions:
1st, I'd add this patch because possible div by zero crashes are bad.
2nd, fix routing qa so we could add this route as a test.
unfortunately I don't have a clue how to do 2nd.

mcapdeville:
Effectively, Using only one map : rhonealpes_3.6,44.1,7.9,46.6.bin don't show the problem.
But having this list of map show it : 
aquitaine_-2.3,42.4,1.5,45.8.bin
auvergne_2.0,44.6,4.5,46.9.bin
bourgogne_2.8,46.1,5.6,48.5.bin
centre_0.0,46.3,3.2,49.0.bin
cotedazur_4.2,42.9,8.0,45.2.bin
franchecomte_5.2,46.2,7.8,48.1.bin
languedoc_1.5,42.2,4.9,45.0.bin
midipyrenee_-0.4,42.2,3.5,45.1.bin
paysdelaloire_-2.9,46.2,1.0,48.6.bin
poitoucharentes_-1.7,45.0,1.3,47.2.bin
rhonealpes_3.6,44.1,7.9,46.6.bin
swiss_5.8,45.7,10.6,47.8.bin
Perhaps because of overlapping maps ?

mcapdeville:
Having only 2 map with this roundabout lead to the probleme : 
rhonealpes_3.6,44.1,7.9,46.6.bin and cotedazur_4.2,42.9,8.0,45.2.bin
or cotedazur_4.2,42.9,8.0,45.2.bin and languedoc_1.5,42.2,4.9,45.0.bin
or languedoc_1.5,42.2,4.9,45.0.bin and rhonealpes_3.6,44.1,7.9,46.6.bin
but not when only one of these are present.
So I think there is another problem elsewhere.

jandegr:
That is a usefull answer, pointing to a duplicate check failure.
Adding a division by zero check is ok, but without your latest info the real issue would never ever have been searched for.
Using overlapping maps has several sideeffects and in case the maps did not come from the same 'mother-planet.bin' it would even be worse.
All of France is only slightly over 3GB, you should be able to install this on any modern device

mcapdeville:
All my maps are downloaded the same day, so I make this test : 
I use one file rhonealpes_3.6,44.1,7.9,46.6.bin and hard link it to rhonealpes_bis_3.6,44.1,7.9,46.6.bin.
This produce the divide by zero error in roundabout_length computation.

PR #220, opened 2017-04-24
Desktop icons in more sizes

metalstrolch:
I recreated the navit desktop icon(s) from the navit.svg present in order to get them for multiple resolutions as required by Sailfish OS.

As this won't hurt others, we can add this to mainline.
P.S. I don't know why the icons up to now were scaled slightly different. Please comment if this is an issue.


zintor:
I agree with metalstrolch, that the icons are needed for Sailfish OS and it shouldn't hurt other.
But this brings me to other OS icon sizes, which are possible:
-------
Windows: 16x16, 24x24, 32x32, 48x48, 256x256
Mac OS X: 16x16, 32x32, 64x64, 128x128, 256x256, 512x512, 1024x1024
Linux: 16x16, 24x24, 48x48, and 96x96
iOS 6: 29x29, 50x50, 57x57, 58x58, 72x72, 100x100, 114x114, 144x144, 1024x1024
iOS 8: 29x29, 40x40, 58x58, 60x60, 76x76, 80x80, 120x120, 152x152, 180x180, 1024x1024
Android L: 24x24, 48x48, 192x192, 512x512
Windows Phone: 62x62, 99x99, 173x173, 200x200
-------
Should/Must we care about these?

Another question: Should some OSs-stylished icons be added?
For example SailfishOS desktop icons looks different than Windows or Android. And so on.

metalstrolch:
Until now nobody seemed to care. Imho, I would prefer generating those icons from the svg as required, but that is only possible if some svg tool is present. Adding "all" sizes will at least blow up the repo.
At least I would not add distribution special icons there. That should go under contrib.

We could agree putting those added here under contrib as well. But as said presumely nobody cared. So I propose not to care either and keep all the eventualities up until someone wants to create packages requiring them.

pgrandin:
Could we generate the icons when building the package instead?

metalstrolch:
Yes we could. The same way we do the map icons. But this will make those icons dependant of any svg processing tool on compile time. With Sailfish, this could be problematic

sleske:
@metalstrolch 

> Yes we could. The same way we do the map icons. But this will make those icons dependant of any svg processing tool on compile time. With Sailfish, this could be problematic.

Why would that be problematic? The build already depends on an SVG processing tool (otherwise you will have next to no icons). Why can't we just re-use this functionality for generating the desktop icons?

I would like to avoid redundantly including PNG files in the repository - not only because of repo size, but also because this is hard to maintain if the icon is changed/corrected later.

metalstrolch:
No, the sailfish build uses svg directly. No need for svg tool and png icons. Dont know if this is available in MER build vm.

metalstrolch:
As there was no more conversation (and this should really not hurt anybody) I merge this to get clean state for sailfish build.

PR #221, opened 2017-04-25
Add rules to build sailfish package

metalstrolch:
This patch adds build scripts for building a (almost) Sailfish OS Harbour compatible package for Sailfish OS. With this, Sailfish OS package can be built from vanilla upstream Navit.

This patch includes:
- Build script for building under Sailfish OS MER VM for Phones and Emulator.
- RPM spec file to cope for the special requirements for Harbour
- Sample configuration with simple OSD and size configuration for Jolla Phone screens.
- Example shell trampoline to call espeak for voice output if espeak is installed.


metalstrolch:
@zintor If you want to start thinking about a QML 2.0 GUI, have a look at the qt5_qml_gui branch.

pgrandin:
Nice, thanks @metalstrolch !

PR #225, opened 2017-04-27
[Android] Enable autozoom by default

jfitie:
Enable autozoom by default so that car navigation automatically zooms in on the vehicle based on current speed. Very useful when navigating.

### DIFF
--- a/navit/xslt/android.xslt
+++ b/navit/xslt/android.xslt
@@ -49,6 +49,7 @@
       <xsl:copy>
          <xsl:copy-of select="@*"/>
          <xsl:attribute name="zoom">32</xsl:attribute>
+         <xsl:attribute name="autozoom_active">1</xsl:attribute>
          <xsl:attribute name="timeout">86400</xsl:attribute>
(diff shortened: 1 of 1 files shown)


pgrandin:
It works for me. @jandegr any opinion on this?

jfitie:
Thanks for considering. Enabling this option by default gives the end user a much better out-of-box experience.

jandegr:
Without some way to change it this discussion is just a matter of taste.
With smth. like below it can be made a user setting, leaving what the default
value should be open for further discussion.

> 

case R.id.toggle_autozoom:
				msg = Message.obtain(N_NavitGraphics.callback_handler, NavitGraphics.msg_type.CLB_CALL_CMD.ordinal());
				b = new Bundle();
				b.putString("cmd", "autozoom_active=autozoom_active==0?1:0");
				msg.setData(b);
				msg.sendToTarget();
				break;

jfitie:
https://github.com/navit-gps/navit/pull/231/files

PR #226, opened 2017-04-27
[Android] 'Toggle POIs' button also toggle POI labels

jfitie:
http://trac.navit-project.org/ticket/1071

The menu-button Toggle POIs in the Android port toggles the POI Symbols layer. It makes sense to have it also toggle the related POI Labels layer. Attached a patch to make this happen.


pgrandin:
Thanks @jfitie for not giving up on this patch!

PR #228, opened 2017-04-28
Add network info menu in gui/internal for Linux

pgrandin:
This PR adds a simple information menu showing the available network interfaces and their respective IP addresses.

This is especially useful when using Linux-based touchscreen devices.

network-info


pgrandin:
Thanks guys!

@metalstrolch : thanks for testing this on Sailfish! Translations should be handled via https://github.com/navit-gps/navit/blob/239c1a9796cd112e882b3424be35c78fad399143/navit/gui/internal/gui_internal_command.c#L801 

This item will be added to the template navit.pot and pushed to launchpad automatically when the code will be merged into trunk. Currently our translation workflow does not deal with branches (this would allow working on translations before merging in trunk) but we could maybe improve that.

@pohlinkzei thanks for your help!

PR #229, opened 2017-04-30
Fix trac1347.

pohlinkzei:
Add 'Follow' and 'Active' to vehicle_demo to remove the corresponding error messages (Unsupported Attribute)

[trac1347](http://trac.navit-project.org/ticket/1347#no1)

The Demo_Vehicle complains about its unsupported attributes follow and active. These Attributes arent really unsupported but processes inside navit.c for all vehicles. Since these Attributes are processes globally we might want to ignore them inside vehicle_demo.c.

So this will be acheived by this PR.


pgrandin:
Thanks @pohlinkzei !

PR #232, opened 2017-05-03
Fix: Don't crash if LANG environment not set

metalstrolch:
Since getenv may return NULL if a variable is not set, current keyboard
code crashes if LANG environment is not set.


pgrandin:
A simple fix to an old problem!
Thanks @metalstrolch !

PR #233, opened 2017-05-03
Add: Add speech module using espeak on QMultimedia.

metalstrolch:
This pull request adds a new speech module using espeak on QMultimedia back end. It builds either against autodetected system installed libespeak, or with the internal (heavily modified) espeak if no external libespeak was found.
This allows to have speech output on Qt5 systems without further external requirement. Espeak unfortunately is seldom provided by the distributions.

Note: This implemention is not multithreaded. It may therefore come to chopping on weak systems as drawing the screen locks the Qt evet loop.
But it works quite well at least on Jolla smart phone.


CMakeLists.txt
```diff
@@ -563,6 +580,9 @@ if(WIN32 OR WINCE)
    SET(CMAKE_RC_COMPILE_OBJECT "<CMAKE_RC_COMPILER> -O coff -I ${CMAKE_CURRENT_SOURCE_DIR}/navit/gui/win32/resources <DEFINES> -i <SOURCE> -o <OBJECT>")
    set_with_reason(support/ezxml "Windows detected" TRUE)
    set_with_reason(speech/espeak "Windows detected" TRUE)
+   #speech/espeak does use windows threads, and therefore needs only parts of
+   #espeak lib. INTERNAL_ESPEAK controls this.
+   set_with_reason(INTERNAL_ESPEAK "Windows detected" FALSE)
```
pgrandin: Looks like a copy/paste issue :)

metalstrolch: No this is OK. INTERNAL_ESPEAK controls if support/espeak builds the complete libespeak or just parts of it. And the (old) Windows version does only want parts of it, so this is set to FALSE.
Think I'll rename INTERNAL_ESPEAK to better reflect this :)

metalstrolch:
What about this? Any reason not to merge?

pgrandin:
Sorry - missed your last comment. Looks good now, thanks!

PR #231, opened 2017-05-03
Add auto zoom toggle

jfitie:
https://github.com/navit-gps/navit/pull/225

PR #237, opened 2017-05-05
Auto manual zoom icons

jfitie:
https://github.com/navit-gps/navit/pull/231#issuecomment-299352860


metalstrolch:
Would suggest keeping the source branch in the future, as this eases up checking

pgrandin:
On an somewhat unrelated topic, @jfitie your patches appear to be against "unkown repository" which doesn't make CI testing easy. Do you prepare your patches against a fork?

jfitie:
@pgrandin I was too fast in removing my branch/fork :-)

PR #242, opened 2017-05-08
[POI] Improve tower.xpm icon

jfitie:


jandegr:
pls read a bit in http://irclogs.navit.ie/%23navit-2017-05-09.log for xpm versus png
If I understood correctly svg is preferred, so when changing / replacing an old xpm, pls drop the xpm in favor of an svg. -- or join IRC to discuss it --

PR #243, opened 2017-05-08
[POI] Improve exit.svg icon

jfitie:


sleske:
Hi, similary to #245 , this change unfortunately means the icon will no longer work well on a dark background. Could you fix this?

PR #245, opened 2017-05-08
[POI] Improve fuel.svg icon

jfitie:


sleske:
Thanks for improving the icons! As it happens, I'm working on the icons at the moment, too (will push soon).

While your change is a good idea, it unfortunately means the icon will no longer work well on a dark background (which the Internal GUI uses by default). Could you fix this? Either by changing the colors, or by adding a light outline.

PR #240, opened 2017-05-06
Fix Car layout issues with wood and water

jfitie:
Fixed some issues in ```Car``` layout where wood (```poly_wood```) was not showing up correctly inside towns (```poly_town```) and water (```poly_water```) was missing in parks (```poly_park```). Also fixed some discrepancy issues in the ```Car-dark``` layout.

PR #247, opened 2017-05-08
Add grass, grassland, wetland and sand

jfitie:
As suggested by @jandegr in https://github.com/navit-gps/navit/pull/240#issuecomment-299688984


jfitie:
@jandegr Are these changes pushed to `maps.navit-project.org` automatically?

jandegr:
yes, but it takes a while untill a new map with the changes is available

PR #251, opened 2017-05-09
Add feature poly_reservoir

jfitie:
https://www.openstreetmap.org/#map=19/52.95293/6.30800


jfitie:
@jandegr Excellent!

screenshot_2017-05-09-15-40-47

PR #253, opened 2017-05-11
Fix: Qt5 fonts

metalstrolch:
Fix native Qt5 font support to get rid of mandatory freetype dependency.


metalstrolch:
Does anybody know:
Which unit is the font size passed to font_new?
How exactly expects navit the bbox outline to be?
What part of the font (baseline / bbox) does the coordinate in draw_text point to?

pgrandin:
Looks good to me. Thanks @metalstrolch 

I don't know the answers to your questions unfortunately. If you find out, please add some documentation to the code!

PR #254, opened 2017-05-13
QT5/qml2 UI proof of concept

pgrandin:
out-1

Based on @metalstrolch 's excellent work on qt5, I built a quick qml2 UI proof of concept.
This is still far from being a usable UI, but it should help us get started.

I'll put some notes in the wiki about this.

Comments and contributions are welcome!


metalstrolch:
Btw. What kind of editor did you use for this? It applied a completely different coding style which made reviewing quite hard.

pgrandin:
Thanks for your review @metalstrolch 

This won't work well in portrait for now indeed. I think that we probably need to design something close but specific for that, otherwise we probably won't use all the display correctly, or some things might look odd.

I've removed the dependency on QtQuicks.Control in https://github.com/navit-gps/navit/pull/254/commits/e46950f9d56cef06795320e33e61ac22cbc1e160 . It was from a specific test, and I think it'd be better to keep this as cross-platform as possible (especially for the first skin we'll build).

I'm using vim mostly (and qtcreator to design some of the qmls). I believe that my config should match our current [coding style standard](http://wiki.navit-project.org/index.php/Programming_guidelines#Coding_Style) which states 8 spaces for tabs. Your clang header does not :)

I personally prefer 4 spaces instead of 8 too, but that's out of the topic of this PR, we can discuss that elsewhere.

metalstrolch:
Finally with the latest change i was able to test on Sailfish, and with my latest changes in https://github.com/navit-gps/navit/pull/255 I could even see the new QML GUI rotate. It looks not as bad as I had imagined.
One thing makes me think, Why is the map view garbled on start up. It gets OK once you drag the map. I think there is a problem in resize event propagation (You can see this on your initial GIF as well). Maybe a bug in my QML backend?

pgrandin:
Oh it isn't as bad as expected in portrait, indeed. 
I think that we should try toggling the number of rows in portrait from 2 to 1, it should use the space better.

Regarding the map, I believe that I have seen a similar behavior using another gui. This is probably a refresh or draw event that's missing.

pgrandin:
Merging the changes, but keeping the branch for for further work.

PR #258, opened 2017-05-24
Add some more common landuse polygons

jfitie:
from https://wiki.openstreetmap.org/wiki/Map_Features#Landuse


jandegr:
hmm, I thik I was a bit too fast to merge.
Probably this needs a few additions as well https://github.com/navit-gps/navit/blob/trunk/navit/item_def.h
CircleCi does not seem to catch that.

jfitie:
@jandegr please check https://github.com/navit-gps/navit/pull/260

PR #260, opened 2017-05-25
Header for more common landuse polygons #258

jfitie:
for https://github.com/navit-gps/navit/pull/258


jandegr:
Hi,
They must be appended at the end, not inserted where it seems logical to place them, unfortunatly.
Otherwise older maps don't work anymore.
I reverted #258 after I merged it, so a new PR will have to be complete including #258 I suppose.
For changes like this it is always usefull to build a testmap.

PR #263, opened 2017-05-26
Qt5 qml gui - POC update

pgrandin:
Some updates on the QML2 UI.
Not yet fully functional.


jandegr:
Hi,
I think it is good practice for GUI's to call navit_block when they cover up the mapview, no need to burn fuel uselessly and all the rest of navit still goes on, that just defers updates of the mapview.

The Lidl poi preview looks very good and the blocking can be lifted just for the time to obtain the poi view.

For the rest verice nice work all of this QML gui.

pgrandin:
Thanks for the review @jandegr 

https://github.com/navit-gps/navit/pull/263/commits/7c2c2b9b8113530665e5bcd88e7082b9da62dc1f should address blocking draw operations.

PR #264, opened 2017-05-30
Add Zoom to route button to Route section in menu

jfitie:


navit/navit_shipped.xml
```diff
@@ -97,6 +97,7 @@ Navigation</text></img>
 				<img src='gui_tools' onclick='network_info()'><text>Network info</text></img>
 			</a>
 			<a name='Route'><text>Route</text>
+				<img src='gui_zoom_route' onclick='zoom_to_route()'><text>Zoom to route</text></img>
```
pgrandin: I think that this should be navit.zoom_to_route().
Also, we should probably close the menu after zooming on the route, to display the route.

jfitie: Agreed; https://github.com/navit-gps/navit/pull/264/commits/524aa23bc7aed3f7a31f5d45b16c1325c5594da0

jfitie:
Without this PR:
screenshot_2017-05-31-19-17-36

With this PR, no route set:
screenshot_2017-05-31-19-20-28

With this PR, route set:
screenshot_2017-05-31-19-20-43

PR #266, opened 2017-06-07
fix: graphics_qt5 fix layer drawing and dragging

metalstrolch:
Now cursor is drawn right with qml and qwidgets


metalstrolch:
Seems to be summer time. No comments on this, so I merge it since it is required for a new sailfish release and just touches qt5 graphics.

PR #268, opened 2017-06-19
Rework on Cmake/Cpack System

jkoan:
As discussed in IRC here are my work from the last days.

Of course it needs future work, but it is working and generating deb and rpm Packages. Also NSIS Packages will follow

Sorry for the commit spam :D


metalstrolch:
Builds on sailfish. Nice.

sleske:
Hi, this looks interesting. However, the work seems to be in a very early state - I could not get the DEB creation to work, this PR deletes the Navit README, there is incomplete code for NSIS,  the DEB settings use "predepends" instead of "depends", etc.

Also, I'm not quite sure whether it's a good idea to set up our own infrastructure for creating Deb and RPM packages. Navit is already packaged by Debian, so building DEB generation into the Navit sources will duplicate that work (and possibly produce Debs that are incompatible with Debian).

I think we should clarify our goals before going further. Then we can see what can be included in Navit.

At any rate, this needs a cleanup (rebase) before it can be merged. I'll be glad to help with that, once it's clear where we want to go.

mvglasow:
If I may give my 2 cents on that… I’ve come across a bunch of projects which require a lot of fiddling to install as they’re not in the Debian repos, or the Debian repos have a hopelessly outdated version, often because Debian maintains debs without official support from the project. As a result, I end up littering my system configuration with PPAs from different projects, or even more invasive under-the-hood fiddling (`sudo make install` and the like), which I try to avoid where I can.

In a perfect world (from a user point of view), the OS distribution of my choice would always have the latest stable version of every package, limited only by the update policy of the distro (which might choose not to do major version upgrades on an already released version). What is preventing this from happening is usually lack of cooperation between projects and distros.

Therefore, if we managed to package Navit in a way that the major distros would just need to take the package, either in binary form or building it with the official Navit recipe, and ship it with their release, that would be a plus. My experience (with F-Droid) is that repo maintainers are usually happy about projects approaching them.

Since Debian is already packaging Navit, it’s worth looking at what they’re doing and reusing as much of their work as possible. If the Debian maintainer is not already a Navit contributor, maybe he/she’d be happy to become one.

We’d want to keep the number of distros low to avoid things getting out of hand. Sticking with a deb (built for and tested with Debian) and a rpm (built for and tested with RedHat/CentOS) should give us the broadest coverage.

As for ensuring compatibility—we already have a CI environment, how much work would it be to add basic packaging tests? (At a minimum, building the package, making sure it installs and that Navit starts without crashing afterwards; further tests with the installed product are optional.)

jkoan:
Ifrst of all i want to say thanke you for your reviews of that work. As I said in my initial Post this is verry much WIP so i will try to fix problems and make this PR better. ;)

> Hi, this looks interesting. However, the work seems to be in a very early state

Yes this is right, probably I should add [WIP] to the title ;)
> I could not get the DEB creation to work

It would be nice if you could provide future details to this? For me its working with:
`mkdir build && cd build && cmake  [OPTIONS] .. && make && make package`
> this PR deletes the Navit README, there is incomplete code for NSIS, the DEB settings use "predepends" instead of "depends", etc. 

README ) Yeah it is right that it deletes **A**  README, but this is the old one. I mad a diff of README.md and README and README.md is the new file.
NSIS ) I dont see why this is not "complete" for NSIS it is the same code than before, also the code is working. But i have only tested with wine.
DEB ) Yes, useing predepends was not intended (will fix soon)

---
> Since Debian is already packaging Navit, it’s worth looking at what they’re doing and reusing as much of their work as possible. 

Yes, I agree we should look at https://packages.debian.org/stretch/navit (on the right there are the source packages which have for example also paches)

> As for ensuring compatibility—we already have a CI environment, how much work would it be to add basic packaging tests? (At a minimum, building the package, making sure it installs and that Navit starts without crashing afterwards; further tests with the installed product are optional.)

To ensure that the package is building is not difficult, because as said before in this post it is only a "make package" at the end. Testing if the package itself is working is a little bit more difficult because the CI only has Ubuntu, so we can test deb of course. Probably a workaround would be unpacking the RPM in a path and run it from there, but this is not very nice.


Jkoan

jkoan:
I will clean up and open new PR soon.

PR #271, opened 2017-06-21
Sailfish config

metalstrolch:
Create configuration for sailfish os by using xslt files to do changes to navit_shipped.xml


metalstrolch:
Merging this as this only affects Sailfish builds

PR #273, opened 2017-06-24
Use all available Cores for Ci

jkoan:
Ci has 32 Cores, not 4 Which is reportet from nproc.
So build should be faster ;)


pgrandin:
Did you see a difference in build time in your fork?

jkoan:
Yes, see on ci: 13:36min with this pull request and ~15min without. That's not so much, but it is a little bit. 

If I remember right the most time is taken to convert all those svgs, probably we could optimize this also ;) But this is not the pull right now.

pgrandin:
Sweet thanks!

PR #274, opened 2017-06-24
Use datecode insted of Tags as Android id

jkoan:
As discussed in IRC here are the new version for the the Android ID so we can delete the old R0000 tags.


pgrandin:
Reference ticket : http://trac.navit-project.org/ticket/1351

PR #275, opened 2017-07-07
[WIP] Framerate boost and EXPERIMENTAL map moving

santeri3700:
**DO NOT MERGE WITHOUT FIXING OR REMOVING THE EXPERIMENTAL PART!**
I discovered a way to increase frame rate by removing a delay on line 598 of navit.c.
There is also another VERY experimental change which makes
map moving better (or at least some part of it)
The map will be loaded while it's being moved **BUT**
there is a problem with the movement.
It is basically done incorrectly and needs to be fixed by a real developer.
See lines: 585-588

KaZeR suggested me to send this even if not completely working.

Tested on Ubuntu 17.04 x86_64 with SDL.

EDIT:
Here is a short video: https://youtu.be/9O2qjOHx3SA

### DIFF
--- a/navit/navit.c
+++ b/navit/navit.c
@@ -582,6 +582,10 @@ navit_handle_motion(struct navit *this_, struct point *p)
 		dx=(p->x-this_->pressed.x);
 		dy=(p->y-this_->pressed.y);
 		if (dx < -8 || dx > 8 || dy < -8 || dy > 8) {
+			//FIXME: Super experimental! Basically makes map moving very hard.
+			//TODO: Fix me
+			update_transformation(this_->trans, &this_->pressed, p);
+			navit_draw(this_);
 			this_->moved=1;
 			if (this_->button_timeout) {
 				event_remove_timeout(this_->button_timeout);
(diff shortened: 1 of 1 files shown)


sleske:
Hi, and welcome to Navit!

About the changes:

Your first change (added lines) makes Navit redraw the map on each drag event, instead of only after a timout has elapsed (see lines below, which you also changed). This just needlessly burns CPU, and without bitmap dragging (drag_bitmap="1" in navit.xml) will also cause Navit to hang/become unresponsive during dragging, because it cannot keep up with redrawing (test with a big map!).

Your second change (removing the 10/100ms timeout) will _also_ make Navit redraw the map on each drag event (again!), instead of every 10ms (if bitmap dragging is on) or every 100 ms (if it is not). This has the same effect as the first change.

I'll close this pull request - this change is a good experiment, but not something we can include in Navit.
This kind of change is a good way to get started understanding the code base. Try experimenting some more, or have a look at some of our open bugs on http://trac.navit-project.org/. We can always use more developers!
s

PR #276, opened 2017-07-08
Fix download link for libpng

jkoan:


jandegr:
I don't really care how it is done but in the past it took just a small edit to bump libpng every now and then.
This fix is a freeze in Navit style and how it became to depend on some badly outdated external code, sometimes even by merging into the repo years ago.
Unfortunatly for libpng I did not find a way to get 'latest' somewhere so these bumps could go automatic.

jkoan:
yes @jandegr  i also see that is not the best solution to have the fixed version in the build script, probably we can later switch to better sources, but for now it seems that this fixes our problem that we are not able to build tomtom things on ci anywhere.

pgrandin:
Well we could host a copy of the dependencies we need : i actually did it for the toolchain https://github.com/navit-gps/dependencies/tree/master/tomtom 

@jkoan : would you mind to update the PR so that it uses the mentioned repo ?

PR #279, opened 2017-07-13
Set default pitch to 5 for Android build

jfitie:
The default pitch of 20 is too high for Android devices which usually have 16:9 or 16:10 screens. This results in some distortion and unused screen space. 5 seems to be a better default value with less unused screen space and a nice 3D look (and it is still a 'nice' number). I've tested this on 7 devices.

Example:

Pitch 20:
pitch20

Pitch 5:
pitch5


jfitie:
Continued in https://github.com/navit-gps/navit/pull/303

PR #284, opened 2017-07-13
label for heightlines

jandegr:


zintor:
I could approve, that labels are shown on heightlines

PR #294, opened 2017-07-18
Fixes displaying the heightline route info

pgrandin:
heightline

@zintor can you please confirm that it fixes your issue?


zintor:
I can confirm, that I am able to see the heightlines.
Debugging informations are also shown.

PR #299, opened 2017-07-20
[POI] Improve Public office icon

jfitie:


sleske:
Hi, thanks for cleaning up our old icons.
Unfortunately, several of the SVG icons you created/modified for POIs have a width/height of 100.
However, the (somewhat complex) scaling code in the Navit build requires POI icons to have a width/height of 22. Note that all the SVG files for POIs have these dimensions.
We should probably change the size of the new SVGs by changing "width", "height" and "viewBox". I'll try to do this in the next days.

jfitie:
@sleske Sorry about that, they seemed to work fine on the Android build I use on multiple devices. I've changed this for my (and also some other) icons in PR https://github.com/navit-gps/navit/pull/304

sleske:
@jfitie Thanks for the quick correction. Yes, Android is not affected, because icons are scaled differently for Android (don't ask why).
There were two more icons with size 100x100 (tower and traffic_signal), I fixed those too. Now all have the right size.

PR #303, opened 2017-07-26
Pitch correction

jfitie:
The default 3D pitch of 20 is too high for recent (Android) devices which usually have high resolution screens. This results in a lot of distortion and unused screen space.

Example:

Pitch 20:
pitch20

Pitch needs to be corrected for window resolution. After studying the transformation code, I found the following solution.
In pseudocode: ```p=ROUND(20*SQRT(320*240)/SQRT(x*y))```

The values 20 (default 3D pitch), 320 and 240 come from this screenshot, which is the first one for the internal GUI on the official site:
HTC Wildfire


jfitie:
Here are some examples of random resulutions where the pitch has been corrected using the pseudocode shown above.

In Bash (excerpt):
```
p=`python -c 'import math;print int(round(20*math.sqrt(320*240)/math.sqrt('$x'*'$y'),0))'` # pitch // ROUND(20*SQRT(320*240)/SQRT(x*y))
sed -i -e 's@navit center="11.5666 48.1333"@navit center="3752 N 12216 W" pitch="'$p'"@' navit.xml # set location and pitch
sed -i -e 's@graphics type="win32"@graphics type="win32" w="'$rx'" h="'$ry'"@' navit.xml # set resolution
```

Resolution 240 x 1280; calculated pitch: 10


Resolution 240 x 1920; calculated pitch: 8


Resolution 320 x 400; calculated pitch: 15


Resolution 320 x 480; calculated pitch: 14


Resolution 320 x 1080; calculated pitch: 9


Resolution 320 x 1600; calculated pitch: 8


Resolution 320 x 2160; calculated pitch: 7


Resolution 400 x 854; calculated pitch: 9


Resolution 400 x 1920; calculated pitch: 6


Resolution 720 x 800; calculated pitch: 7


Resolution 720 x 1280; calculated pitch: 6


Resolution 720 x 1920; calculated pitch: 5


Resolution 800 x 400; calculated pitch: 10


Resolution 800 x 1080; calculated pitch: 6


Resolution 800 x 1440; calculated pitch: 5


Resolution 800 x 1600; calculated pitch: 5


Resolution 854 x 240; calculated pitch: 12


Resolution 854 x 1600; calculated pitch: 5


Resolution 1080 x 320; calculated pitch: 9


Resolution 1280 x 720; calculated pitch: 6


Resolution 1280 x 1440; calculated pitch: 4


Resolution 1440 x 320; calculated pitch: 8


Resolution 1440 x 800; calculated pitch: 5


Resolution 2160 x 320; calculated pitch: 7


Resolution 2160 x 1280; calculated pitch: 3


Resolution 2560 x 240; calculated pitch: 7


Resolution 2560 x 400; calculated pitch: 5


Resolution 2560 x 1440; calculated pitch: 3

jfitie:
Almost, but not quite @ https://github.com/navit-gps/navit/pull/303/commits/afe1e1e7281591febde553f99eff919c3c159582

jfitie:
According to the (new) log, the calculation is performed before the window resolution is set.
```debug:gui_internal:gui_internal_new:w=0 h=0 pitch=-2147483648```

pohlinkzei:
I havent understood how this is intended to work, but your calculation does not calculate what I assume it should.
pitch*sqrt(DEFRESX*DEFRESY)/sqrt(t->w*t->h) is always pitch because t->w*t->h = DEFRESX*DEFRESY.

Your Debug:
debug:navit:transform_setup_matrix:yaw=0 pitch=20 center=0x11ba31,0x5b5e2d w=240 h=320

jfitie:
Commit 535f5e107783ae1a1496a1f3f7d7ce1a8aae7e38 results for different resolutions aspect/ratios:

jfitie:
@pohlinkzei Correct, but I didn't know how to pass the current window width and height yet, so they were hardcoded still at the time of your reply. That's what ```[WIP]``` in the pr title stands for :smile:

jfitie:
Results of the Android build with the latest code; as expected with a visual pitch similar to the screenshots above and nothing like the distorted first image at the top:

screenshot_20170727-183438

screenshot_20170727-183447

jfitie:
@pohlinkzei 

- True, but since the calculation is not done frequently, it makes sense to keep it like this instead of replacing it with just some number.
- Resize: was talking about this with @pgrandin, but we couldn't get the UI to resize and this could not be tested. However, according to the debug log, the pitch is set outside of the initialization (```debug:navit:navit_set_attr_do:enter, this_=0x2475e8, attr=0xaae740 (pitch), init=0```). As for changes in orientation: this doesn't matter as the correction is based on the total number of pixels, not the aspect ration. See the 2 (Android) images just above your comment.
- Not sure what you mean, but the pitch can be set to a single level, like the default of 20. It will just be corrected based on the window size so that a pitch of 20 will look the same on any device/resolution.

jfitie:
Rebase

pgrandin:
Kudos on the thorough testing!

PR #305, opened 2017-07-31
Maptool multicountry towns

mdankov:
This branch allows map to include towns to more than one country. There are a few regions in the world mapped in such a manner, and stock maptool leaves only one of the countries active, randomly ignoring the rest.

We have a planet built with this branch, it's available for download at http://maps6.exp.navit-project.org/


pgrandin:
Nice feature, thanks!

PR #306, opened 2017-07-31
Auto pitch and orientation + Apply pitch correction in reverse when reading attribute

jfitie:
Suggestion for improving the out-of-box experience for builds using ```osd_minimum.xslt``` (Android and iPhone).

Depending on the ```autozoom_active``` setting (which is enabled by default on Android; https://github.com/navit-gps/navit/blob/trunk/navit/xslt/android.xslt#L52) the following paramers are set automatically:

- ```pitch``` (```pitch=autozoom_active==0?pitch:(follow>1?0:20)```)
- ```orientation``` (```orientation=autozoom_active==0?orientation:(follow>1?0:-1)```)

And a new OSD element is added:

- _Zoom auto_ or _Zoom manual_ (depending on active setting/```autozoom_active```)

Initial startup: **auto mode; driving/navigating**:
screenshot_20170731-200546
Because autozoom is active by default, this sets the ```pitch``` to a value of 20 for a 3D look. ```orientation``` is set to -1 so the map rotates in the direction of travel.
**Edit:** Zoom in and out buttons are now also shown in this mode since https://github.com/navit-gps/navit/pull/306/commits/ba2b45475842d43cf6ac60039a727b4c1c456fa4.

Dragging the map: **auto mode; browsing**:
screenshot_20170731-200601
By moving the map, ```orientation``` is automatically set to 0 to force a North-up view. ```pitch``` is set to 0 so the map is shown top-down. OSD Zoom buttons are shown, as the user might want to manually zoom in and out from this view.

Select manual zoom (M icon):
screenshot_20170731-200619
Enabling manual zoom mode always shows the OSD zoom buttons and allows ```pitch``` to be toggled using the standard (2D/3D) buttons in the menu. ```orientation``` is unchanged.

Select automatic zoom (A icon) and follow cursor:
screenshot_20170731-200631
Enabling automatic zoom and centering on the cursor again, ```pitch``` is set to 20 and ```orientation``` is set to -1.
**Edit:** Zoom in and out buttons are now also shown in this mode since https://github.com/navit-gps/navit/pull/306/commits/ba2b45475842d43cf6ac60039a727b4c1c456fa4.

Zoom-to-route:
screenshot_20170731-202828
North-oriented and top view.

### DIFF
--- a/navit/navit.c
+++ b/navit/navit.c
@@ -2596,7 +2596,7 @@ navit_set_attr_do(struct navit *this_, struct attr *attr, int init)
 	case attr_pitch:
 		attr_updated=(this_->pitch != attr->u.num);
 		this_->pitch=attr->u.num;
-		transform_set_pitch(this_->trans, ceil(this_->pitch*sqrt(240*320)/sqrt(this_->w*this_->h))); // Pitch corrected for window resolution
+		transform_set_pitch(this_->trans, round(this_->pitch*sqrt(240*320)/sqrt(this_->w*this_->h))); // Pitch corrected for window resolution
 		if (!init && attr_updated && this_->ready == 3)
 			navit_draw(this_);
 		break;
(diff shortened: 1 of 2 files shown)


jfitie:
Latest commit was tested and works as expected, including the fix for reading the pitch attribute.

pgrandin:
Nice work @jfitie 
These screenshots look good, I'm going to use them for the Play Store :)

PR #308, opened 2017-08-25
Repo cleanup

jkoan:
Cleaned up unused Files and fixed baudrate in example config


pgrandin:
A readme file is a common good practice, I'm not sure that we should get rid of it.
Also, your PR does some cleanup and fixes a documentation issue, which is a slightly different scope.

Removing the files is also what makes the CI fail : 
```
CMake Error at /usr/share/cmake-2.8/Modules/CPack.cmake:395 (message):
  CPack package description file: "/home/ubuntu/navit/README" could not be
  found.
```

jkoan:
Hi Kazer, 
you are right, one Readme file is a must have, but we have the README and the README.MD. Where the md file is as readable as the old one, also the md file has more informations. So my suggestion would be that I fix the build (it's only adding ".md" in one line) and than we can delete the old Readme because everything what is inside the old file is in the new Readme.md file.

PR #309, opened 2017-08-25
WIP: Cmake rework

jkoan:
Another time the Rework of cmake/cpack to be able to build components.

This is usefull to have separate deb/rpm files as example for seperate gui components (if you have qml you dont need to have gtk installed)

But to handle this we need to set the dependencies in the main CMakeList.txt (arround line 820)

- [ ] Finish dependencies for Debian
- [ ] Set dependencies for rpm
- [ ] More work into NSIS (nicer layout for example
- [ ] ...


sleske:
Note: This is a new revision of the work in PR #268.

sleske:
Looks fine to me at first sight. I have not tested it yet - I'll try to do that later to give feedback.
Some cosmetic changes I'd like to see:
* I'd prefer putting the CPack-specific settings into a separate file (probably loading it using `include`).
* This PR contains a lot commits with small fixes. I'd prefer rebasing it before merging (I'll glady do that once the technical details are worked out). Any objections?

pgrandin:
"Also IMHO we generate way too many DEB files - if we just want to make installation easier, what's the point of having one for each plugin? Just install them all..."

I agree that we should keep it simple, but what about dependencies?
If you install the graphics-gtk plugin, it should pull GTK but not QT, and vice versa. I'm not sure that we can provide that without multiple sub-packages.

pgrandin:
@jkoan @aerostitch what are your thoughts?

sleske:
> I agree that we should keep it simple, but what about dependencies?
> If you install the graphics-gtk plugin, it should pull GTK but not QT, and vice versa. I'm not sure that we can provide that without multiple sub-package.

Yes, that makes sense. Debian does something similar. For us, maybe:

* main package
* GTK package  (with gui/gtk and graphics/gtk and gui/internal) 
* Qt package (with graphics/qt5 and the internal and QML GUI) s

jkoan:
Hi @sleske,
@pgrandin and I discussed in the IRC and have the same opinion. More packages are better than fewer packages.
The reasons for that are easy:
- More packages are bringing more flexibility
- More packages are no problem when using package manages (which is planed with #366)
- For debugging you don't need to actually install navit
- If you really want a all in one package you can even build it by setting cpack to all in one (CPACK_MONOLITHIC_INSTALL) 

So my conclusion is that navit is a extremely modular platform which should also be treated modular.

aerostitch:
For a smooth upgrade we need the current main package to still brings what it brings now by making it a virtual package.
As discussed on irc, I'd say we need  navit, navit-minimal, navit-bindings, then merge the gui, graphics, speech and vehicles that are not in code into 1 package for each graphics system. Main map drivers in core and specific ones in a plugin package, maybe also split j580 and pedestrian in their own packages
like navit-plugin-j580

PR #310, opened 2017-09-01
Upgrade to CircleCI v2

pgrandin:
This PR allows up to upgrade to Circle CI v2.
This has a few benefits :
- each build step is now independent and can be run completely separated from the previous one (in v1, we would break the build scripts between different containers, but a single container would execute several build scripts). 
- we can chain jobs for better testing : build a linux test image, and use that image to perform various regression tests. The workflow feature in v2 is quite powerful https://circleci.com/docs/2.0/workflows/
- allows us to restart a single step instead of restarting the full build
- allows us to choose on what image (OS) we will build instead of being limited to Ubuntu 12.04/Ubuntu 14.04

Example view from the workflow result :
ci-workflow

We still need to fix the android_x86 build, but all other builds are working.


pgrandin:
In case someone wants to help fix the android x86 build, here's the current failure:
```
Building C object navit/support/wordexp/CMakeFiles/support_wordexp.dir/wordexp.c.o
In file included from /root/project/navit/navit/support/zlib/crc32.c:29:0:
/root/project/navit/navit/support/zlib/zutil.h:23:22: fatal error: string.h: No such file or directory
 #  include <string.h>
                      ^
compilation terminated.
Scanning dependencies of target support_shapefile
Built target navit_config_xml
/root/project/navit/navit/support/ezxml/ezxml.c:25:20: fatal error: stdlib.h: No such file or directory
 #include <stdlib.h>
                    ^
compilation terminated.
[  1%] [  1%] In file included from /usr/lib/gcc/x86_64-linux-gnu/4.8/include-fixed/syslimits.h:7:0,
                 from /usr/lib/gcc/x86_64-linux-gnu/4.8/include-fixed/limits.h:34,
                 from /root/project/navit/navit/fib-1.1/fib.c:33:
/usr/lib/gcc/x86_64-linux-gnu/4.8/include-fixed/limits.h:168:61: error: no include path in which to search for limits.h
 #include_next <limits.h>  /* recurse down to the real one */
                                                             ^
/root/project/navit/navit/fib-1.1/fib.c:34:20: fatal error: stdlib.h: No such file or directory
 #include <stdlib.h>
                    ^
compilation terminated.
```

So I'm guessing that it's a missing include for the NDK

PR #311, opened 2017-09-05
Revert "Repo cleanup"

pgrandin:
Reverts navit-gps/navit#308

This breaks some external builds


pgrandin:
@jkoan let's talk about a better strategy to address this. @hoehnp found that it was breaking external packages such as Gentoo's ebuilds.

jkoan:
@hoehnp @pgrandin I understand that builds should work, but if the external builds are depending on Files that are waste probably not our project is broken. Instead of this the probability they build navit a wrong way.

Don't get me wrong, I also want that all builds of navit in the world are working, but if they depend on something outdated it's not our fault.
Think of the switch from autotools to cmake. If we say that everything that was working some time should work always in the future we can not change anything.

We must work together with all the people who are building navit or do something with navit but we can't revert every change that's made on navit only because external things are not in sync.

Although the deleted files are no big change, so we can analyze what exactly brakes the external build and revert that. Probably it's that the build depend on the "README" instead of the "README.md" ore something like that.

hoehnp:
@jkoan. This is indeed the case. Both ebuilds reference to a README file and dont find it. For me, it is just unfortunate that I do not hv acces to the two repositories. For now, I could suggest to see, if their repo is on Github or somewhere and I send them a pull-request with that small update. In case you think bringing back the README file is a bad idea.

sleske:
I'm a bit late to the discussion, but just an idea: As far as I understand, GitHub will also use a file called just "README" as the readme file. README.md is preferred if it exists, but is not required.

So why don't we just rename the new README.md to README? Then we get a nice display on GitHub, and a README file for other environments.

We don't _have_ to give the file a .md extension just because it contains Markdown.

jkoan:
I like this idea!

metalstrolch:
I too like the Idea

sleske:
Done, see commit 3299ac4db93.
I'm closing this PR; feel free to reopen if more problems occur.

sleske:
Hmm.... no, that does not work. GitHub also recognizes a file "README", but does not render it as MarkDown, but as plain text :-(. I have reverted my changes. We'll have to find a better solution.

If all else fails, we can include a placeholder README, which points readers to README.md. A bit ugly, but will do until all builds have been changed.

jkoan:
What about a symlink? This should work even with git. I only dont know what windows does with it.

Here is a good explanation of symlinks in git: https://stackoverflow.com/a/954575

sleske:
> What about a symlink? This should work even with git. I only dont know what windows does with it.

It does not work on Windows, at least not very well (Windows does have symlinks, but they work differently, and require admin privs). I don't think that approach is worthwhile.

It's probably best if any build scripts that require a README file are updated to point to README.md (like we did for cpack). Debian for example has many packages that contain a README.md in place of a README, so we're not the only ones doing this.

As to breaking Gentoo's ebuilds: If that refers to the ebuild from [Gentoo bug 176552](https://bugs.gentoo.org/show_bug.cgi?id=176552) - that looks like a single line change. The ebuild explicitly references "README", should be easy to change.

So if there are no objections, I think we should close this PR and let Gentoo update their ebuild. We simply do not have a README file, and playing games with symlinks or copies is not the way to go.

hoehnp:
fine with me. I will try to get both Gentoo ebuilds updated.

sleske:
Thanks! Then I'll close this PR.

PR #314, opened 2017-09-25
Patch:core:More Talker IDs other than GP accepted

naggety:
NMEA messages starts with '$' followed by 2 digits that indicates
the Talker ID. The most common was GP for GPS, but now many modules
can use other satellites systems or even 2 systems at the same time.

Other Talker IDs are:
- GL:russian GLONASS
- GA: european Galileo
- BD: chinesse BeiDu
- GN: Global Navigation: a combination of 2 or more systems.

### DIFF
--- a/navit/vehicle/file/vehicle_file.c
+++ b/navit/vehicle/file/vehicle_file.c
@@ -451,7 +451,7 @@ vehicle_file_parse(struct vehicle_priv *priv, char *buffer)
 		*p++ = '\0';
 	}
 
-	if (!strncmp(buffer, "$GPGGA", 6)) {
+	if (!strncmp(&buffer[3], "GGA", 3)) {
 		/*                                                           1 1111
 		   0      1          2         3 4          5 6 7  8   9     0 1234
 		   $GPGGA,184424.505,4924.2811,N,01107.8846,E,1,05,2.5,408.6,M,,,,0000*0C
@@ -475,14 +475,15 @@ vehicle_file_parse(struct vehicle_priv *priv, char *buffer)
 			if (!g_ascii_strcasecmp(item[5],"W"))
 				priv->geo.lng=-priv->geo.lng;
 			priv->valid=attr_position_valid_valid;
-            dbg(lvl_info, "latitude '%2.4f' longitude %2.4f\n", priv->geo.lat, priv->geo.lng);
+
+			dbg(lvl_info, "latitude '%2.4f' longitude %2.4f\n", priv->geo.lat, priv->geo.lng);
 
 		} else
 			priv->valid=attr_position_valid_invalid;
 		if (*item[6])
 			sscanf(item[6], "%d", &priv->status);
 		if (*item[7])
-		sscanf(item[7], "%d", &priv->sats_used);
+			sscanf(item[7], "%d", &priv->sats_used);
 		if (*item[8])
 			sscanf(item[8], "%lf", &priv->hdop);
 		if (*item[1]) 
@@ -500,7 +501,7 @@ vehicle_file_parse(struct vehicle_priv *priv, char *buffer)
 			}
 		}
 	}
-	if (!strncmp(buffer, "$GPVTG", 6)) {
+	if (!strncmp(&buffer[3], "VTG", 3)) {
 		/* 0      1      2 34 5    6 7   8
 		   $GPVTG,143.58,T,,M,0.26,N,0.5,K*6A
 		   Course Over Ground Degrees True[1],"T"[2],Course Over Ground Degrees Magnetic[3],"M"[4],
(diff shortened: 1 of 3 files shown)


sleske:
Thank you for the PR.

Could you give some background on why this is necessary? Did you find a GPS device that uses a talker ID different from "GP"?

I cannot personally test this, not having a GPS device handy, so some details would be appreciated.

naggety:
Hi!


My GPS module has the capability to read signals from GPS, GLONASS, Galileo and BeiDu positioning systems.


This is the module: http://www.telit.com/gnss/sc872-a/

This is the manual: http://www.telit.com/fileadmin/user_upload/products/Downloads/GNSS/se868aas/1VV0301202_SC872-A_Product_User_Guide_r3.pdf


I can send propietary commands (NMEA compatible) to the module to select which system I want to use. It is also possible to activate GPS and GLONASS at the same time, improving accuracy, satellites coverage... In this case the talker ID sent is GN.


Looking for more info about Talker IDs, I couldn't find a lot of info, maybe this is the best one: http://www.catb.org/gpsd/NMEA.html#_talker_ids


About GN talker ID, it seems that is commonly used by many modules when they use more than one positioning system at the same time to do the calculations (including combinations other than GPS + GLONASS). I haven't been able to confirm this, but it seems to be as I say. You can search for info about GNSS and GN talker ID, maybe you are luckier.


I have also been looking if there are modules that sends data from more than one system but sending a message for each one of them, instead a combined one with GN talker ID. My module doesn't allow to do that, and I think it would be a violation of NMEA specs, since it is defined as a 'single talker, multiple listeners' protocol.


Because of this, I think it is safe to ignore the talker ID, because only one talker ID is going to be received for position and speed messages.



Additional info not very important for this pull request:


The calculation of position, speed, etc can be combined (using GN Talker ID), but the message of satellites view is sent once per system.


For example, if GPS and GLONASS are being used, position message is sent as $GNGGA, but satellites view messages are sent as $GPGSV and $GLGSV (one for GPS satellites and one for GLONASS satellites). At least, my module does it this way, I can't test if other modules work the same way.


Because of this, the info about satellites collected by navit should be modified. Now it only collects info about GPS satellites.


I intended to correct this in a future patch.




Please, tell me if you need more info or clarification.


Regards

Iñigo

________________________________
De: Sebastian Leske <notifications@github.com>
Enviado: lunes, 25 de septiembre de 2017 12:28:17
Para: navit-gps/navit
Cc: naggety; Author
Asunto: Re: [navit-gps/navit] Patch:core:More Talker IDs other than GP accepted (#314)


Thank you for the PR.

Could you give some background on why this is necessary? Did you find a GPS device that uses a talker ID different from "GP"?

I cannot personally test this, not having a GPS device handy, so some details would be appreciated.

navit/vehicle/file/vehicle_file.c
```diff
@@ -475,14 +475,15 @@ vehicle_file_parse(struct vehicle_priv *priv, char *buffer)
...
+
+			dbg(lvl_info, "latitude '%2.4f' longitude %2.4f\n", priv->geo.lat, priv->geo.lng);
 
 		} else
 			priv->valid=attr_position_valid_invalid;
 		if (*item[6])
 			sscanf(item[6], "%d", &priv->status);
 		if (*item[7])
-		sscanf(item[7], "%d", &priv->sats_used);
+			sscanf(item[7], "%d", &priv->sats_used);
```
pgrandin: Indentation issue

navit/vehicle/file/vehicle_file.c
```diff
@@ -475,14 +475,15 @@ vehicle_file_parse(struct vehicle_priv *priv, char *buffer)
 			if (!g_ascii_strcasecmp(item[5],"W"))
 				priv->geo.lng=-priv->geo.lng;
 			priv->valid=attr_position_valid_valid;
-            dbg(lvl_info, "latitude '%2.4f' longitude %2.4f\n", priv->geo.lat, priv->geo.lng);
+
+			dbg(lvl_info, "latitude '%2.4f' longitude %2.4f\n", priv->geo.lat, priv->geo.lng);
```
pgrandin: Indentation issue

navit/vehicle/wince/vehicle_wince.c
```diff
@@ -491,7 +491,8 @@ vehicle_wince_parse(struct vehicle_priv *priv, char *buffer)
 			if (!g_strcasecmp(item[5],"W"))
 				priv->geo.lng=-priv->geo.lng;
 			priv->valid=attr_position_valid_valid;
-		dbg(lvl_info, "latitude '%2.4f' longitude %2.4f\n", priv->geo.lat, priv->geo.lng);
+
+			dbg(lvl_info, "latitude '%2.4f' longitude %2.4f\n", priv->geo.lat, priv->geo.lng);
```
pgrandin: Indentation issue

sleske:
Thank you for the long explanation. I think we can merge the change.

About the indentation problems: I can cherry-pick the change. Then I can fix the indentation in the process, and add some of the excellent information to the commit message. I'll try to do this shortly.

navit/vehicle/wince/vehicle_wince.c
jkoan: no, here was one before

navit/vehicle/file/vehicle_file.c
jkoan: here was one issue before he fixed that

navit/vehicle/file/vehicle_file.c
jkoan: also a fix

PR #315, opened 2017-09-29
Disable some things

jkoan:
- Disable support for things that are not needed inside the android app
- Disable "mg" in general because its unmaintained

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,6 +58,15 @@ jobs:
           command: |
             bash ci/setup_common_requirements.sh
             bash ci/build_win32.sh
+  build_wince:
+    <<: *defaults
+    steps:
+      - checkout
+      - run:
+          name: Build for Windows CE
+          command: |
+            bash ci/setup_common_requirements.sh
+            bash ci/build_wince.sh
   build_tomtom_minimal:
     <<: *defaults
     docker:


CMakeLists.txt
```diff
@@ -709,6 +709,13 @@ if(ANDROID)
    set_with_reason(vehicle/android "Android detected, NDK:${ANDROID_NDK_API_VERSION}, API:${ANDROID_API_VERSION}" TRUE)
    set(SHARED_LIBNAVIT TRUE)
 
+   # We have a mobile Device they dont need null components and most devices have limeted space, also download size is important 
+   set_with_reason(vehicle/null "Android detected, saveing space" FALSE)
```
pgrandin: `saveing` -> `saving`

CMakeLists.txt
```diff
@@ -489,7 +489,7 @@ add_module(gui/internal "Default" TRUE)
 add_module(map/binfile "Default" TRUE)
 add_module(map/filter "Default" TRUE)
 if(NOT MSVC)
```
pgrandin: Why do we disable it also for Windows?

CMakeLists.txt
```diff
@@ -709,6 +709,13 @@ if(ANDROID)
    set_with_reason(vehicle/android "Android detected, NDK:${ANDROID_NDK_API_VERSION}, API:${ANDROID_API_VERSION}" TRUE)
    set(SHARED_LIBNAVIT TRUE)
 
+   # We have a mobile Device they dont need null components and most devices have limeted space, also download size is important
```
pgrandin: `limeted` -> `limited`

sleske:
Disabling map/mg and the "null" drivers for Android makes sense - I don't think anyone uses them. I would keep shapefile for now - it's only 164 kiB (or about 1% or the Navit APK). 

At any rate, I strongly object to globally disabling map/mg - either we keep building it, or we delete it. Keeping it in the source tree but not building it will probably cause it to break, so we might as well delete it.
Actually, deleting it might be the best option - while it is historically important (it was Navit's first map driver), the maps it can read are proprietary and no longer available (not to mention over ten years old). Or is there a reason to keep them?

jkoan:
Hi @sleske thanks for your opinion. I will delete map/mg after this PR has been merged. Therefor i will open another PR. About Android im Okay. Update will come in a short time.

CMakeLists.txt
jkoan: See the main discussion. All others are fixed

PR #324, opened 2017-10-14
add:icon:dam

youte62:
add dam.svg designed by me - free to use
dam


pgrandin:
Looking good. Thanks!

sleske:
Hi, thanks for your icon!
I noticed you submitted numerous icons in the same style, with a big blue bubble around them. I don't think this style will work well for Navit. The big bubble does not convey any information, and causes all icons to look similar. At the moment, icons are resized to 22x22px for the Navit GUI. Your icons become very hard to distinguish at that size. 

I resized two of your icons (dam.svg and diving.svg) to 22x22 px, and included one existing icon (viewpoint.svg) for comparison. Note how the viewpoint icon is clearly distinguishable, while the others are not:
dam-diving-viewpoint

Could you consider this in your designs?

youte62:
Hello. The icon on 22px is an error. The other size is 200px

Le 18 oct. 2017 08:42, "Sebastian Leske" <notifications@github.com> a
écrit :

metalstrolch:
I can clearly second the concerns of @sleske . See the attached screenshot of navit on Sailfish jolla, where you cannot notice the blue bubble beeing the "drinking water" icon at 15x15, while the "old" icons are clearly distinguishable. We could opt at increasing the icons to 20x20 again as was the default previously, but icons as big as 200x200 would clearly clobber the small screen.

metalstrolch:
Just for fun: the same icon at 200x200

metalstrolch:
The style does not work well in POI lists as well. There the pointing bubble makes no sense at all. I would just get rid of the bubble and keep the symbols inside them. If they get a contrast (white border around dark symbol) they could work quite well.

sleske:
@metalstrolch : 
> The style does not work well in POI lists as well. There the pointing bubble makes no sense at all. I would just get rid of the bubble and keep the symbols inside them.

Yes, that's what I wanted to say. The bubble is a) problematic due to the small icon size, and b) does not make sense in the context we use the icons.

sleske:
@youte62: 

> Hello. The icon on 22px is an error. The other size is 200px

Sorry, I think there's a misunderstanding. 22px is the size that Navit uses on the map (and it would be tricky to change that). All icons are automatically resized to that size. Or what do you mean by "the other size"?

PR #328, opened 2017-10-14
add:icon:dangerous

youte62:
add dangerous.svg designed by me - free to use

dangerous


pgrandin:
As discussed in #334 we should avoid text in the icons as we can't translate it.
Also can you please update this one so that it doesn't have the circle around? 

Thanks!

metalstrolch:
I would have a look on european pictograms for danger areas like this one: https://de.wikipedia.org/wiki/Datei:DIN_4844-2_D-P006.svg
Think that could work quite well.

pgrandin:
Too long without updates, closing the PR.

PR #333, opened 2017-10-14
add:icon:governement_building

youte62:
goverment_building-montage

add governement_building.svg designed by me - free to use


pgrandin:
I wonder if we wouldn't need a white background between the columns for this one.

PR #334, opened 2017-10-15
Add:icon:height

youte62:
height-montage

### DIFF
--- a/navit/icons/height.svg
+++ b/navit/icons/height.svg
@@ -1,0 +1,130 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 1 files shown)


pgrandin:
Nice icon, but it looks more like a weight, not 'height' :)

metalstrolch:
Technically OK for me, but this kind of "road sign" style is unusual outside of the US. Beside that, we have no framework to translate text on icons, have we?

pgrandin:
You're right @metalstrolch icon should probably avoid using text as it won't be translated.
Also, I'm actually not sure if we really need this one.
OSM uses `maxheight=*` for bridge clearance. `height` is used for building heights, which we currently don't use in Navit : http://wiki.openstreetmap.org/wiki/Key:height

The only reference to the height icon I could find in Navit was in the Garmin types : https://github.com/navit-gps/navit/blob/trunk/navit/map/garmin/garmintypes.txt#L205-L206

Reading that file I am unsure of its usage.

pgrandin:
Too long without updates, closing the PR.

PR #336, opened 2017-10-18
Sailfish icon size fix

metalstrolch:
Navit medium GUI icon size fix

This changes the medium size icon for HMI to 64x64px. This is necessary
because of changed icon set and therefore different icon default sizes.
Otherwise the icons on POI screen were way too big.


metalstrolch:
Merging this, as this does not affect anything else than SailfishOS build.

PR #343, opened 2017-10-22
Icon communication

youte62:
communication-montage


metalstrolch:

sleske:
@youte62 : I really appreciate all these icons you are creating, however some confuse me a bit. What is the purpose of this icon in Navit? Is it supposed to be shown on the map? In a menu? While it's good to have nice icons, there's little point in including them in Navit without knowing what to use them for. So what is your goal?

youte62:
I was asked to create icons missing. What I'm doing, later I do not know where it will be used.

sleske:
@youte62 : Thanks for the explanation. I just saw there is actually a POI type "communication" in Navit (though it's only used by the Garmin map driver), so we may be able to use the icon.

PR #344, opened 2017-10-22
Icon dam

youte62:
dam-montage 1


metalstrolch:

metalstrolch:
Looks way better than than previous version in #324. But as mentioned there, think of adding a contrasting colour (either black or white typically, depending on how dark the icon is) outline to the symbols.
Have a look to the screenshots posted by me on #324. On the POI screen, you clearly see the white outline making the mostly black old icons work on the black background.
This is not "as much" a problem with this icon, as there is seldom any "all blue" background, but it would give more contrast having a, white in this case, outline on a black backgrounded screen.

pgrandin:
I rebased @youte62 's branch, and will work with him to fix the other PRs.
Green light for this one?

metalstrolch:
OK for me. Except i did not mean putting a solid background behind the symbol for outline. But that's OK for now.

pgrandin:
I actually like your idea of using an outline instead of a background. 
@youte62 what do you think?

PR #346, opened 2017-10-22
Icon drinking water

youte62:
drinking_water-montage


metalstrolch:

metalstrolch:
Getting these changes into trunk, as the "intermediate" state in trunk is not good.

Issue #347, opened 2017-10-23
Offer Taiwan as a separate download in the Android download dialog


jidanni:
Why do I have to download 981 megabytes of three countries when all I want is Taiwan. I have not left Taiwan in a quarter century. Maps. Me allows one to even download just half of Taiwan. And other applications allow one to have a rectangle so they can just download a few square kilometers. I can't even get started using this application till I download 981 megabytes!

metalstrolch:
You can use the Navit planet extractor to download any custom region you want. See http://maps.navit-project.org

pgrandin:
@jidanni what platform are you using?

jidanni:
Android.

sleske:
I think what @jidanni wants so say is: "Why does the Android download helper only let me download "Japan+Korea+Taiwan" (about 800 MiB), when I only want Taiwan.
That is indeed a valid point. I edited the title.

We should consider offering Taiwan as a separate download. According to http://maps.navit-project.org/ , just Taiwan would only be about 150 MiB.

sleske:
Incidentally, it would also make sense to offer Korea as a separate download. Japan + Korea makes sense, because we can only download rectangles, and a rectangle containing Japan also contains Korea, because of Japan's banana shape (no offense :-) ).

pgrandin:
@sleske indeed!

@jidanni the technical reason is that currently the list of countries is hardcoded into Navit for the android map download, and that list could be more granular.

This is where Taiwan is defined : https://github.com/navit-gps/navit/blob/trunk/navit/android/src/org/navitproject/navit/NavitMapDownloader.java#L127

As a short term fix, I moved Taiwan to its own menu entry in https://github.com/navit-gps/navit/commit/bcd60ee37e970973c17c2e71d8953c800dbe7ebf we will still need to merge that in trunk.

We can push that to the Play Store today or tomorrow. In the mean time, if you want to test it you can download an apk with this change here : https://3271-30791823-gh.circle-artifacts.com/0/root/project/android-arm/navit/android/bin/navit-bcd60ee37e970973c17c2e71d8953c800dbe7ebf-arm-debug.apk

The long term fix will be to have a more detailed list of countries (and regions, for some countries) and keep that list outside of Navit's code. Our member @jkoan started working on this but this is not completely ready yet.

pgrandin:
It works :)

[...]

PR #348, opened 2017-10-24
fix:android:moved Taiwan into its own map download entry

pgrandin:
Split the download block from Japan + Korea + Taiwan into Japan + Korea, and Taiwan in a different block.

Addresses #347


jkoan:
I think this is good. :) Mergeing

PR #354, opened 2017-10-29
Build doxygen doc and push the result to a gh-pages branch

aerostitch:
This way it's available on http://navit-gps.github.io/navit/ and the doxygen.navit-project.org can be used there.


pgrandin:
Nice work @aerostitch !

PR #355, opened 2017-10-30
Proposal: Follow semver rules and update the changelog to make it usable

aerostitch:
Proposal: use semver and base the changelog on the http://keepachangelog.com/ template.
There's been no release in a long time and I'm not 100% sure if all the changes I put in the current version were totally significant for the current release, but I wanted to have the PR out  to see what you guys think about it.
If you're ok with the principle, I'll continue updating the changelog with the other changes I see.

Maybe we just put the most important ones for this release (currently 617 commits since v0.5.0...) and do more frequent releases in the future which would make the changelog more manageable.


pgrandin:
Nice work @aerostitch ! Thanks for taking care of that. 
I think that this is a good idea indeed. Let's hear other's opinion.

sleske:
The template from http://keepachangelog.com/ looks good. Originally Navit used to enforce a certain format for commit message, with the idea that the commit log would become the changelog - see https://wiki.navit-project.org/index.php/SVN_guidelines#Format_of_the_commit_log . However, we have long stopped enforcing that, and we have loads of commits which don't follow the rule - and arguably it was always a problematic idea to use the commit log that way, so a better way probably makes sense.

However, I don't quite understand the idea to use Semantic Versioning. semver seems to be for versioning APIs, not end user software - it continually talks about "API changes". I don't see how this relates to Navit. At any rate, the more important thing is to decide when and if we want to have releases...

mvglasow:
Re @sleske’s comments: I guess the commit message format was abandoned (de facto) pretty much around the time we switched to git. That is in part because with SVN only a few chosen people could create commits (and were given commit access only after being informed of that convention) while others would send patches and have someone with commit permission create the commit, message included. Now people file merge requests, and those of us with commit access merge them, including the author’s comments. Git has made this format harder to enforce—unless we start rejecting all merge requests with commit messages that do not follow our convention.

Semver seems to be all about mitigating dependency hell. Makes perfect sense for a library or anything with a public API. Navit, however, is pretty much at the top of the “food chain”—few packages depend on it, and while it has some things that can be considered an API (dbus comes to mind), its primary use is as an end-user app. One might argue whether the UI itself is a kind of API as well.

For end-user apps, it is quite common to increase the major version number when major new functionalities are added, which for semver would only warrant a minor version number increase. Conversely, if we consider the UI to be part of the API, we might need a new major version whenever we move a menu item (e.g. Menu > Quit instead of Menu > Actions > Quit).

I’d therefore propose a variation of semver, adapted to an end-user app:

* x.y.Z (patch): only bug fixes or refactoring, no changes in functionality
* x.Y.z (minor): added or changed functionality but can be used as a drop-in replacement for the previous version (all data formats and interfaces are still supported); minor UI changes (such as moving individual menu items) are also allowed
* X.y.z (major): at least one of the following:
  * Major new functionality (such as Augmented Reality, inertial navigation or support for live traffic services): de-facto standard for end-user apps
  * New user interface (such as moving from the old pulldown menu UI to the Internal GUI): this is definitely the UI equivalent of a breaking API change
  * Dropped support for a data format or interface: also a breaking change and usually tends to occur along with larger changes which would warrant a new major version anyway

aerostitch:
Yes, @mvglasow what you are proposing is exactly what I have in mind on how to apply semver to end-user apps.
One of the idea would also be to encourage more frequent releases (from a packaging prospective, when you are packaging for let's say debian, no release since 2015 make the project look dead even if it's not at all)

jkoan:
I completely agree the idea, explicitly with the last two comments from @aerostitch and @hoehnp.
I also want to add what we should do for the nightly channel:
- own fdroid repo with beta builds from circleci
- own repo with builds for deb, rpm and tar.gz which can be used by adding to the system and also probably a rss feed 
- dev blog on the new and upcoming website (build by github pages (jekyll) --> add post with git and markdown in the future)

aerostitch:
I updated the changelog but I can't fit everything in there, it's way too long and I don't know the tool enough.
If you're ok with the current version, we can merge and start from now on to add the new changes to it.
Thx

jkoan:
I Would like to merge this PR in NOW even if the Changelog is probably not perfect. It will never be perfect, but it is really really good and it will be very useful for the whole project. 
If this would be now merged in we could track new changes directly in the file ;)
So: :shipit:

PR #356, opened 2017-10-31
Android platform tools workaround

jkoan:
This pull request adds a workaround to the CI because current platform tools from google are broken.
This fixes Ticket #1354


aerostitch:
We were woeking with @pgrandin on the branch google_play to upgrade to the new sdk commands as the current ones are deprecated. Might be a longer term approach. Currently the only problem left lies in the order of the include directives.

aerostitch:
Note that you're more than welcome to give a new eye on this branch @jkoan ;)
Sorry I was supposed to update the ticket yesterday but got bumped with work.

jkoan:
I will defensively look into it, but the fix (see #357 ) only for the time until the new command is ready.
I have spoken to @pgrandin about that. He pushed me to do this PR ;)

pgrandin:
Yeah we definitely need to finish the work in the google_play branch as the `android` command is deprecated, but at least we can still build and release until we figure out the issue @aerostitch just mentioned.

PR #357, opened 2017-10-31
Add self checking workaround for android platform tools

jkoan:
This pull request adds a workaround to the CI because current platform tools from google are broken.
This fixes Ticket #1354

(This is the replacement for PR #356 because I had a problem with my repo)


pgrandin:
Thanks @jkoan ! We still need to work on the toolchain update as the `android` command is now deprecated but at least we can build again.

PR #359, opened 2017-10-31
fix:gtk:fixed missing imperial units in the GTK ui.

charlescurley:
modified:   navit/gui/gtk/gui_gtk_statusbar.c

The status bar under GTK was only showing metric units. So I coded up a fix.

I tested as follows:

* Initialized the local variable imperial to true. Had imperial set to one in my navit.xml. Got imperial measures, as expected.

* Changed the local variable initialization to false. Recompiled and ran. Got imperial, as expected. This indicates that the function collected it as expected.

* Changed the attribute in navit.xml to 0. Ran again. Got metric units as expected.

### DIFF
--- a/navit/gui/gtk/gui_gtk_statusbar.c
+++ b/navit/gui/gtk/gui_gtk_statusbar.c
@@ -22,6 +22,7 @@
 #include <time.h>
 #include <math.h>
 #include <gtk/gtk.h>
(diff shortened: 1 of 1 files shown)


charlescurley:
The fix is on branch fix_gtk_imperial_use

pgrandin:
Thanks @charlescurley for fixing that!

PR #360, opened 2017-11-01
Roadbook imperial measures fix

charlescurley:
The road book now respects imperial for two more columns. The first column (Command) was good for both feet and miles. However, column 2 (Length), and the 4th column (Destination Length), were showing up only in metric. This commit fixes columns 2 and 4. This should fix bug 780.


pgrandin:
This PR is "reversed" : into roadbook.imperial.fix from trunk 
Should have been the opposite, but commits have been pushed to trunk.

PR #362, opened 2017-11-03
Gtk add follow vehicle toggle

charlescurley:
Added the ability to toggle whether navit follows the active vehicle on its display to the Display menu.


pgrandin:
👍

PR #365, opened 2017-11-10
 Fix:Core:only GGA or RMC message required, ticket #913

naggety:
Fixed bug from ticket #913


navit/vehicle/file/vehicle_file.c
```diff
@@ -377,6 +378,42 @@ vehicle_file_enable_watch_timer(struct vehicle_priv *priv)
 
 
 //***************************************************************************
+/** @fn static void vehicle_file_fix_timeout_cb( struct vehicle_priv *priv )
+*****************************************************************************
```
pgrandin: Kudos for the documentation, but your documentation style is a bit different from what we do usually.
Would you mind adapting that so that it's more consistent?
Thanks!

naggety: Done. I had used the same style than in the rest of the file. Now I've changed the style of all the comments of the 2 files to be the same than you usually use.

PR #367, opened 2017-11-11
fix:gtk:Fix imperial factors

charlescurley:
* Some new conversion factors in navigation.h.

* Corrected conversions to to use the correct factors.

* Got rid of a hard coded value, replaced it with a #define.


charlescurley:
On Sat, 11 Nov 2017 20:10:53 +0000 (UTC)
Pierre GRANDIN <notifications@github.com> wrote:

> Merged #367.

AAARRRGGHH! I need another pair or two of eyeballs. Now the roadbook is
reporting miles incorrectly. I think I am missing something here.

-- 

The right of the people to be secure in their persons, houses, papers,
and effects, against unreasonable searches and seizures, shall not be
violated, and no Warrants shall issue, but upon probable cause,
supported by Oath or affirmation, and particularly describing the
place to be searched, and the persons or things to be seized.
-- U.S. Const. Amendment IV

PR #368, opened 2017-11-11
add:audio:add audio framework

pgrandin:
This has been discussed in the past, but I'd like to bring back the topic.

This PR adds audio playback capabilities to Navit.
Is is fully modular and pluggable, as any other components within Navit.

Having audio playback could be considered somewhat of a scope creep, but there is actually a really important use case in my opinion : interface feedback. When the input device is a piece of glass, having audio feedback is actually useful to inform the user that the touch event was recorded. I do believe that this is part of what we could/should improve with the new QT UI.

I've been maintaining the media playback feature in a fork for several years, but a few things are driving me to try to get it integrated in trunk:
- this is a must have for carputer users. Other solutions exist, usually trying to embed navit within a media player, or worse switching between apps. In terms of user experience, this is quite poor.
- maintaining this in a fork is painful and I'd really like to stop doing that
- this code is modular, documented, and tested in real life

While carputer users are not our major userbase, this PR makes Navit a first class citizen for this use case and could help us expand our community.


hoehnp:
Would be audio based on qt supported on all of our platforms?

pgrandin:
@hoehnp yes, if we implement an output-qt plugin, this should support any and all platform that is supported by QT.

metalstrolch:
Persobally I don't have a good feeling of this. We should concentrate on making Navit a good routing application, and not a egg-laying-wool-milk-pig.
Do we want to link Navit to the <insert favourite streaming provider>  next to spotify too? And what about video playback? 

I think we should cope for input feedback in the GUI component if not supported by the platform anyway if that is really required. I'd prefer to see a feedback framework for internal gui instead. This would help augmenting the internal GUI with feedback. But mixing this with a spitify client seens to be Nuts.
Btw. this would interfere with voice prompting typically.

sleske:
> When the input device is a piece of glass, having audio feedback is actually useful to inform the user that the touch event was recorded.

Sorry, that does not make sense. How can you use a "piece of glass" as the input device?

And while audio feedback can be useful, I don't see the relation to audio playback - you'd typically use voice output for that, which we already have. You don't usually play a song to acknowlegde a button press...

mvglasow:
I see the point for audio feedback (which I take to mean ”make a sound when a UI element is pressed”)—so far, so good. But the PR is largely about adding full media player capabilities to Navit, effectively taking it from a navigation system towards a full car infotainment system.

**TL;DR: We can think about having Navit coexist and interact with a media player, but I consider media playback out of scope for Navit.**

Navit covers a wide range of devices, including car PCs as well as smartphones. The latter category of devices will usually have all the media player features in some other app, and smartphone users who get Navit likely expect a good navigation app, and will not care much about media playback in Navit.

I understand that for a car PC the use case may be different—but I still don’t think this is something which should go into Navit core. Different suggestions:

* Extend the plugin interface so that Navit can control media plugins—but keep it a generic interface, not a purpose-built one. Then offer the media plugin as a separate add-on—residing in its own source tree and installable as an add-on package (for Linux and Windows at a minimum, this should not be too difficult). That way, users can decide if they want that functionality or not.
* Provide an interface for Navit to control external media players. Again, keep that generic, e.g. by adding everything to send DBus commands—then media player control would just be a configuration thing.

Generally, for a car infotainment scenario, I would aim for a generic framework around both Navit and a media player app (and whatever else should be part of that system). IIRC the police in the German state of Brandenburg uses Navit in that kind of configuration.

Another example, somewhat similar in spirit, is the UX of the Audi RNS-E:
* There’s a row of buttons at the bottom of the screen through which I can select the feature I would like to use: e.g. radio, CD/MP3, navigation, traffic information. When I choose one, that UI takes up the entire screen.
* There are hardware controls for volume and media back/forward, which always apply to whatever media is playing (radio, CD or MP3), even when the media UI is not visible (e.g. because the screen is in navigation mode).
* Information about currently playing media is shown on the external display, even when the head unit display is showing something else (e.g. navigation).

As for the framework, I find that a good solution (it’s just the sat-nav itself that is poor, but that’s why I’m here to help build a better one :-)

Without the external display, one could instead have a top bar on the car PC screen which shows media information (and possibly media control buttons if no hardware buttons are available). Maybe one could even adapt the solution I coded for Android a while back (late 2015–early 2016), which makes the system bars a semitransparent overlay over the map, and allow the framework to display the media control bar in the same way.

sleske:
@mvglasow : You've taken my words out of my mouth :-).
I appreciate the work which pgrandin did - it's just not a good fit for the main Navit project and repository.

mvglasow:
@sleske Absolutely—it’s great work (besides illustrating nicely how to add a new plugin type to Navit), maybe it can be refactored in some way to be independent of Navit

jkoan:
Hi guys,
my suggestion would be to separate the audio interface(like @sleske suggested) from the music player itself.

After splitting both my suggestion would be a plugin dev header file. Like other projects we should bring out devel packages. This would bring the benefit that plugins can totally build separate from navit itself. You would only need the headers which are in the devel package. Then you can build whatever you want with the Navit api. 

Of course we need to make clear how to setup and use such a plugin. I would suggest a own repo with examples how to use specific components of the api (for example vehicle, gui, graphics and so on)
After that we have more flexibility and if should be easier for non team members to build own plugins for navit.

Then the media player plugin would go into its own repo because it can be built without building it within navit.

@pgrandin what do you think? The work is definitely really nice so we should provide a easy way to the end user to use it if they want.

pgrandin:
Thanks for the feedback everyone.
I'll move that code to a different repository. Please do not delete the branch until it's done.

PR #374, opened 2017-11-23
Android notification fix

jkoan:
It now uses NotificationCompat from android support to support android from api 4 and up.


pgrandin:
Thanks for fixing this @jkoan !

PR #375, opened 2017-11-23
Add some Documentation to Plugins

jkoan:
This PR add Documentation to the Plugins and the Plugin Sytem.
Also all support libraries are excluded to keep focus on navit and because they are not navit related.

### DIFF
--- a/navit/Doxyfile
+++ b/navit/Doxyfile
@@ -809,7 +809,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
-EXCLUDE_PATTERNS       =
+EXCLUDE_PATTERNS       = */support/*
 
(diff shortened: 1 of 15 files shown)


navit/Doxyfile
```diff
@@ -1504,7 +1504,7 @@ MATHJAX_CODEFILE       =
 # The default value is: YES.
 # This tag requires that the tag GENERATE_HTML is set to YES.
 
-SEARCHENGINE           = NO
+SEARCHENGINE           = YES
```
pgrandin: Of course!

navit/plugin.c
```diff
@@ -1,4 +1,4 @@
-/**
+/*
  * Navit, a modular navigation system.
  * Copyright (C) 2005-2008 Navit Team
```
pgrandin: No big deal for this one, but we could probably update the (c) statement to a current year when we edit a file.

PR #376, opened 2017-11-25
fix:android:Use date of commit as version code instead of system time

jkoan:
Android version code now uses again the timestamp of the last commit instead of current system time and in the about menu the commit hash of the commit is shown again


jkoan:
In the current state, the git client is not able to print out the time of the last commit in a given format like the ubuntu version of git does, i will ether try to update our infrastructure to build on newer images or parse the datecode from a available format. But we need to respect timezones because navit devs are located all over the world

cmake/version.cmake
```diff
@@ -2,28 +2,65 @@ FIND_PROGRAM(GIT_EXECUTABLE NAMES git git.exe DOC "git command line client")
...
+      COMMAND ${GIT_EXECUTABLE} log "--format=%ct" # output as unix timestamp
+      WORKING_DIRECTORY "${SOURCE_DIR}"
+      OUTPUT_VARIABLE GIT_OUTPUT_DATE
+      ERROR_VARIABLE GIT_ERROR
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+      ERROR_STRIP_TRAILING_WHITESPACE
+   )
+   if(NOT GIT_OUTPUT_DATE)
+      message(STATUS "Cannot determine current git commit - git error: '${GIT_ERROR}'")
+      set(GIT_OUTPUT_DATE "0000")
```
pgrandin: This doesn't seem to be consistent with the code on line 33, is that on purpose?

jkoan: oh yes, it seems that i missed that one. Even through it would not brake anything because the version code would be zero not relevant how much zeros you have

PR #377, opened 2017-11-25
fix:cmake:Make copy of files more reliable

jkoan:
This PR should help with failing builds. It uses the cmake copy function instead of an "externel" copy. The function will check if the action was successful.


sleske:
Could you explain how this will help with failing builds (and with which ones)? As far as I see, the old version also uses commands internal to CMake, it just re-executes CMake, which is something we do in many places.

jkoan:
As far as i know: If you execute cmake as an external task you must check for the error message explicitly, but if you include it as a function if has all checks included.
But i will make more changes into this PR at other places, so copy errors should be away.

sleske:
> As far as i know: If you execute cmake as an external task you must check for the error message explicitly, but if you include it as a function if has all checks included.

Yes, good point. WIth that in mind, your changes make sense.

PR #384, opened 2017-12-04
fix:ci:Install imagemagick for android build to convert images 

jkoan:


pgrandin:
Let's move it to setup_common_requirements instead.

PR #386, opened 2017-12-23
fix:wince:fixed the wince build

pgrandin:
Since our previous efforts to get the wince toolchain to work on a recent setup failed, I took the problem the other way : building wince on a ubuntu-8.04 container.

The build generates valid .exe files, but currently has one drawback : something is broken or outdated in the certificate chain and ssl requests won't work. This isn't a big problem for the build itself, but this is currently preventing us from uploading the artifacts in circleci ( this will fail with `x509: failed to load system roots and no roots provided` ).

The same setup can be used to build locally :
- `docker run -ti -v /home/pierre/dev/wince/:/root/ pgrandin/wince:8.04 /bin/bash`
- `mkdir /root/navit/wince/ &&  cd /root/navit/wince/`
- `cmake -DCMAKE_TOOLCHAIN_FILE=$NAVIT_PATH/Toolchain/arm-mingw32ce.cmake -Dsvg2png_scaling:STRING=0,16 -Dsvg2png_scaling_nav:STRING=32 -Dsvg2png_scaling_country:STRING=16 $NAVIT_PATH`
- `make`


metalstrolch:
Maybe CircleCI is using a custom root cert?
Compare /etc/ssl/certs/ca-certificates.crt of your new docker container with some of the working ones.
Or just try replacing it by the one from a working container.

pgrandin:
> Or just try replacing it by the one from a working container.

I tried that quickly indeed, by copying the /etc/ssl-certificates folder from a 16.04 to this 8.04, it didn't work unfortunately.

pgrandin:
I managed to fix the certificate issue, using the following Dockerfile : https://github.com/navit-gps/Dockerfiles/blob/master/wince/Dockerfile

 🍾 
 
A resulting .cab is available here : https://4514-30791823-gh.circle-artifacts.com/0/root/project/wince/output/navit.cab (all the wince artifacts are available here : https://circleci.com/gh/navit-gps/navit/4514#artifacts/containers/0 )

sleske:
@pgrandin : It's cool you got the build to work. Just out of curiosity: Which devices does this build support? The [Wiki page on WinCE](https://wiki.navit-project.org/index.php/WinCE) just says Windows CE 5.0 is supported (and does not even say on which devices).
Maybe you could update that page? There's not much point in having a build for Windows CE if noone knows what hardware it runs on...

pgrandin:
@sleske good point. The toolchain should build content compatible with "Windows Mobile 6.1 and higher". That's what they, without more details. We have a user on IRC that has been around for a few weeks now, trying to get Navit to work on a AVIC-MRZ02 head unit. It took some trial but we are now able to start Navit on that platform (we still need to figure out an issue with image loading though). 

Hopefully we can get a couple of users who still own wince devices to run some test and provide feedback. I'll update the wince page with what I have so far though.

sleske:
> Hopefully we can get a couple of users who still own wince devices to run some test and provide feedback.

Yes, that is the crucial point, I think. Windows Mobile is quite old (Win Mobile 6.1 is from 2008) - not to speak of WIndows CE. The AVIC-MRZ02 is described on [Pioneer's page](http://www.pioneerelectronics.com/PUSA/Car/GPS-Navigation/AVIC-Z2) as being from 2007. Frankly, I'm afraid we are maintaining code for a platform nobody uses - but maybe I'm wrong. We'll see.

PR #387, opened 2017-12-26
add:graphics-qt5:add missing watchEvent slot

pgrandin:
This PR adds a missing watchEvent slot, which is required to get things like gpsd vehicle update to work.


pgrandin:
> These are the mistakes that could be avoided by using the new function pointer based connect syntax.
I'm not sure that I'm aware of that. Do you have an example?

> Seems the qt5 based vehicle I use on sailfish doesn't need this. So I never encountered the err.
Yep, makes sense.

PR #388, opened 2017-12-26
Sailfish build on newer SDKs

metalstrolch:
This pull request enables Navit to be built on newer Sailfish SDKs which uses different tool paths than before. Beside that is adds now mandatory build requirement "git".


jkoan:
I would like to have a check for the variable (only a small message if the variable VERSION_ID is not set)

jkoan:
Its only for sailfish so it has no big impact --> merge

Issue #383, opened 2017-12-04
Merging rules


metalstrolch:
While having strong requirements for something being able to be merged into navit's trunk is generally a good idea, I encountered the question if enforcing a branch to be up to date with trunk before merging plus the requirement to have all checks passed in combination is not too much.

This makes working with multiple pull requests in parallel a really time consuming job. Since the merge of one pull request forces the updating of the others which forces rerunning all ci tests which seem to be somehow unreliable at the moment too. Think about coping for all those new icon things.

So I would prefer returning to old behaviour where it was not explicitly required for a branch to be up to date when merging.

What do you think?

jkoan:
I would suggest that we should keep the behavior as it is now, except for images/icons and documentation. So probably because github cant handle this we need to disable the rule on github but need to respect it in our mind.

sleske:
I'm also for "disable on github, respect in our mind". If you are certain the problems shown in the Github interface are unjustified, don't let them stop you.
Of course, I usually commit directly, not via a PR, so who I am to argue...

metalstrolch:
My main reason to bring this on topic was mainly the reason that at the moment github refuses to merge even if there are no conflicts, just because the latest trunk was not merged to the branch. Which is quite annoying, since after every push to trunk you have to redo this, including otherwise unnecessary ci run.

So I vote for disabling that rule as well.

sleske:
As I wrote in #389 , I think the present rules are  too restrictive.
Feature branches (which PRs are) should be restricted to risky, disruptive changes, or if a discussion is desired. Having a PR for every trivial change just feels wrong.

@pgrandin explained that one goal of the setting was to avoid breaking the main branch. That is a good point - however, that is why we have both "trunk" and "master". trunk is where development happens, and may break occasionally. trunk can be merged to master (or rather, master fast-forwarded to trunk) periodically once trunk has passed all CI tests.

If there are no objections, I'll disable the branch protection for trunk, so pushing and merging is unrestricted. We have worked like this for a long time, and I don't recall any drastic problems.

Comments? Especially from @pgrandin ?

jkoan:
@sleske what about disableing the branch protection for trunk? Any progress? :D

sleske:
@jkoan : Sorry this took so long, was busy for a while

Branch protection for trunk has been reduced - merging & pushing are now unrestricted.

The branch is still marked as "protected", which means that force-pushing and deletion are forbidden. That seems reasonable to me, so I left it as-is.

[...]

PR #389, opened 2018-01-04
Refactor:route:Improve documentation

mvglasow:

### DIFF
--- a/navit/route.c
+++ b/navit/route.c
@@ -1823,12 +1823,10 @@ route_path_add_line(struct route_path *this, struct coord *start, struct coord *
 }
 
 /**
- * @brief Inserts a new item into the path
+ * @brief Inserts a new segment into the path
  * 
- * This function does almost the same as "route_path_add_item()", but identifies
- * the item to add by a segment from the route graph. Another difference is that it "copies" the
- * segment from the route graph, i.e. if the item is segmented, only the segment passed in rgs will
- * be added to the route path, not all segments of the item. 
+ * This function adds a new segment to the route path. The segment is copied from the route graph. If
+ * `rgs` is part of a segmented item, only `rgs` will be added to the route path, not the other segments.
  *
  * The function can be sped up by passing an old path already containing this segment in oldpath - 
  * the segment will then be extracted from this old path. Please note that in this case the direction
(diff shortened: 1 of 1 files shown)


sleske:
Thanks for improving the docs. Often overlooked, yet imporant 👍 .
And BTW: I don't think simple changes like this really need a PR. Just commit them to trunk - we can always fix any mistakes later.

mvglasow:
@sleske afaik direct commits to trunk have been disabled, which leaves pull requests as the only option.

sleske:
@mvglasow : Ah, I didn't know. Seems it does not apply to me, as I'm a team admin.
Personally, I think that is too restrictive. Feature branches (which PRs are) should be restricted to risky, disruptive changes, or if a discussion is desired. Having a PR for every trivial change just feels wrong.
I'm not sure where to bring this up, however - do you know how this was decided?

mvglasow:
@sleske: @pgrandin added that a couple of weeks ago (in fact it seems I was the first to stumble across this, trying to push some documentation changes). We talked about this in IRC (https://irclogs.navit.ie/%23navit-2017-10-27.log) and I made some suggestions, but I didn’t follow up any more after that.

Which again raises the question: where would we discuss such things? There’s ticket #383, somewhat related; I would in fact favor trac for such discussions (feel free to CC me if you open a ticket).

metalstrolch:
I would prefer Github, as I dont have ( and dont want to) have a Track account. Beside that you can easily refer things in Github. So why a double infrastructure!

mvglasow:
@metalstrolch Trac was there before we switched to Github, and as a result, all tickets affecting Navit live there—so technically, Github issues are the redundant part in the infrastructure ;-) Trac supports OpenID with a number of authentication providers (Navit wiki as well as Google, among others, not sure about Github), thus you don’t need a separate account. I do see your point about Github references—maybe that can be added somehow. Though all of these are even more OT than the discussion about changes to trunk…

sleske:
Well, it's really unfortunate that we have tickets both in Trac and in Github (issues). I initally opposed enabling issues in Github because of this, but I was overruled.
I think the best option at this point is to migrate the Trac issues to Github and shut down Trac (or make it readonly). But that's a different discussion...

sleske:
As to the commit / merge rules:
As explained above, I strongly object to requiring PRs for everything.
I think we should continue the discussion in #383 . I'll add a comment there.

PR #395, opened 2018-01-20
fix:build:fixed compiler warning in maptool module

hoehnp:


navit/maptool/buffer.c
```diff
@@ -34,8 +43,16 @@ save_buffer(char *filename, struct buffer *b, long long offset)
...
-void
+/**
+ * @brief Loads a buffer from a file
+ *
+ * This function loads a buffer from a file. 
+ *
+ * @param filename The name of the while to where the buffer is loaded from.
+ * @param b Buffer in which file is loaded.
+ * @param offset
+ */
```
pgrandin: Can you please document the return value too?

navit/maptool/buffer.c
```diff
@@ -54,16 +71,24 @@ load_buffer(char *filename, struct buffer *b, long long offset, long long size)
...
+	return 1;
 }
-
+/**
+ * @brief Determines size of buffer for file 
+ *
+ * This function determines the size of the buffer required to read a file. 
+ *
+ * @param  filename Name of file for which the required size of the buffer is determined
+ */
```
pgrandin: Can you please document the return value too?

navit/maptool/maptool.c
```diff
@@ -141,7 +143,10 @@ progress_time(void)
 	seconds%=60;
 	pos+=assafe_strcp2buf(seconds>9?":":":0", buflen-pos, buf+pos);
 	pos+=assafe_lltoa(seconds, buflen-pos, buf+pos);
-	write(2,buf,pos);
+	write_result = write(2,buf,pos);
+	if (write_result == -1){
```
pgrandin: we can probably avoid the need of the write_result var and check the return directly via if ()

navit/maptool/tile.c
```diff
@@ -405,7 +405,10 @@ write_aux_tiles(struct zip_info *zip_info)
 		assert(buffer != NULL);
 		f=fopen(at->filename,"rb");
 		assert(f != NULL);
-		fread(buffer, at->size, 1, f);
+		
+	        if (fread(buffer, at->size, 1, f) == 0){
+			dbg(lvl_warning, "fread failed");
```
pgrandin: we should probably skip to the next iteration if the read failed, not forgetting to close (f)

navit/maptool/buffer.c
hoehnp: Can you please remind me wheat it returns? Don't know the function ftello.

navit/maptool/zip.c
```diff
@@ -187,10 +187,13 @@ zip_write_index(struct zip_info *info)
...
 	int size=ftell(info->index);
 	char *buffer;
-
+	
 	buffer=g_alloca(size);
 	fseek(info->index, 0, SEEK_SET);
-	fread(buffer, size, 1, info->index);
+	
+	if (fread(buffer, size, 1, info->index) == 0){
+		dbg(lvl_warning, "fread failed");
```
pgrandin: We should avoid calling `write_zipmember` if `fread()` failed

jandegr:
Hi,

Since routech was removed a while ago

https://github.com/navit-gps/navit/commit/316eefa41213edf0e62fa25658674047fbadae95#diff-e459604cce7ac3c010d84d01ae881e3c

you might consider removing ch.c instead of mending it.

jkoan:
I fixed the build issue on your branche (see last commit) the problem was that ch.c was in the CMakeList.txt but the file itself was deleted. 
This also only applyes to the linux build because we build the maptool only with this build.

jkoan:
Okay now the CI failed because we can not delete ch.c because at least two functions are required by maptool.c so we need to look into this later. I would suggest to revert the last two commit (mine and @hoehnp ) and merge this PR so we can advice the other topics in an new PR.
whats your oppinion @pgrandin and @jandegr ?

jandegr:
No problem to deal with ch.c in another PR,

I just made the remark about ch.c because 

-it is an undocumented producer
-probably abandonded many years ago for whatever reason
-the consumer is already removed, so de facto it was already decided then to dump contraction hierarchies

So it appeared trivial to keep maintaining it, but sure the references to it in maptool and cmake need to be removed as well.

pgrandin:
Actually this made Coverity a bit unhappy:
```
*** CID 262887:  Resource leaks  (RESOURCE_LEAK)
/root/project/navit/maptool/buffer.c: 81 in load_buffer()
75     	
76     	fseeko(f, offset, SEEK_SET);
77     	b->base=malloc(b->size);
78     	dbg_assert(b->base != NULL);
79     	if (fread(b->base, b->size, 1, f) == 0){
80     		dbg(lvl_warning, "fread failed");
>>>     CID 262887:  Resource leaks  (RESOURCE_LEAK)
>>>     Variable "f" going out of scope leaks the storage it points to.
81     		return 0;
82             }
83     	fclose(f);
84     	return 1;
85     }
86     /**
```
```
*** CID 262886:  Resource leaks  (RESOURCE_LEAK)
/root/project/navit/maptool/tile.c: 404 in write_aux_tiles()
398     	char *buffer;
399     	FILE *f;
400     	int count=0;
401     	
402     	while (l) {
403     		at=l->data;
>>>     CID 262886:  Resource leaks  (RESOURCE_LEAK)
>>>     Overwriting "buffer" in "buffer = malloc(at->size)" leaks the storage that "buffer" points to.
404     		buffer=malloc(at->size);
405     		assert(buffer != NULL);
406     		f=fopen(at->filename,"rb");
407     		assert(f != NULL);
408     		
409     	        if (fread(buffer, at->size, 1, f) == 0){
```

PR #397, opened 2018-01-28
Remove old speech binding to proprietary tts lib 

jkoan:
Android has the TextToSpeech API since API Version 4 so we dont need to keep it in our repo.

Main reasons to drop the support for the old Version:
- proprietary
- Unused on must of the devices
- Out of date (not updated since 9 years) 

This is required by:
https://gitlab.com/fdroid/fdroiddata/merge_requests/2855

API Doc from Android:
https://developer.android.com/reference/android/speech/tts/TextToSpeech.html


hoehnp:
For android I see the point. Would it be just removed from the android build or in general? I wonder if it would be still useful for other platforms.

jkoan:
all changes in this PR are only for android

Issue #398, opened 2018-02-09
Multipolygons


jandegr:
A place to document my progress and findings on a new approch to display openstreetmap multipolygons and to get some participants to test/tweak all the existing graphics drivers to handle the multipolygons.

On the maptools side this approach just assembles all the outers and inners of a multipolygon relation into a single way instead of the old approach/attempt in navit/trac to triangulate all.

jandegr:
A sample relation : http://www.openstreetmap.org/relation/1825800

displayed with the old gtk (before the move to cairo) on windows but with the maptool tweak for multipolygons 

![gtk_old](https://user-images.githubusercontent.com/9937104/36030152-5a4ab792-0da6-11e8-908e-c5116a96cf54.JPG)

The display of the outer and inner is correct but it just shows a blue line where it moves from the outer to the inner.

same sample on android after a little tweak in android/graphics to make it "lift the pen" when moving from the outer to the inner

![android_pen_lifted](https://user-images.githubusercontent.com/9937104/36030305-ed2c3180-0da6-11e8-995e-99d4bcb2711e.JPG)

jandegr:
For maptool I used psql, so step one
https://github.com/navit-gps/navit/pull/399
allows to build a map from a psql pgsimple database, 
without the multipolygon tweaks

jandegr:
On request, the no-multipolygon version

![dry](https://user-images.githubusercontent.com/9937104/36046466-4b2ef04c-0dd9-11e8-9f1f-4d32bb715399.JPG)

jandegr:
and now a more complete sample, also the rivers 'ringvaart' and 'leie' are now showing properly, those are also multipolygon relations as well.

![wet](https://user-images.githubusercontent.com/9937104/36046717-2468a074-0dda-11e8-8312-7f618fdd4649.JPG)

jandegr:
A sample map with relations tagged with natural=water and waterway=riverbank processed.


[luxembourg.zip](https://github.com/navit-gps/navit/files/1713242/luxembourg.zip)

sleske:
Nice! The missing support for multipolygon relations was a real annoyance with Navit - loads of buildings, lakes etc. missing.

jandegr:
An attempt to modify win32 graphics

https://github.com/navit-gps/navit/commit/0652cead7cb553c8e2b417d77dd41bc2a026851b

displaying a multipolygon before adapting win32 graphics
[...]

PR #399, opened 2018-02-09
Psql

jandegr:
Extend psql with relations.
Allows to build a map from a postgresql pgsimple database.
Use -DHAVE_POSTGRESQL=TRUE to compile it.

set up a database using osmosis from a pbf
bin\osmosis.bat --read-pbf belgium-latest.osm.pbf --log-progress --write-pgsimp database="osmpgsimple" user ="postgres" password="somepassword"

to build the map
./maptool -d postgresql://postgres:somepassword@someIP:5432/osmpgsimple belgium.bin

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,8 @@ jobs:
           command: |
             bash ci/setup_common_requirements.sh
             bash ci/build_win32.sh
+      - store_artifacts:
+          path: win32/navit
   build_wince:
     docker:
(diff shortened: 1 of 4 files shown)


jandegr:
Was used as a starting point in 
https://github.com/navit-gps/navit/issues/398
to add multipolygons

jandegr:
Since the existing code did not process relations I am confident it has not been used in ages to produce a map, so these changes won't break any box or server.

PR #409, opened 2018-03-08
add:build:Add zip for wince build

jkoan:


pgrandin:
Implemented differently via #411

PR #410, opened 2018-03-08
fix:fix download center for wince

jkoan:
Repo download on the wince docker image was broken due to really old ssh lib and git client


ci/build_wince.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/sh
+mkdir /var/lib/apt/lists/partial
+apt-get update
+apt-get install ssh git
```
pgrandin: This should probably be added here instead : https://github.com/navit-gps/Dockerfiles/blob/master/wince/Dockerfile#L3

jkoan: Agreed :+1:

PR #413, opened 2018-03-12
change:osd/core compass appearance update

lains:
Update compass to look like a compass with a red handle showing north, a bit like google maps navigation.
Slightly changing the arrow pointing towards destination so that is looks like the arrow used in bows, and updating the associated green color to be less "flashy" on screen.


zintor:
Hi lains!
Thx for the update, also for the update of the doxygen comments.
Could you please upload a screenshot of the compass?
Thx.
Kind regards,
zintor

zintor:
Just a thought:
Make it sense to make the color of the two arrows customizable in navit.xml?
https://wiki.navit-project.org/index.php/OSD#compass

pgrandin:
Nice, and bonus point for updating the documentation of other functions.
Can you please attach a screenshot of the result?

lains:
This is how the compass looks like on my Android device.

screenshot_2018-03-12-15-59-06 1

The color of the south half of the compass (white on the attached screenshot) is actually coming from `opc->osd_item.graphic_fg`.
The color of the north half is red because it is the "standard" color used on compasses, but this could also be customized. I deliberately made north and south different not only by changing color, but also by drawing the north indicator as a plain polygon in comparison while south is displayed only as a contour.
For the direction of the target, I think that it is indeed a good idea to allow customization of the color so the user can choose something else than green.

lains:
Added some color customization for compass.
Example below with <osd type="compass" (...) destination_dir_color="#ff00ff" north_color="#00a0a0"/> tag in navit.xml:
screenshot_2018-03-13-07-23-03 1

jkoan:
Tanks for your contibution :smiley: we would be happy to see more PR´s from you

PR #414, opened 2018-03-12
Necessary changes to build Navit for Android SDK 23

tsadowski:
I persevered to build Navit for Android from source. Here are the changes I had to make.


jkoan:
Hello @tsadowski 
First of all thanks for contributing to the Navit project.
Could you please explain why you want to increase the sdk version? Navit trys to use the lowest version as possible in order to support even old devices. So the min api version stays low while the target api version may be increased.
For the rest of the changes I need to take a closer look.

tsadowski:
Hello @jkoan,
I don't want to increase the minSdkVersion unnecesessary. But the minSdkVersion is only an installer hint and  don't know what happens when an app using a more recent API function on an old Android. I had to use 23 as target API because Navit uses Activity.onRequestPermissionsResult new in 23. 

My goal is to improve the selection of the map storage location. I can't browse the file system on my phone because of permission restrictions Google introduced somewhere between 5.0 and 7.1. There is an explanation how it should be done:
https://developer.android.com/training/data-storage/files.html

My phone has 2 external locations, emulated and the SD card. The best method to get both locations is getExternalFilesDirs added in API level 19. I will try to implement what I want and try to maintain backwards compatibility.

Best Regards,
Torsten

tsadowski:
I had a look into the circleci output. I will try the circleci build script (contents) instead of the information from the wiki and rework my changes.

Cheers,
Torsten

jkoan:
Hi @tsadowski 
i want to know which Navit official version you tried. Because external Storage should work in Navit.
But the minAPI Version does not need to be raised. If you got Problems with your storage please leave another Issue about it here :)
But i will close this PR here because we don't want to increase the min version. But contributions are always welcome :+1:

tsadowski:
Hi @ sleske

this is an installer hint, a hint for the installer not to install.

Cheers,
Torsten

    android:minSdkVersion

    The Android system will prevent the user from installing the application if the system's API Level is lower than the value specified in this attribute.

PR #418, opened 2018-03-30
doc: fix misspell

aerostitch:
Fixing a minor typo in the man page for navit:
information is a non-count noun, so the convention is that information is both singular and plural


aerostitch:
thx! :)

PR #419, opened 2018-03-30
maptool: prevent issues when realloc fails

aerostitch:
Based on cppcheck's warning:
```
[navit/maptool/tile.c:516]: (error) Common realloc mistake: 'th' nulled but not freed upon failure
```

### DIFF
--- a/navit/maptool/tile.c
+++ b/navit/maptool/tile.c
@@ -513,9 +513,14 @@ load_tilesdir(FILE *in)
 #if 0
 			printf("subtile '%s'\n",subtile);
 #endif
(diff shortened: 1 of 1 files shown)


mdankov:
I do not think this is a very good way to go:

1. We do not want a simple message in the map build log in the case we're going to produce a broken map. Maptool would better completely fail if it's unable to produce a proper map file.

2. We're intensively use glib memory allocation functions in maptool. These functions abort the program in the case they are unable to allocate some memory. I'd suggest to re-factor maptool code to switch to g_malloc/g_realloc/g_new/g_free functions everywhere.

aerostitch:
Good point. I'll work on that.

PR #421, opened 2018-03-30
Fix various mispells

aerostitch:
See the different commit for each typo.
Pretty basic change.


aerostitch:
Hm. I actually did tha @jkoan

jkoan:
Okay, i didn't seen that :X your right

lains:
When I ran the script "itemgra.pl", I also saw that there is a "poi_bussines_service" item in garmintypes.txt
Don't know if fixing this typo may break anything though...

PR #422, opened 2018-03-31
navit/fib: cleanup unused configure* files

aerostitch:


jkoan:
As discussed in irc right now

PR #423, opened 2018-03-31
Update:icon:status_position_wait_bk adding question mark, making sharper pointer

lains:
It seemed to me more significant to have a question mark inside the icon to mean that the position is still unknown.
I have also made the pin sharper, to get closer to the cursor used in common online map websites.
The original icon:
status_position_wait_bk_orig

The new icon:
status_position_wait_bk_new

I donate these changes to the navit project, in case it can be useful for merging into trunk.


jkoan:
Thx for your contribution :) i will merge it after all tests are passed

Issue #427, opened 2018-04-02
Refactor maptool to use g_malloc/g_realloc/g_new/g_free


aerostitch:
Based on #419:

> We're intensively use glib memory allocation functions in maptool. These functions abort the program in the case they are unable to allocate some memory. I'd suggest to re-factor maptool code to switch to g_malloc/g_realloc/g_new/g_free functions everywhere.

And on https://wiki.navit-project.org/index.php/Programming_guidelines

Just putting this as a reminder to work on it. I'll do it.

PR #429, opened 2018-04-03
maptool: migrate malloc/free/realloc to g_malloc/g_free/g_realloc

aerostitch:
Closes: #427.
Migrating the `malloc` to `g_malloc` and removing some asserts already done by `g_malloc`.

The other operations will come later in another PR.

### DIFF
--- a/navit/maptool/buffer.c
+++ b/navit/maptool/buffer.c
@@ -60,8 +60,7 @@ load_buffer(char *filename, struct buffer *b, long long offset, long long size)
 	long long len;
 	dbg_assert(size>=0);
 	dbg_assert(offset>=0);
-	if (b->base)
-		free(b->base);
+	g_free(b->base);
 	b->malloced=0;
 	f=fopen(filename,"rb");
 	fseeko(f, 0, SEEK_END);
--- a/navit/maptool/buffer.c
+++ b/navit/maptool/buffer.c
@@ -74,8 +74,7 @@ load_buffer(char *filename, struct buffer *b, long long offset, long long size)
 	dbg_assert(b->size>0);
 	
 	fseeko(f, offset, SEEK_SET);
-	b->base=malloc(b->size);
-	dbg_assert(b->base != NULL);
+	b->base=g_malloc(b->size);
 	if (fread(b->base, b->size, 1, f) == 0){
 		dbg(lvl_warning, "fread failed");
 		return 0;
--- a/navit/maptool/ch.c
+++ b/navit/maptool/ch.c
@@ -529,7 +529,7 @@ ch_assemble_map(char *map_suffix, char *suffix, struct zip_info *zip_info)
 	}
 	th=tile_head_root;
         while (th) {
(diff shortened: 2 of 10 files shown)


aerostitch:
@mdankov let me know if you prefer to have to review just 1 function migration at a time or if you want to wait for me to do the migration from `free` to `g_free` and `realloc` to `g_realloc`. Not sure when I'll get to that though.

mdankov:
Good work, @aerostitch !

The commit itself looks good for me so far.

But before merging, we should make sure there are no execution flows which call libc realloc() or free() on the memory allocated by a g_* function, and vice versa. 

In practice, I've never seen a failure after something like free(g_malloc(...)), but it's prohibited in glib documentation to go that way.

Is it ok if I'll review and comment your changes as you suggested (function by function), but we'll finally merge them only after completing the transition for all memory handling functions in maptool?

aerostitch:
> Is it ok if I'll review and comment your changes as you suggested (function by function), but we'll finally merge them only after completing the transition for all memory handling functions in maptool?

That does totally make sense. Thanks! :)

aerostitch:
Hi @mdankov,

I believe I finished to refactor the maptool code to use:
* g_malloc instead of malloc (and remove unnecessary checks for failed allocation)
* g_free instead of free
* g_realloc instead of realloc (and remove unnecessary checks for failed allocation)
* g_strcmp0 instead of strcmp

Those are the functions mentioned in https://wiki.navit-project.org/index.php/Programming_guidelines
Do you want me to have a look at migrating also to other GLib functions?
Are the changes I've done ok?

Thanks
Joseph

mdankov:
Hi!

Looks good.

But there's no need to check that value passed to g_free() is not a NULL.

Also, I would like to move g_strcmp0 switch to a different PR, so we can finish with memory allocation functions beforehand.

And there's no need to check for NULL in cases like this:

`if (admin_level && !g_strcmp0(admin_level, "2"))`

That's because g_strcmp0(NULL, "2") would neither crash or return 0. 

But beware, if both g_strcmp0 arguments could be NULL, comparison is most probably still needed as it may change the behaviour. That would depend on actual code flow.

Please consider the programming guidelines note about older GLib versions, don't switch every libc function to its GLib counterpart.

Thank you.

aerostitch:
Hi @mdankov thanks for the review.
I moved the strcmp-related changes to #434.
I also changed removed the check for null before `g_free`.
I only moved the functions that were mentioned in the navit programming guide is that the one you were referring to?

mdankov:
Thank you, now it looks almost complete for me.

But there's another memory allocation function which we still have not touched: strdup(). It would call some malloc() flavour internally, so we have to change it to g_strdup(), so later g_free() would be safe.

I suggest to do the change in this PR.

aerostitch:
Thanks for the review @mdankov . I did the change on the only 2 places that were using straight `strdup` instead of `g_strdup`. I also added it to the coding style guide examples as it's a commonly used one.

aerostitch:
Thx!

PR #434, opened 2018-04-07
maptool: use g_strcmp0 instead of strcmp

aerostitch:
TODO:
> there's no need to check for NULL in cases like this:
> if (admin_level && !g_strcmp0(admin_level, "2"))
> That's because g_strcmp0(NULL, "2") would neither crash or return 0.
> But beware, if both g_strcmp0 arguments could be NULL, comparison is most probably still needed as it may change the behaviour. That would depend on actual code flow.


aerostitch:
Concerning:
> But beware, if both g_strcmp0 arguments could be NULL, comparison is most probably still needed as it may change the behaviour. That would depend on actual code flow.

Currently if one of the inputs of `strcmp`, the result is undetermined and may vary from one execution to the other according to the documentation. With `g_strcmp0` ti will return 0 as the 2 arguments are equal in a consistent manner.
The places where that could eventually happen because we are comparing variables only are:
* `navit/maptool/google/protobuf-c/protobuf-c.c`
* `navit/maptool/itembin.c` but there's a verification that the variables are not NULL before
* `navit/maptool/maptool.c`
* `navit/maptool/tile.c`

What would you like to have returned if both arguments of `g_strcmp0` are NULL @mvglasow ?

aerostitch:
Rebased the branch on current trunk to be able to merge after review.

jkoan:
@aerostitch i think you need to rebase this PR

jkoan:
Update? Merge? @aerostitch in the description it's said todo, is everything finished and able to. Merge?

aerostitch:
Everything is done on my side and ready to merge. Just waiting for validation.

PR #435, opened 2018-04-11
[WIP] Fix:bookmarks:Ensure that bookmarks are synced to disk

pgrandin:
When using navit on some devices that could loose power without being properly shut down, the bookmark files was lost (content was empty).

This problem is reproduceable : running Navit on a buildroot image on the raspberry pi (for example one built using https://github.com/navit-gps/buildroot ), if the pi looses power and the bookmark file has been modified, the bookmark file will be lost everytime.

This simple fix ensures that the files get flushed to the disk when we write the bookmark files, and solves the issue according to my tests.


pgrandin:
Thanks @hoehnp 
I forgot to make it windows compliant too - will fix that, and renamed the PR as `WIP` meanwhile.

jkoan:
Will open a new Pr for this

PR #235, opened 2017-05-03
Make Image and Scale osd types obey osd_configuration setting

pohlinkzei:
This PR shall fix [trac-1202](http://trac.navit-project.org/ticket/1202)

PR #439, opened 2018-04-13
PR to fix Trac 1202 (Make Image and Scale osd types obey osd_configuration setting)

pohlinkzei:
This PR shall fix [trac-1202](http://trac.navit-project.org/ticket/1202).
After looking into it again after a year, I could fix the behavior for Image and Scale.

Additionally I renamed the OSD flags to what they probably do.

I took a new branch (with typo -.- ) because we do not want to see all commit from the past  year. 

#235 can stay closed.

### DIFF
--- a/navit/osd/core/osd_core.c
+++ b/navit/osd/core/osd_core.c
@@ -539,6 +539,7 @@ osd_route_guard_new(struct navit *nav, struct osd_methods *meth,
 	opc->osd_item.font_size = 200;
 	opc->osd_item.meth.draw = osd_draw_cast(osd_route_guard_draw);
 	meth->set_attr = set_std_osd_attr;
+	// TODO: DEFINE & NAME FLAGS
 	osd_set_std_attr(attrs, &opc->osd_item, 2);
 
 	attr = attr_search(attrs, NULL, attr_min_dist);
@@ -968,8 +969,9 @@ osd_odometer_reset(struct osd_priv_common *opc, int flags)
 {
   struct odometer *this = (struct odometer *)opc->data;
 
-  if(!this->bDisableReset || (flags & 1)) {
-    if (!(flags & 2)) 
+	// TODO: DEFINE & NAME FLAGS
+  if(!this->bDisableReset || (flags & FLAG1)) {
+    if (!(flags & FLAG2)) 
       this->bActive         = 0;
     this->sum_dist        = 0;
     this->sum_time        = 0;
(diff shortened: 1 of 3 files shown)


navit/osd/core/osd_core.c
```diff
@@ -3901,14 +3909,14 @@ osd_scale_draw(struct osd_priv_common *opc, struct navit *nav)
 
 	if (!navit_get_attr(nav, attr_transformation, &transformation, NULL))
 		return;
-	if (this->use_overlay) {
+	//if (this->use_overlay) {
```
pgrandin: We should probably delete code instead of commenting it

navit/osd/core/osd_core.c
```diff
@@ -3943,14 +3951,14 @@ osd_scale_draw(struct osd_priv_common *opc, struct navit *nav)
 	graphics_draw_rectangle(opc->osd_item.gr, opc->osd_item.graphic_fg, p+6, 4,opc->osd_item.h/5+4);
 	graphics_draw_rectangle(opc->osd_item.gr, opc->osd_item.graphic_fg, p+7, p[1].x-p[0].x, 4);
 	graphics_draw_rectangle(opc->osd_item.gr, opc->osd_item.graphic_fg, p+8, 4,opc->osd_item.h/5+4);
-	graphics_draw_lines(opc->osd_item.gr, this->black, p, 2);
-	graphics_draw_lines(opc->osd_item.gr, this->black, p+2, 2);
-	graphics_draw_lines(opc->osd_item.gr, this->black, p+4, 2);
+	graphics_draw_lines(opc->osd_item.gr, opc->osd_item.graphic_fg_text /*this->black*/, p, 2);
```
pgrandin: Same here, it's probably not worth keeping the old code

pgrandin:
It's nice to get another contribution from you @pohlinkzei!
Would you mind attaching a screenshot also showing the improvement of your PR?

Thanks!

navit/osd/core/osd_core.c
pohlinkzei: Ha! I obviously missed those two....

pohlinkzei:
Yes, I Will.

pohlinkzei:
Complete config file to be found here: https://github.com/pohlinkzei/Pi_MFD_Layout/tree/trac1202

pgrandin:
Nice work

PR #444, opened 2018-04-19
New version for sailfish 0.5.1-r5

metalstrolch:
This bumps version for navit to 0.5.1-r5

This removes gcc from build dependencies

This automatically zyppers in build requirements as given in spec file
into sailfish build virtual machine. So no hand work required anymore.


jkoan:
Only relevant for sailfish --> merging

PR #445, opened 2018-04-19
Update CHANGELOG.md for 0.5.1 and add configurations for gitchangelog script

aerostitch:
I've used https://github.com/vaab/gitchangelog, which is based on commits only (I've tested a bunch of other ones but as we're not always using a consistent commit message pattern and that we don't always use PR, that's the best option I found).
All the commits about CI/CircleCI have been skipped.
Let me know if that is ok for you. If so I'll push `0.5.1`.

I plan to try to do the next releases every months or so, so the changelog will be way smaller.
Plus, every time we bump a minor or a major version we could archive the changelog of the previous version in a file named `CHANGELOG-0.5.md` for example (we can figure that one out later).


jkoan:
I really much like this. In the future we also need to update Authors file. My idea would be to use the git log again and do it with a .mailmap file

PR #448, opened 2018-04-20
Add:CI:Re-add the merge to master job when circleci runs on trunk and is successful

aerostitch:
Currently master is about 190 commits behind trunk.
I think this should work


pgrandin:
Let's see!

PR #451, opened 2018-04-20
Change:CI:Move doxygen to the build workflow to only run it when the build succeeds

aerostitch:
@jkoan pointed out with good reasons that it would be better to only run the doxygen job when the builds are all passing. This is the corresponding PR.

Putting @pgrandin as reviewer to make sure there were no specific reasons to put this one in a separate workflow.


pgrandin:
It was done this way to ensure that doxygen was always up to date with the code in trunk, even if the code in trunk fails to build. But now that we're leveraging pull requests more, build errors should be caught before the code reaches trunk, so I'm good with that.

PR #452, opened 2018-04-20
[WIP] Add:Dbus:Added code to route to a bookmark via dbus

pgrandin:
This code allows us to trigger a route calculation to a bookmark via Dbus.

### DIFF
--- a/navit/binding/dbus/binding_dbus.c
+++ b/navit/binding/dbus/binding_dbus.c
@@ -1436,6 +1436,59 @@ request_navit_route_export_gpx(DBusConnection *connection, DBusMessage *message)
 }
 
 /**
+ * @brief Look for the bookmark passed in parameters, and route to it if found
+ * @param connection The DBusConnection object through which a message arrived
+ * @param message The DBusMessage including the 'bookmark_name' parameter
+ * @returns An empty reply if everything went right, otherwise DBUS_HANDLER_RESULT_NOT_YET_HANDLED
+ */
+static DBusHandlerResult
+request_bookmarks_set_as_destination(DBusConnection *connection, DBusMessage *message)
+{
+        char * bookmark_name;
+        struct navit *navit;
+        DBusMessageIter iter;
+
+        struct attr attr,mattr;
+	struct navigation * nav = NULL;
+	struct item *item;
+	struct coord c;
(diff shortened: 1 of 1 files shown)


pgrandin:
Sure:
```
dbus-send  --print-reply --session --dest=org.navit_project.navit /org/navit_project/navit/default_navit org.navit_project.navit.bookmarks.set_as_destination string:"Test"
```

But it looks like sometimes this call will fail here and segfault :
https://github.com/navit-gps/navit/blob/d9de140c3d3f02198e53c2d251db93cf3d10db27/navit/binding/dbus/binding_dbus.c#L1472

```
#0  0x00007f7469c5595e in waitpid () from /lib64/libc.so.6
#1  0x00007f7469bd4f6b in ?? () from /lib64/libc.so.6
#2  0x000055676f93d84c in sigsegv (sig=<optimized out>) at /home/pierre/dev/navit/tmp/navit/navit/debug.c:83
#3  <signal handler called>
#4  item_attr_get (it=0x70e96e38, attr_type=attr_label, attr=0x7ffde00399d0) at /home/pierre/dev/navit/tmp/navit/navit/item.c:258
#5  0x00007f7469534a58 in request_bookmarks_set_as_destination (connection=0x556770e91860, message=0x556770e91ec0) at /home/pierre/dev/navit/tmp/navit/navit/binding/dbus/binding_dbus.c:1473
```

I need to look into it.

jkoan:
Hi @pgrandin 
I think it is possible to give the same name to more then one bookmark, what would then happen?

Generally, what about rework the bookmarks and give everyone a uuid so we could call them in the api/dbus.
Then we could also switch to some more readable format or something less complex (in the meaning of faults) like sqllite.

lains:
Maybe this be implemented using two calls:

- One that gets details (including GPS coords) from a bookmark, that would be something like `bookmark_search`
- A second one that routes to a GPS coord (using the already existing `set_destination` Dbus method.

This would allow to have common code for the destination set, and start a whole family of methods for manipulating bookmarks (there could also be `bookmark_rename`, `bookmark_delete`, `bookmark_export_as_xml`, `bookmark_export_as_json`...)

jkoan:
Closing for now since we need to think again about this and rework it like @lains said ;)

Issue #455, opened 2018-04-20
Wince build is broken


pgrandin:
The wince build is failing and the CI is not catching the issue :
```
Linking C executable navit.exe
gui/internal/libgui_internal.a(gui_internal_command.c.obj): In function `gui_internal_cmd2_route_height_profile':
/root/project/navit/gui/internal/gui_internal_command.c:584: undefined reference to `fmax'
/root/project/navit/gui/internal/gui_internal_command.c:585: undefined reference to `fmin'
collect2: ld returned 1 exit status
make[2]: *** [navit/navit.exe] Error 1
make[1]: *** [navit/CMakeFiles/navit.dir/all] Error 2
make: *** [all] Error 2
cp: cannot stat `navit/navit.exe': No such file or directory
cp: cannot stat `navit/navit.xml': No such file or directory
cp: cannot stat `locale/': No such file or directory
cp: cannot stat `navit/maps/*.bin': No such file or directory
cp: cannot stat `navit/maps/*.xml': No such file or directory
  adding: icons/ (stored 0%)
  adding: icons/Makefile (deflated 78%)
  adding: icons/desktop_icons/ (stored 0%)
  adding: icons/desktop_icons/navit.desktop (deflated 41%)
  adding: icons/cmake_install.cmake (deflated 81%)
  adding: maps/ (stored 0%)
Source dir: wince/
PocketPc dir: navit
ls: cannot access wince//locale/*/*/*.mo: No such file or directory
cp: cannot stat `wince//navit/icons/*.xpm': No such file or directory
```

From https://circleci.com/gh/navit-gps/navit/6154#artifacts/containers/0

This creates an almost-empty zip file.

PR #456, opened 2018-04-20
Fix:CI:Make sure the wince build fails when make or cmake fail

aerostitch:
Closes: #455
1st run should fail on wince, then I'll work on the underlying issue.

### DIFF
--- a/ci/build_wince.sh
+++ b/ci/build_wince.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
 
 mkdir wince && cd wince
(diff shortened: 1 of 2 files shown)


aerostitch:
Ok, so the 1st commit works as expected. :)

aerostitch:
The wince issue has been introduced by https://github.com/navit-gps/navit/commit/54dac2a0f22efbe15d4b081d69881a344bdc0419 in #441
@jandegr FYI

aerostitch:
So I fixed the inner issue (the absence of `fmin`/`fmax` on wince) by using the `MIN`/`MAX` macros we defined in gmacros.
Now it works and I don't think that will be a problem because we are comparing int.
@jandegr can you confirm that this change is ok for you please?

Thanks
Joseph

pgrandin:
Nice, thanks @aerostitch for taking care of that!

aerostitch:
`set -e` works too.
I updated the PR to use `set -e`, it works as expected.

PR #460, opened 2018-04-21
Update:CI:Make sure that all the scripts in the ci folder have a shebang and use set -e

aerostitch:
Updated all the ci scripts to have a shebang and use `set -e` except for `ci/update_download_center.sh` who has a retry mechanism that prevents us from using `set -e`.

Also fixes an error happening when extracting the android sdk as the owner of the extracted files is not the same as in the tarball.


pgrandin:
Nice! Thanks!

PR #539, opened 2018-04-23
CI:Android:Use a pre-built docker image for android toolchain

pgrandin:
Using a pre-built docker image ( https://github.com/navit-gps/Dockerfiles/blob/master/android/Dockerfile ) we can reduce the build time by ~2:30 minutes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,17 +38,10 @@ jobs:
       - store_artifacts:
           path: /root/project/doc
   build_android_arm:
-    <<: *defaults
+    docker:
+      - image: navit/android
     steps:
       - checkout
-      - run:
-          name: Setup common requirements
-          command: |
-            bash ci/setup_common_requirements.sh
-      - run:
-          name: Prepare the Android build environment
-          command: |
-            bash ci/setup_android.sh
       - run:
           name: Build for Android ARM
           command: |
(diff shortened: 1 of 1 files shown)


jkoan:
The idea is really good, but you need to update the software inside the image on ci side to make sure we not stuck on a old version of the android tools.

So my idea would be to use the setup scripts which where removed in this PR also to be sure that everything is up to date.

pgrandin:
> The idea is really good, but you need to update the software inside the image on ci side to make sure we not stuck on a old version of the android tools.
I'm not sure I understand that.

This image was built exactly the way we are setting up android in the CI : ubuntu 14.04 + a bunch of packages and then some commands.
When it'll be time to support android-26, we just need to add it to the Dockerfile and update the image tag that we use in the CI.

Does it answer your concern?

jkoan:
Not really. The point is that Docker images are static. So thy only update when you update them manually.
For us that means when Google pushes out bug fixes for for example a support library we will not see that on the next build.
there are two ways to solve this problem. First one is to update the image automatically when something changes (which is quite hard to figure out)
or the more easy one would be to use the outdated image and update (not install) on runtime.

So usage of the new image and using the scripts would fix problems we could have with a minimal add of time and we still have the advantage of pre-installed components. The scripts will only update things if they are not on the latest version.

jandegr:
Please consider using gradle with a prebuilt image for the build, faster, better and almost maintenance-free,
just bump the used image from time to time.
Already did a bit of testing over there https://github.com/navit-gps/navit/tree/Android_gradle

pgrandin:
@jandegr looking at your branch, it actually replaces https://github.com/navit-gps/navit/pull/539 
Do you have any reason for not submitting a pull request with that branch? Your approach seems better.

jkoan:
I would also like to do it with gradle instead when its ready. So probably close this one @pgrandin ?

jandegr:
I still have a bit of work on cmake to make it work not just on CI but in androidstudio as well, but then for the signing and publishing and such it will require your help / intervention, so I'd suggest to open an issue for it since it is not ready for a PR yet.

PR #540, opened 2018-04-23
Fix:CI:Automated upload to playstore

pgrandin:
This PR fixes automated uploads to playstore from trunk.
Tested in https://circleci.com/gh/navit-gps/navit/6401

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,6 +85,15 @@ jobs:
       - store_artifacts:
           path: android-x86/navit/android/bin
       - run:
+          name: Install git
+          command: apt-get update && apt-get -y install ca-certificates git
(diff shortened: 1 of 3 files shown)


pgrandin:
> Could we change the order so update download center is on the end? Then we can also upload the release signed apk.
Sure

> Also installing new software should be path of the setup requirements I think
Depends on what we do with https://github.com/navit-gps/navit/pull/539 :)

pgrandin:
@jkoan better now?

pgrandin:
> Your update to Navit for Android, created on Apr 25, 2018 at 4:24 PM GMT, is live in the store.

From https://circleci.com/gh/navit-gps/navit/6794

It works :D

PR #546, opened 2018-04-25
Fix:debug:Change line separators for dbg to work also on win* platform

### LINKED ISSUE #545: Change line separators for dbg to work also on win* platform
1. Remove the return char at the end of the existing strings in dbg calls
2. Detect the right EOL char and add it in dbg directly
3. optionally replace `\n` by the right EOL if the OS's EOL is different and a `\n` is detected at the end of the string

aerostitch:
Closes #545


aerostitch:
Thank you guys! :)

PR #552, opened 2018-04-27
Update:xmlconfig:Move from malloc/free/strncpy to g_malloc/g_free/g_strndup

aerostitch:
So I wasn't sure if we should g_strndup instead of strncpy or directly g_strdup and drop the string length altogether. Any thoughts?
Also does the xml support utf8? In this case I would have to use g_utf8_strncpy.
Thanks!


aerostitch:
I think we could remove the things in:
```C
#if 0
...
#endif
```
Let me know if you want to do that.

aerostitch:
thx!

PR #553, opened 2018-04-27
[WIP]Android:gradle

jandegr:
The aim is to allow building with gradle for Android

- builds a multi-arch apk in one container on CI in less time than one of the 2 containers now.
- can be imported as an Androidstudio project as-is, CI builds the resources as well but there are no provisions to do so locally, for now the fastest way is to grab the resources from an existing apk
- gradle/androidstudio is the only tool supported by Google.
tested on CircleCI and win10, will probably work as-is on linux and osX

feel free to comment


jkoan:
@jandegr could you give me some hints how I can set up the Navit project within androidstudio?

CMakeLists.txt
```diff
@@ -703,10 +703,11 @@ if(ANDROID)
    set(HAVE_API_ANDROID 1)
    set_with_reason(speech/cmdline "Android detected" FALSE)
    set_with_reason(graphics/null "Android detected" FALSE)
-   set_with_reason(graphics/android "Android detected, NDK:${ANDROID_NDK_API_VERSION}, API:${ANDROID_API_VERSION}" TRUE)
-   set_with_reason(speech/android "Android detected, NDK:${ANDROID_NDK_API_VERSION}, API:${ANDROID_API_VERSION}" TRUE)
-   set_with_reason(vehicle/android "Android detected, NDK:${ANDROID_NDK_API_VERSION}, API:${ANDROID_API_VERSION}" TRUE)
+   set_with_reason(graphics/android "Android detected TRUE)
```
jkoan: Ending quotes are missing

jandegr:
[DRAFT] How to imoprt it as an Androidstudio project

- provide the sources of this branch in you androidstudioprojects folder
- provide the resources, either by doing the same conversions as in the CI script if you're on Linux or by grabbing them from a prebuilt apk as a shortcut on Linux or the most practical way on Windows
- in the startupscreen select "Import project (Gradle, Eclipse ADT, etc.)
- ...
- ...
- restart Androidstudio, triggers some necessery step (sync ?)
- ...

pgrandin:
@jkoan and @jandegr : that androidstudio doc would be a nice addition to the wiki ;)

mvglasow:
I’ve started a small writeup at https://wiki.navit-project.org/index.php/Android_development#With_gradle (focus is on building Navit on a Linux box without Android Studio installed).

I recently tried to build with gradlew and have found two issues with the build:

* Both `vehicle/gpsd` and `map/garmin` cause it to fail with errors. Are they supposed to be included in the Android build at all? They currently get enabled because the respective libraries are found, but I’m not sure if these results are accurate for cross-compilation. Maybe you can have a look at that. I managed to get the build to complete without errors by editing `CMakeLists.txt` and disabling these two specifically for Android.
* After running `gradlew build` and `gradlew installDebug`, Navit installs fine but all icons are missing from the internal GUI (menu items, menu title bar, dialogs) as well as from the map. OSD icons display, though. The CircleCI build is OK, just my local gradle build is having issues.

lains:
I had the same issue with gradle, and also had to disable detection of gpsd or I was getting an failure during compilation on `#include "gps.h"`
At some stage, my build was also missing files, the best way to make sure you have everything is to manually perform all the steps in `.circleci/config.yml`

mvglasow:
Thanks for the heads-up, I’d tried to find the CI stuff but missed the `.circleci` dir.

To sum it all up, there seems to be a lot of stuff which has to be done outside of gradle, though the old CMake setup would do everything automatically… not great. Gradle gurus, is there any way to have gradle trigger these things?

Otherwise, yes, a build script for Android is probably the best way to go. However, I’d advise keeping build dependency installation out of the build script—check if dependencies are there, and if not, exit with an error telling the user how to install them. Optionally, add another script to install all build dependencies.

jkoan:
For now this should do the trick:
(please adapt at least the code location as needed)
```
WORKDIR=$(mktemp -d)
cd $WORKDIR
git clone https://github.com/navit-gps/navit
export JVM_OPTS="-Xmx3200m"
export GRADLE_OPTS='-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
cd navit
sudo apt-get install cmake gettext libsaxonb-java librsvg2-bin
cmake ./ -Dsvg2png_scaling:STRING=-1,24,32,48,64,96,128 -Dsvg2png_scaling_nav:STRING=-1,24,32,48,64,96,128 -Dsvg2png_scaling_flag:STRING=-1,24,32,64,96
cd navit/icons
make
mkdir ../android/res/drawable-nodpi
rename 'y/A-Z/a-z/' *
cp *.png ../android/res/drawable-nodpi
cd ../../
cd po
make
mkdir ../navit/android/res/raw
rename 'y/A-Z/a-z/' *
cp *.mo ../navit/android/res/raw
cd ../
cd navit
saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitldpi.xml OSD_SIZE=1 ICON_SMALL=24 ICON_MEDIUM=32 ICON_BIG=48
saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitmdpi.xml OSD_SIZE=1.33 ICON_SMALL=32 ICON_MEDIUM=48 ICON_BIG=64
saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navithdpi.xml OSD_SIZE=2 ICON_SMALL=48 ICON_MEDIUM=64 ICON_BIG=96
saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitxhdpi.xml OSD_SIZE=2.67 ICON_SMALL=64 ICON_MEDIUM=96 ICON_BIG=128
saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitxxhdpi.xml OSD_SIZE=4 ICON_SMALL=96 ICON_MEDIUM=128 ICON_BIG=192
saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitxxxhdpi.xml OSD_SIZE=5.33 ICON_SMALL=128 ICON_MEDIUM=192 ICON_BIG=256
mv *dpi.xml android/res/raw
cd ../
chmod +x ./gradlew
./gradlew -v
sdkmanager ndk-bundle
./gradlew assembleDebug
ln -s navit/navit.dtd navit.dtd
mkdir /home/circleci/code/test-results
./gradlew lint test

echo Build finished
echo Apk should be in "navit/android/build/outputs/apk"
find navit/android/build/outputs/apk -iname "*.apk" 
```

mvglasow:
Just tried the following:

* saved this script as `scripts/build-android-gradle.sh` (shebang added)
* removed all `apt-get`, `sdkmanager` and `git clone` occurrences (no installations from a build script, and the script is meant to be run from an already existing repo)

I get an error about `rename` not being found, and the following message from gradle:

```
> Task :navit:android:mergeDebugResources FAILED
/home/michael/src/navit/navit/android/res/raw/zh_TW.mo: Error: 'T' is not a valid file-based resource name character: File-based resource names must contain only lowercase a-z, 0-9, or underscore
/home/michael/src/navit/navit/android/res/drawable-nodpi/country_US_96_96.png: Error: 'U' is not a valid file-based resource name character: File-based resource names must contain only lowercase a-z, 0-9, or underscore
```

Something isn’t working yet…

Also, I would suggest exiting as soon as a command fails, something like


```
command_which_may_fail || exit 1
```

Advantages:
* no waste of dev time and CPU cycles if something early in the chain fails
* continuing after a failure usually does not produce a usable result
* running more stuff after a failure clutters up the display, making the actual error harder to spot

Lastly, running this from the root of the source tree pollutes the source tree quite badly, any easy way to avoid that?

mvglasow:
Sounds good, thanks. `sudo apt-get install rename` turned out to fix the issue, and now at least I'm getting a working build. Now I'm working on a cleanup script…

mvglasow:
About the errors in components: `map/garmin` throws the following error:

```
  FAILED: cd /home/michael/src/navit/navit/android/.externalNativeBuild/cmake/debug/x86/navit/map/garmin && gentypes /home/michael/src/navit/navit/map/garmin/garmintypes.txt /home/michael/src/navit/navit/android/.externalNativeBuild/cmake/debug/x86/navit/map/garmin/g2nbuiltin.h
  /bin/sh: 1: gentypes: not found
  ninja: build stopped: subcommand failed.
```

`vehicle/gpsd` complains about not being able to find `gps.h`, this should probably be disabled on Android

lains:
@mvglasow, the fix for gps.h (and garmin gentypes) that you mentionned in https://wiki.navit-project.org/index.php/Android_development#With_gradle should probably be submitted as a new pull request that adds the following two lines to `CMakeLists.txt` as you suggested in the Wiki (BTW thanks for the tip):
```
set_with_reason(vehicle/gpsd "Android detected" FALSE)
set_with_reason(map/garmin "Android detected" FALSE)
```

This way, it will improve the build with gradle when done locally (outside of CircleCI).

jkoan:
I think we have never enforced the support for it but also never said it's not supported since the map is loaded directly from the file system over the c code.
This is what I got in mind, but I am not totally sure about it. So for Garmin we could repair it or totally drop the support, because I don't know if it's used anymore. OpenStreetMaps are quite good and they are probably the best tested map driver we have.

But for now I think exactly in android it's the best to disable it.

mvglasow:
Added in 9f57fe1 and 0e0b7a5, respectively.

mvglasow:
As for the build script, I had to modify the `saxonb-xslt` calls by adding `-strip:none`. Without it, it would mangle the XML files, stripping a bunch of whitespace characters in various places. The man page states that not stripping whitespace is default behavior, but that doesn’t seem to be the case everywhere. Adding the command line switch fixed this for me.

PR #555, opened 2018-04-28
Coverity in pr

pgrandin:
Testing ideas to integrate coverity scans with the pull requests.


pgrandin:
Bummer, doesn't work as expected. The checks ran when i pushed the commits to the branch, and the PR isn't triggering a new build, so $CIRCLE_PR_NUMBER is never set.

jkoan:
I think the CIRCLE_PR_NUMBER is only raised when a new PR is made (not tested)

jkoan:
Still relevant?
I tested some time ago that the Variable is only set when the PR is not from the same repo

PR #558, opened 2018-04-30
cleanup:global:Cleanup some unreachable code

aerostitch:
This PR removes some sections of code unreachable because in a:
```C
#if 0
...
#endif
```
block.

Note that I didn't remove the blocks that were having complete functions or the ones that seemed to be notes for later improvements.


aerostitch:
Awesome, thanks for the fix! ;)

PR #559, opened 2018-04-30
Refactoring:Xmlconfig:Moving layout definition in their own .xml file (makes main navit.xml file lighter)

lains:
Layouts contain quite heavy xml code, and there are many layouts available for navit.
They are all inserted inside the main navit.xml, which makes it hard to edit because of its size, even if changes or customizations by users are very rarely on the layout code.
I have thus moved the layout code away from navit.xml, each layout having its own xml definition file, called navit_layout_*.xml
These files are inserted inside the main navit.xml file by using the already existing xi:include mechanism, this also allow for backwards compatibility (old monolithic navit.xml files are still valid and can be used).
The other advantage for this is that f the user wants to have his/her own customized navit.xml, he/she can still include the shipped layout files, making their xml lighter. This also allow to enable/disable specific layouts easily by including or not each layout file.
It is also easier to perform side-by-side comparison between two layout files.
I have run an auto-indenter on XML files so that they all use tabs for indentation. File size in lines are now:

- 449 bytes for `navit_shipped.xml`
- 1052 bytes for `navit_layout_bike_shipped.xml`
- 1338 bytes for `navit_layout_car_android_shipped.xml`
- 1725 bytes for `navit_layout_car_dark_shipped.xml`
- 2015 bytes for `navit_layout_car_shipped.xml`
- 122 bytes for `navit_layout_car_simple_shipped.xml`
- 547 bytes for `navit_layout_th_shipped.xml`

This has been tested and works well on the Android and Linux builds. Test should be done on other platforms, especially the files that are packaged should include the new navit_layout*.xml but I think they all use the xml_config target in the Makefile, and this seems to be common amongst all platforms except Android.

There is maybe some work to do on the NavitConfigurator tool to handle this new type of config file, I haven't checked... however, I have made sure that the sanity tool navit/script/itemgra.pl was updated to handle the standalone layout files.


navit/navit_shipped.xml
```diff
@@ -419,6897 +426,18 @@ Waypoint</text></img>
...
-					<icon src="communication.xpm"/>
-				</itemgra>
-				<itemgra item_types="poi_concert" order="12-">
-					<icon src="concert.xpm"/>
-				</itemgra>
-				<itemgra item_types="poi_cove" order="12-">
-					<icon src="cove.xpm"/>
-				</itemgra>
-				<itemgra item_types="poi_crossing" order="12-">
-					<icon src="crossin
```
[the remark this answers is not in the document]
hoehnp: very nice idea. Later on I would suggest that later on we offer an option to also download custom layouts which we provide maybe from a separate repository.

pgrandin:
Keeping a monolithic xml file vs splitting it into subfiles has always been a hot topic.
There's actually a compelling use case to split it : on some devices it looks like our current 360Kb xml file is making the device run out of memory. We've noticed that last week on DXOS's wince device. Removing unused routing profiles and layouts to shrink the file to ~60Kb fixed that issue.

That being said I'd like to get @sleske and @mdankov 's comments on this.

jandegr:
Hi,
XML certainly needs to be worked on but I have a few questions 

Why are the layoutnames hardcoded in navit.java ? Up to now they were retreived from xml, any graphics artist could work on the layouts without hardcoding.

Furthermore I have the impression this PR does a lot more than what the title says but now it is organized as a take all or leave all PR

thx.

navit/navit_shipped.xml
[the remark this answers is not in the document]
lains: Yes, this could be tried, I think it should work.

lains:
Hi @jandegr.

> Furthermore I have the impression this PR does a lot more than what the title says but now it is organized as a take all or leave all PR

Most of the changes are actually linked to the split of navit.xml in separate files.
Maybe the changes in navit/xmlconfig.c are not stricly needed because I only needed it initially when I did not set NAVIT_ACTIVECONFDIR and only had layout files location relative to the navit.xml.
I thought anyway I would keep it because have relative path in xi:include expanded to the parent's directory seems logical to me (rather than to be relative to the cwd of the process as it seems it is the case without this change).

> Why are the layoutnames hardcoded in navit.java ? Up to now they were retreived from xml, any graphics artist could work on the layouts without hardcoding

I agree, hardcoding layout files in the Java code is not that great. In order to get rid of this, it requires being able to browse the content of the package from the Java...
It seems it cannot be done when data is stored in res/raw, but only when using [AssetManager](https://developer.android.com/reference/android/content/res/AssetManager). I have to say I started to have a look, but this seemed to me even more work...
However, if this PR is OK for merge, I think I can implement this so that there is no layout name hardcoded in Java anymore. Just let me know before I start working on this.

To go even further, I would say that it would be better to avoid hardcoding layout names in CMakeLists.txt's variable XMLCFG_ALL_FILE_PREFIXES but rather use a wildcard... I did not go that far either...

aerostitch:
I'm wondering if it could be easier to use some templating engine like jinja (or literally anything else) to generate a minimal navit.xml based on the configuration options on the build or even at install time for some platforms.
I guess that if we split the xml we would generate a zip or tarball with all the config files to be able to distribute it the same way we do today.
Another way to do would be to have:
* a configuration manager (that you would contact via RPC or shared memory calls or whatever you think is the most appropriate when you need to be able to call it from both C and Java) that would have sane defaults based on your platform & current compilation options
* a little program that would generate the xml based on values you choose like a UI that proposes a list of possible fields and the possible values or a textbox with the current defaults.
This way navit.xml would only contain stuffs that the user himself really wants customized (either via the xml config generator or via the program options), which would decrease greatly its size. I think a lot of user rarely touch the xml manually.

lains:
@aerostitch, I'm not sure how I can add or modify the notes at https://github.com/navit-gps/navit/projects/7

One note on:

> On compilation we should have the navit xml config generated based on the compilation parameters and the target platform

From what I have seen during my work in this PR, it seems this is already (at least partly) done when appying xslt transformation to navit_shipped.xml templates during build

lains:
Let me also explain why there are so many files modified.
Splitting navit.xml in several files means shipping more files, thus the change in CMakeLists.txt
Also, during build, an xlst transformation is applied to navit.xml, I had to update it to be able to also transform navit_layout*.xml, thus the change in android.xslt
It also means extracting more than just the navit.xml file from the archive in Android, thus the change in Navit.java
I tried to reuse the existing xi:include mechanism, but the issue is that includes seem not to be relative to the XML file that use xi:include, but rather to the current working dir of the process, this is why I changed  xmlconfig.c, file.h and file.c
Later on, I thought I would directly go for an absolute path in the xi:include, but this path cannot be hardcoded, as it depends on where navit finds its navit.xml file and this is guessed at runtime.
I thus added a new variable NAVIT_ACTIVECONFDIR (change in main.c), and I need to set it to the correct value before parsing the config (change in start_real.c)
Finally, I did a grep on what points to navit.xml and I found the script itemgra.pl that I updated to be able to parse current navit.xml file or navit.xml that is using xi:include. I also a script called check_itemdef, but it seemed it did the same as itemgra.pl, so I updated only itemgra.pl (check_itemdef would not be able to parse navit.xml files with xi:include)

navit/navit_shipped.xml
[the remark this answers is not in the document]
lains: It works, I have updated the code in the PR accordingly. Thanks.

aerostitch:
Hi @lains,

> I'm not sure how I can add or modify

You probably need to be in the org for that.

> From what I have seen during my work in this PR, it seems this is already (at least partly) done when appying xslt transformation to navit_shipped.xml templates during build

Yes, but we don't do it for every parameters, for example on wince we still have `<graphics type="gtk_drawing_area"/>` which is not really adapted for win*-based configurations.

For the rest, I'll leave it to people with stronger experience in that code.

Thanks,
Joseph

lains:
> I'm wondering if it could be easier to use some templating engine like jinja (or literally anything else) to generate a minimal navit.xml based on the configuration options on the build or even at install time for some platforms.
> I guess that if we split the xml we would generate a zip or tarball with all the config files to be able to distribute it the same way we do today.
>
> Another way to do would be to have:
> *    a configuration manager (that you would contact via RPC or shared memory calls or whatever you think is the most appropriate when you need to be able to call it from both C and Java) that would have sane defaults based on your platform & current compilation options
> *    a little program that would generate the xml based on values you choose like a UI that proposes a list of possible fields and the possible values or a textbox with the current defaults.
> *    This way navit.xml would only contain stuffs that the user himself really wants customized (either via the xml config generator or via the program options), which would decrease greatly its size. I think a lot of user rarely touch the xml manually.

I like a lot the second option.
This way, the user selects custom options for his/her needs, the rest remains inherited from the shipped config.
Maybe xslt could be used there again, to store only diffs with navit.xml, but this means shipping the xslt processor on the target to reapply user-defined changes at runtime after each upgrade of the package...
... or just rework the whole config format to something else than can be parsed a better way, but this would be much more work.

The initial scope of my PR was much more humble to be honest.
It was a first step in the way to make the main xml lighter so that it is less scary and can be edited a bit better.
Most of the lines counts in navit.xml consists of layouts data... and there are several isolated layouts concatenated, so this was the rationale for deciding that each could fit in its separate file. This makes navit.xml short, and by using xi:include, barely changes the way parsing is done. Also, it allows adding/removing layouts easily.
This was my initial proposal (quick and easy update, backwards compatible for people that would already have a monolithic navit.xml file of their own).

Does someone know if any user edits layout data? If not, then, they would only customise the OSD and a few other parameters in the main navit.xml, and just point to a system-wide folder of shipped layouts for inclusion.
In that case, maybe I need to add another variable called `NAVIT_DEFAULTCONFDIR` that points to the default layout folder, this way, it is only necessary to copy navit.xml in the user folder, and make adjustments. Maybe copying the shipped navit.xml to a user folder can even be done using a menu in the GUI. The XML editing would then be done using a third party application.

mvglasow:
I like the idea of separating layouts from the main XML file.

As for hardcoded layout names, yes, we should find a way to eliminate that (I believe hardcoding them would negate the main benefit—being able to just drop in extra layout files).

One question: a bunch of platforms ship a default XML config file, which gets ignored if a user-supplied file is found in a designated location (which differs from platform to platform) or supplied on the command line. How would this mechanism behave with layouts separated into their own files?

In a perfect world:

* I could just drop my own layouts into the designated location for them, even if I use the default config XML, and choose between my custom layouts and the default ones
* If name collisions are an issue, user-supplied layouts take precedence over default ones
* I could customize navit.xml and have an easy way to get the same behavior (e.g. by copying the include stuff from the original XML file)

Would these things work?

About users editing layout data—I am definitely one of them, using a customized layout based on the [0606.at layout](https://wiki.navit-project.org/index.php/Android#0606.at_Android_Layout) (my modifications include scaling it for HDPI and creating a corresponding night layout).

lains:
> As for hardcoded layout names, yes, we should find a way to eliminate that (I believe hardcoding them would negate the main benefit—being able to just drop in extra layout files).

(1) I'm checking this in the Java code. I think this is feasible. I will commit some more code to handle this.

> In a perfect world:
>    I could just drop my own layouts into the designated location for them, even if I use the default config XML, and choose between my custom layouts and the default ones
>    If name collisions are an issue, user-supplied layouts take precedence over default ones
>    I could customize navit.xml and have an easy way to get the same behavior (e.g. by copying the include stuff from the original XML file)
>
> Would these things work?

I am pretty sure there can be a technical solution for this.
My first guess for such a solution would be:
By having the default (shipped) navit.xml include first `$NAVIT_DEFAULTCONFDIR/navit_layout_*.xml` then `$NAVIT_USERCONFDIR/navit_layout_*.xml` (and (2) getting navit set the proper directories values in `$NAVIT_DEFAULTCONFDIR` and `$NAVIT_USERCONFDIR`).
Let's then say that (3) when the XML is parsed, layouts are taken in the order they come, and a layout with an already existing name overrides a previous definition. I think it should solve the first and second requirement (user-supplied layouts coming after default system ones, they would override them).
Copying the navit.xml file to the user-specific location would still use the default shipped layouts, but copying one or several layout files as well to the user-specific location (and modifying them) would allow to override them.
One could even think of (4) a mechanism to make a layout disappear from the list by adding an empty (but with the same name) <layout></layout> section in the user-specific location.

Does this proposal meet your needs?

If so, then it's a matter of implementing (in the C-code) items (2), (3) and (4) and I think this new behaviour can still be backwards compatible with users that want to keep a monolithic navit.xml as it exists today.

lains:
In `.circleci/config.yml`, I saw the following code:

```
      - run:
          name: Move xml's
          command: |
            cd navit
            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitldpi.xml OSD_SIZE=1 ICON_SMALL=24 ICON_MEDIUM=32 ICON_BIG=48
            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitmdpi.xml OSD_SIZE=1.33 ICON_SMALL=32 ICON_MEDIUM=48 ICON_BIG=64
            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navithdpi.xml OSD_SIZE=2 ICON_SMALL=48 ICON_MEDIUM=64 ICON_BIG=96
            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitxhdpi.xml OSD_SIZE=2.67 ICON_SMALL=64 ICON_MEDIUM=96 ICON_BIG=128
            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitxxhdpi.xml OSD_SIZE=4 ICON_SMALL=96 ICON_MEDIUM=128 ICON_BIG=192
            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitxxxhdpi.xml OSD_SIZE=5.33 ICON_SMALL=128 ICON_MEDIUM=192 ICON_BIG=256
            sudo ls -la
            mv *dpi.xml android/res/raw
            cd ../

```

Does this code take precedence to the xslt processing that is done by cmake?
I guess I need to update this code as well to take new navit*.xml files and their new location for AssetManager.

jandegr:
Hi,
The code in circleci/config.yml replaces the one in cmakelists when building with Gradle / Ninja generator, as CI now does.

regards,

mvglasow:
@lains Re coexistence of customized vs. built-in layouts and XML vs. layout customization—that sounds good to me.

lains:
@mvglasow, I have made the changes for:
* (1) - commit fb06d0244406dee7f6603d404ddacf68b05e4c78
* (3) - commits 4e4396cae26b9e79401f601260fdf9a2f61c7288 and 1f54165021d81015da13a800a268d61f86c70a66... actually, it was much easier to implement it the other way around: when two layouts have identical names the first one is used, the following occurrences are discarded.

(2) did not need any change because there were already variables `NAVIT_SHAREDIR` and `NAVIT_USER_DATADIR` for this.
(4) could be implemented using an additional enable/disable attribute to the layout tag.

lains:
OK, so I think most of the code is there.
This should allow backwards compatibility with existing navit.xml, also allow user-defined layouts to override default ones, and layout names are not hardcoded in the Java code anymore (but they need to be listed inside `navit/CMakeLists.txt` and `.circleci/config.yml`.

I have also changed the way config .xml files are stored inside the Android package, from res/raw to assets. This allow the android app to browse the list of included files.
For this purpose, a new method `extractAsset()` has been created. Maybe translation files (*.mo) should also be moved from res/raw to assets since they are the only remaning files that still reside in res/raw. If they are moved as well, than the method `extractRes()` becomes unused and can be removed.

Please feel free to test, especially with custom layout setups, and to give me any feedback on whether this feature can be useful to the navit project.

lains:
Update PR to match new indentation guidelines.
This code is not ready for merge though, there are a few changes that is still need to do related one ant:checkstyle error. I have also switched to building Android packages using gradle instead of cmake, so I will need to retest on Android.

lains:
Build tested successfully, after gradle build, on Android.
CodeFactor is seeing duplicate code between assets/ and res/raw extractions. This would be fixed by moving all translations also to assets as explained [above](https://github.com/navit-gps/navit/pull/559#issuecomment-388484666).
Not sure why sanity_check is failing... indentation should have been lined-up with the current trunk.

pgrandin:
@lains regarding the sanity checks, here's what I see : if you look at the full log output ( https://circleci.com/api/v1.1/project/github/navit-gps/navit/9051/output/103/0?file=true ) towards the end there's some diffs between the code you pushed and the one that has been re-linted.

in gradlew.bat we have some lines terminated with `^M` :
```
diff --git a/gradlew.bat b/gradlew.bat
index aec9973..8a0b282 100644
--- a/gradlew.bat
+++ b/gradlew.bat
@@ -1,90 +1,90 @@
-@if "%DEBUG%" == "" @echo off^M
-@rem ##########################################################################^M
-@rem^M
-@rem  Gradle startup script for Windows^M
-@rem^M
-@rem ##########################################################################^M
-^M
[...]
```

I'm not sure why this one differs:
```
diff --git a/navit/support/espeak/CMakeLists.txt b/navit/support/espeak/CMakeLists.txt
index 55f461a..efada42 100644
--- a/navit/support/espeak/CMakeLists.txt
+++ b/navit/support/espeak/CMakeLists.txt
@@ -2,8 +2,8 @@
 if(INTERNAL_ESPEAK_COMPLETE)
        set(ESPEAK_LIBRARY_ADDITIONAL speak_lib.c)
 endif()
-supportlib_add_library(support_espeak compiledict.c dictionary.c intonation.c readclause.c setlengths.c
-       numbers.c synth_mbrola.c synthdata.c synthesize.c translate.c tr_languages.c voices.c wavegen.c
+supportlib_add_library(support_espeak compiledict.c dictionary.c intonation.c readclause.c setlengths.c
+       numbers.c synth_mbrola.c synthdata.c synthesize.c translate.c tr_languages.c voices.c wavegen.c
        phonemelist.c klatt.c speak_init.c ${ESPEAK_LIBRARY_ADDITIONAL})

 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/espeak-data DESTINATION ${SHARE_DIR} PATTERN ".svn" EXCLUDE)
```

Probably a trailing space here?
```
diff --git a/navit/support/glib/CMakeLists.txt b/navit/support/glib/CMakeLists.txt
index 56d1ae8..d53a0a8 100644
--- a/navit/support/glib/CMakeLists.txt
+++ b/navit/support/glib/CMakeLists.txt
@@ -1,3 +1,3 @@
-supportlib_add_library(support_glib fake.c galiasdef.c gatomic.c gerror.c ghash.c
+supportlib_add_library(support_glib fake.c galiasdef.c gatomic.c gerror.c ghash.c
        glib_init.c glist.c gmem.c gmessages.c gprimes.c gprintf.c gslice.c gslist.c
-       gstrfuncs.c gstring.c gutf8.c gutils.c)
+       gstrfuncs.c gstring.c gutf8.c gutils.c)
```

Probably a leading space here:
```
diff --git a/navit/support/wordexp/glob.c b/navit/support/wordexp/glob.c
index fabceb6..f971fd2 100644
--- a/navit/support/wordexp/glob.c
+++ b/navit/support/wordexp/glob.c
@@ -159,7 +159,7 @@ glob_recursive(const char *path1, const char *path2, const char *pattern, int fl
                                }
                        }
                        closedir(dh);
-               }
+               }
        } else {
                glob_recursive(path, fname, next, flags, errfunc, pglob);
        }
```

Probably a trailing space here:
```
diff --git a/navit/support/xgetopt/XGetopt.c b/navit/support/xgetopt/XGetopt.c
index 60bc0f4..255fdca 100644
--- a/navit/support/xgetopt/XGetopt.c
+++ b/navit/support/xgetopt/XGetopt.c
@@ -11,7 +11,7 @@
 //     - Added Unicode support
 //
 //     Version 1.1 - 2002 March 10
-//     - Added example to XGetopt.cpp module header
+//     - Added example to XGetopt.cpp module header
 //
 // This software is released into the public domain.
 // You are free to use it in any way you like.
```

lains:
@pgrandin 
OK I just gave a look.
Issue with `navit/support/espeak/CMakeLists.txt` is actually also because of trailing spaces, as for almost all other failing checks.
Trailing CR seems normal to me on a Windows-dedicated .bat file
All these fails on sanity_check are on files that are not changed, so I guess they should be fixed in a separate PR just not to mix them with the actual intent of this PR.

lains:
This PR should also modify `ci_sanity_checks.sh` to run xmllint on XML layout files in addition to  `navit_shipped.xml`. See PR #609.

lains:
@aerostitch, I am not really sure about how to rework `ci_sanity_check.sh` for the new XML layout files (how to make xmllint undestrand `xi:include` tags, which, I guess, is preferred over having a new partial dtd for include layout XML files)

For the rest of this PR, test was successful on Android build using gradle.

There could probably be a GUI tool to copy shipped files to the user local directory, so that the user can then edit these files (indeed, it is not trivial at all to do this on Android... it is actually listed in https://wiki.navit-project.org/index.php/Usability#Android_port_specific_issues)

aerostitch:
@lains I think we can update the sanity check in a separate PR and focus on your modifications for this PR to avoid confusion.

lains:
OK, so I keep this PR as is, I think it does the job.
The rest will come in future PRs.
Is it OK for merge?

lains:
Builds correctly also on WinCE.
Still needs to be tested on WinCE.

jkoan:
Tested on WinCE its working nicely :)

img_20180715_155407

lains:
This is great news, thanks for the test!

.circleci/config.yml
```diff
@@ -88,13 +88,20 @@ jobs:
...
           command: |
             cd navit
-            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitldpi.xml OSD_SIZE=1 ICON_SMALL=24 ICON_MEDIUM=32 ICON_BIG=48
-            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitmdpi.xml OSD_SIZE=1.33 ICON_SMALL=32 ICON_MEDIUM=48 ICON_BIG=64
-            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navithdpi.xml OSD_SIZE=2 ICON_SMALL=48 ICON_MEDIUM=64 ICON_BIG=96
-            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitxhdpi.xml OSD_SIZE=2.67 ICON_SMALL=64 ICON_MEDIUM=96 ICON_BIG=128
-            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitxxhdpi.xml OSD_SIZE=4 ICON_SMALL=96 ICON_MEDIUM=128 ICON_BIG=192
-            saxonb-xslt -s:navit_shipped.xml -xsl:xslt/android.xslt -o:navitxxxhdpi.xml OSD_SIZE=5.33 ICON_SMALL=128 ICON_MEDIUM=192 ICON_BIG=256
-            mv *dpi.xml android/res/raw
+            export XMLCFG_ALL_FILE_PREFIXES="navit navit_layout_car navit_layout_car_dark navit_layout_car_android navit_layout_car_simple navit_layout_bike navit_layout_th"
```
lains: This should be generic without having to list all files, like what has been done in cmake.

lains:
I'm switching this PR back to WIP because I have hardcoded the list of xml files in circleci ([again!](https://github.com/navit-gps/navit/pull/559#issuecomment-385463186)) and this should be improved

mdankov:
Unfortunately, it didn't work for me. I see no layouts besides the "route" layout which is shipped in navit.xml

```
I have a few errors in navit.log:

error:navit:xinclude:Unable to include '\SDMMC\MobileNavigator/$NAVIT_USER_DATADIR/navit_layout_*.xml'
error:support_libc:_stat:no file
error:navit:xinclude:Unable to include '\SDMMC\MobileNavigator/\SDMMC\MobileNavigator/navit_layout_car_dark.xml'
error:support_libc:_stat:no file
error:navit:xinclude:Unable to include '\SDMMC\MobileNavigator/\SDMMC\MobileNavigator/navit_layout_car_simple.xml'
error:support_libc:_stat:no file
error:navit:xinclude:Unable to include '\SDMMC\MobileNavigator/\SDMMC\MobileNavigator/navit_layout_car_android.xml'
error:support_libc:_stat:no file
error:navit:xinclude:Unable to include '\SDMMC\MobileNavigator/\SDMMC\MobileNavigator/navit_layout_car.xml'
error:support_libc:_stat:no file
error:navit:xinclude:Unable to include '\SDMMC\MobileNavigator/\SDMMC\MobileNavigator/navit_layout_bike.xml'
error:support_libc:_stat:no file
error:navit:xinclude:Unable to include '\SDMMC\MobileNavigator/\SDMMC\MobileNavigator/navit_layout_th.xml'
```
Looks like a xi:include parser problem, it attempts to expand $NAVIT_SHAREDIR twice, or maybe is always prepending navit.exe's folder name. IIRC, there's no 'current working folder' semantics on WinCE, so we may have a workaround which conflicts to xi:include parser.

It's quite surprising @jkoan has reported it to be working well. Maybe he's using a more standard navit base folder name? On my device, I have to put navit into \SDMMC\MobileNavigator\ folder.

Thank you!

navit/xmlconfig.c
```diff
@@ -750,17 +762,31 @@ static void xinclude(xml_context *context, const gchar **attribute_names, const
...
         dbg(lvl_debug,"%d results", count);
-        if (file_exists(we_files[0])) {
-            for (i = 0 ; i < count ; i++) {
-                dbg(lvl_debug,"result[%d]='%s'", i, we_files[i]);
-                doc_new.href=we_files[i];
-                parse_file(&doc_new, error);
+        for (i = 0 ; i < count ; i++) {
+            included_filename = g_strdup(we_files[i]);
+            if (*included_filename != '\0') { /* Non empty href */
+                if (*included_filename != '/') {	/* The filename's path is relative */
```
mdankov: On Windows, absolute paths may start with /, \ or drive letter name followed by a colon (like c: ).

Actually, It's a bit harder on Windows, as it keeps current working directory for each drive, but I doubt we should care about it here.

mdankov: @lains g_path_is_absolute should solve the problem.

jkoan:
Hi @mdankov,
my navit installation just goes into the /flash storage/navit directory, I got a script which runs on boot where I can configure what Programm to run.
I can give further informations about my deceive and setup tomorrow. But to be short I just unzipped all contents of the artifact zip file from circleci to the flash and it worked perfectly. I also removed some layouts and rebooted to make sure everything goes as expected.

navit/xmlconfig.c
lains: I am going to work on this. Thanks.

lains:
I have checked for navit_layout_* hardcoded names, and this is actually also a hardcoded list in navit/CMakeLists.txt.
Maybe this is actually not a bad thing to keep control of exactly which navit_layout_*.xml file is shipped in the package, following the concept of Makefile where you often explicitly have to list files...

mdankov:
@lains looks like we have the that function disabled in our own glib copy as we didn't use it yet.

https://github.com/navit-gps/navit/blob/dc8f7f1b182450ed017d0262872820cc26f08dc9/navit/support/glib/gutils.c#L759

Just move that #if  to line 787.

lains:
@mdankov, I have switched to `g_path_is_absolute` as you suggested, can you test that this fixes the issue you found on WinCE?
If there are still issues, can you get more logs by adding the following in navit.xml:
`<debug name="navit:xinclude" dbg_level="debug"/>`

The mix of Windows/UNIX directory separators that you have in your logs worries me however, but we shouldn't see them anymore now that xi:include href absolute path are fixed (and absolute path are what is in this PR's default navit_shipped.xml).
Would you mind also testing this code using relative links in navit_shipped.xml please? (this is to make sure relative links also work on WinCE). I have switched to `g_path_get_dirname()` in my code, so that may have fixed the potential issue altogether. 
In order to test this, you would have to change the xi:include in navit_shipped.xml with the following:
`<xi:include href="navit_layout_*.xml"/>
`

mdankov:
The mix of Windows/UNIX directory separators is not a problem. Windows recognizes both / and \ as separators and allows to mix them in a single path.

Unfortunately, my WinCE device hardware has failed just now. Looks like its screen backlight lamp is dead. So I can't do any tests with it right now, sorry. Will try to fix it later.

lains:
I have been able to test the code from this PR on a WinCE 4.2 device, It seems the expansion of wildcards and test for relative/absolute paths work so the issue with WinCE is now fixed:

```
debug:navit:xinclude:At level 0, processing xi:include in document href="\Storage Card\NAVIT/navit.xml"
debug:navit:xinclude:expanding '$NAVIT_USER_DATADIR/navit_layout_*.xml'
debug:navit:xinclude:1 results
debug:navit:xinclude:converted relative filename='$NAVIT_USER_DATADIR/navit_layout_*.xml' to absolute filename='\Storage Card\NAVIT\$NAVIT_USER_DATADIR/navit_layout_*.xml'
debug:navit:xinclude:result[0]='\Storage Card\NAVIT\$NAVIT_USER_DATADIR/navit_layout_*.xml'
error:support_libc:_stat:no file
error:navit:xinclude:Unable to include '\Storage Card\NAVIT\$NAVIT_USER_DATADIR/navit_layout_*.xml'
debug:navit:xinclude:At level 0, processing xi:include in document href="\Storage Card\NAVIT/navit.xml"
debug:navit:xinclude:expanding '$NAVIT_SHAREDIR/navit_layout_*.xml'
debug:navit:xinclude:1 results
debug:navit:xinclude:result[0]='\Storage Card\NAVIT/navit_layout_car.xml'
error:navit:layer_set_attr_do:Ignoring reference to unknown layer 'Found items' in layer 'Found items'.
```

@mdankov, as you said, the mix of both Windows and UNIX-style directory separators is handled properly.
Also, one can see that on my WinCE device, as on yours, NAVIT_USER_DATADIR variable is probably undefined, thus not replaced (first wildcard expansion), but NAVIT_SHAREDIR is expanded as expected.

mvglasow:
While we’re at it, how about scaling maps for different DPI displays automatically? Currently things look pretty poor on a higher-DPI Android display. I hacked together a script to scale Navit layouts (https://github.com/mvglasow/navit-goodies/tree/master/scripts/mapscaler), maybe you can use some of it.

lains:
Tested after merge and working as expected on Android.
It also worked on my WinCE device. @mdankov, did you manage to get your device working again? If so, can you retest please?

lains:
Note: the failure on sanity_check is due to:

> scripts/ci_sanity_checks.sh: line 57: xmllint: command not found

Can someone check why xmllint is not working anymore on the CI machine image for android?

lains:
@hoehnp, I don't think anyone has tested this but I am interested in the result. Did you manage to do it?

lains:
@hoehnp, did you have any success on testing this feature on Sailfish OS?

hoehnp:
@lains, i am not aware of any built package of Sailfish OS. As soon as i have a modified one, i can give it a shot.

metalstrolch:
OK. Building Sailfish the usual way fails with:
```
[ 19%] Generating navit.xml
Applying sailfish_disable.xslt
Applying sailfish_qt5.xslt
Applying sailfish_osd.xslt
Applying sailfish_cursor.xslt
Applying sailfish_mapset.xslt
Applying sailfish_svg.xslt
Applying sailfish_gui.xslt
Error renaming from "/tmp/_home_src1_rpmbuild_BUILD_navit-build_navit__navit.xml/navit_shipped.xml.temp" to "/home/src1/rpmbuild/BUILD/navit-build/navit//navit.xml": Invalid cross-device link
make[2]: *** [navit/navit.xml] Error 1
make[1]: *** [navit/CMakeFiles/navit_config_xml.dir/all] Error 2
make: *** [all] Error 2
```
Don't think this is that Sailfish specific. I think this is because /tmp is on a differrent partition and rename refuses to work then. Cross checking this on plain Linux.

navit/CMakeLists.txt
```diff
@@ -124,48 +124,93 @@ add_custom_command(
 	)
 
 # additional parameter are passed to the stylesheet processor as parameter
-macro(process_xslt SRC_XML DEST_XML)
-	set(XSLT_COMMANDS COMMAND ${CMAKE_COMMAND} -E copy "${PROJECT_SOURCE_DIR}/navit/navit_shipped.xml" "${CMAKE_CURRENT_BINARY_DIR}/${DEST_XML}" )
+macro(process_xslt SRC_DIR SRC_XML DEST_DIR DEST_XML)
+	if(WIN32)
+		set(TMPDIR "$ENV{TEMP}")
+	else()
+		set(TMPDIR "/tmp")
```
metalstrolch: Suggest changing this to
`set(TMPDIR "${CMAKE_CURRENT_BINARY_DIR}")`
to put intermediate files into CMAKEs build directory. Putting it hard into /tmp is preventing parallel build runs, and somehow uncommon. It should all be placed in CMAKEs build dir shouldn't it?
And this prevents the rename from failing if /tmp is on a different filesystem than the build directory.

hoehnp:
@metalstrolch : should we maybe add building on sailfish using the container by coderus to the common build steps? If you don't mind, I would try this.

metalstrolch:
Good idea if you manage to get this working. I'm no docker expert, so I never tried.
This would allow us to produce sailfish nightly and release packages the usual way as well.

navit/CMakeLists.txt
lains: Yes, sure, let me try that.
I am also considering changing the gradle build scripts to use the cmake target for xslt processing (rather that duplicating the code)

lains:
I will do some experiments on this (rework the xslt). This should:

- avoid using harcoded /tmp for intermediate xslt files)
- allow to remove the xslt processing code from the `.circleci/config.yml` and just issue a `make navit_config_xml` instead... this is not perfect though, as I guess this should all be integrated in `build.gradle` but I'm not very good at gradle config files...

lains:
There are indentation issues that a re not strictly related to this code.
I'd rather submit fixes for those in a preliminary PR rather than adding unrelated changes to this PR.
Is there a way to allow build check in circleci even if sanity check fails?

lains:
Good point on default layout, probably this can be fixed by adding an attribute on layouts with a priority of some kind. Default layout (when no selection has been done by the user) would be the highest value os something like this...
About Sailfish, did you refer to PR #679? Is there anything change to be done here?

pgrandin:
This default layout change is annoying as it kind of prevents merging this PR. We could maybe rename the `car` profile to `auto` or some other name that keeps it first, and implement a proper default layout in a second step. Thoughts?

jkoan:
@pgrandin i totaly agree with you. And after that we should finaly merge :)

mvglasow:
Is there currently no option to set a default layout by name? That would be the cleanest approach IMHO, anything else is a workaround at best.

lains:
> Another option would be to store the default layout in the parent (i.e. navit): add a layout or default_layout attribute to it which stores the name of the default layout. (If the attribute is missing or refers to a nonexistent layout, fall back to the first layout in the list).

I will look into this... it could be quite easy.
I just need to identify the code that selects the default layout.

lains:
Default layout is set by [function navit_add_layout()](https://github.com/navit-gps/navit/blob/trunk/navit/navit.c#L2812). Initially, `layout_current` is NULL, thus the first layout element that is parsed and added to navit's list of layouts will be selected as `layout_current`.
This is where a more complex logic must be implemented.

mvglasow:
Either there or after the function returns: Read the to-be-defined attribute, and if it is non-NULL, cycle through the layouts, and if a match is found, set it as the current layout. If no match is found, nothing is touched, so the first layout is the default one.

pgrandin:
> Another option would be to store the default layout in the parent (i.e. navit): add a layout or default_layout attribute to it which stores the name of the default layout. (If the attribute is missing or refers to a nonexistent layout, fall back to the first layout in the list).

exactly how i would have done it.

pgrandin:
Unless someone objects, I think that we can still merge this PR as-is and work on the default layout mechanism that has been discussed in another PR.

mvglasow:
My layout is called Android-XXHDPI and thus appears before auto, hence no objection from my part :-) Unless the default layout mechanism turns out to be a real quick addition…

lains:
OK, if you give me another 1 or 2 days, I will try to add the default layout mechanism.

lains:
Almost done...

lains:
I have renamed the `navit_layout_auto.xml` file into `navit_layout_car.xml` as initially (and updated the sailfish spec file accordingly).
There is a new `default_layout` attribute to the <navit> tag in the config file. If a layout name is provided there, and it is valid, then this is the default layout that will be used (instead of the first file found during wildcard expansion of navit_layout_*.xml). Otherwise, the first file will be used, which is then going to be "bike".

lains:
The term "layout" should maybe be renamed "map_theme" (in the code and in the confid) so that it is aligned with the user interface and probably is a bit more self explainatory.
But this is cosmetic change and impacts quite a lot of files, so it is another PR.

lains:
All changes related to default config selection have been implemented.

mvglasow:
> The term "layout" should maybe be renamed "map_theme" (in the code and in the confid) so that it is aligned with the user interface and probably is a bit more self explainatory.

As far as I can tell, though, the internal UI currently has “layout” for the map—unless that has changed since I merged the last upstream config. The UI probably needs to be verified as well and updated where needed.

lains:
> As far as I can tell, though, the internal UI currently has “layout” for the map—unless that has changed since I merged the last upstream config. The UI probably needs to be verified as well and updated where needed

You're right, it's only in the french translation that "layout" has been translated as "theme".

lains:
OK to merge? Terminology change for <layout> tags should be another PR anyway...

hoehnp:
@lains: Can you give me a few days. I would put it on my todo list to download the RPM for sailfish and test if the build starts, runs, etc.

navit/CMakeLists.txt
lains: Build for android now uses cmake for xslt processing (was implemented https://github.com/navit-gps/navit/pull/559/commits/4e1a6f2314380c0fc5c7d52634de69bead6ee490)

navit/xmlconfig.c
lains: This was submitted in:
https://github.com/navit-gps/navit/pull/559/commits/d09cb0cdb54f301afbfffa2dc054032867e8aab8
https://github.com/navit-gps/navit/pull/559/commits/093f5cbbf90e11eb753bc5beb7bb54253c132e85

hoehnp:
@lains: Just after installing it starts, however, I only see a full navit.xml in the rpm file. Can you provide me a sample file?

hoehnp:
@lains: I corrected the issue. It was a problem caused by me in the circleci config file. I can confirm, that navit starts on Sailfish. Maybe tomorrow, I can download a map and test it on a short distance.

hoehnp:
@lains: I did further tests. The short distance navigation I did not do yet because of #714. Do you think both things are related?

pgrandin:
@hoehnp this PR is quite massive and aside from the issue you mentioned seems ready to me to be merged. We should probably track your issue separately to avoid delaying this PR even more.

lains:
Wiki pages updated to explain how to set default layout and how layout are split out of navit.xml

PR #560, opened 2018-04-30
Fix:debug:Correction on segfault in debug_vprintf() introduced in 221f783e

lains:
When using the dbg(log_error, ...) function, I now get segfaults.
The bug has been introduced at commit 221f783ea1caaaab2f5ceadc6b0fb3e720aac3df
g_strconcat's arguments should be a list of pointers, the last has to be NULL.
I have reworked the append so that it does not require a temporary allocation of the format string, but rather adds EOL characters manually.
I changed the hardcoded 4095 to a sizeof() call which is more future-proof in case the buffer size is updated in the future.
I finally, I also force \0 termination of the resulting string (maybe there could be issues if the message is longer than the 4095 bytes of the buffer). Please review this last point as it is maybe not really needed.

### DIFF
--- a/navit/debug.c
+++ b/navit/debug.c
@@ -308,14 +308,14 @@ debug_vprintf(dbg_level level, const char *module, const int mlen, const char *f
 #if defined HAVE_API_WIN32_CE
 #define vsnprintf _vsnprintf
 #endif
-		gchar *fmt_newline;
+		vsnprintf(debug_message+strlen(debug_message),sizeof(debug_message)-1-strlen(debug_message),fmt,ap);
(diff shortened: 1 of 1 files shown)


lains:
Actually, there is still an issue with my code... strings might not be NUL terminated when I overwrite with either \r or \n let me try to refactor a bit more and add doxygen comments. I think I can even use only one 4096 buffer instead of two... I will propose a subsequent PR (or shall I commit in the same PR?)

jkoan:
Just commit in the same PR, this is fine

jkoan:
As the PR is already merged you need to open a new PR but you can just use the same branche

PR #564, opened 2018-05-01
Fix:debug:Fix potential unterminated C--string and refactoring debug_vprintf()

lains:
Following PR #560, when inserting '\r' or '\n', the string may not be NUL terminated.
This fixes this and also reduces the string buffer to only 4096.
I also added doxygen comments and dbg_assert() to protect from unlikely buffer overflow. Buffer overflow on the message itself is directly checked at run time (it is probably the only realistically possible occurrence of buffer overflow).
This should be checked on Windows, especially with timestamps, as I have no environment to test.


navit/debug.c
```diff
@@ -281,51 +292,103 @@ dbg_level_to_android(dbg_level level)
...
 				debug_timestamp(debug_message);
-			strcpy(debug_message+strlen(debug_message),dbg_level_to_string(level));
-			strcpy(debug_message+strlen(debug_message),":");
-			strcpy(debug_message+strlen(debug_message),message_origin);
-			strcpy(debug_message+strlen(debug_message),":");
+				len = strlen(debug_message);
+				end = debug_message+len;
+			}
+			dbg_assert(end+strlen(dbg_level_to_string(level))+1 < debug_message+sizeof(debug_message)); /* Make sure we don't get any overflow */
+			strcpy(end,dbg_level_to_string(level));
```
aerostitch: Could you use the glib function here instead of `strcpy` please?

lains: Thanks for pointing this out. Yes, it can be used instead, but this implies allocation and then freeing the string.
Given that debug output is used quite extensively on debug builds, shouldn't alloc/dealloc be avoided as much as possible?
With strcpy() above, this allows direct write to the only one 4096-byte buffer.
There shouldn't be any buffer overflow because of the dbg_assert() above, but this is unlikely anyway also because we are alsmot at the beginning of the 4096-byte buffer and we are only copying a few bytes (the debug level as a string).
What is you opinion?

aerostitch: Good point. If we want to avoid allocations, maybe `g_strlcpy` would be an interesting alternative and would avoid the `dbg_assert` (it would have an automatic truncation). What do you think?

lains: Yes! Very good idea. I didn't know about g_strlcpy(), I am going to review my proposal accordingly.

aerostitch:
The changes look good to me but I'd be more comfortable if a 2nd pair of eyes could double-check.

charlescurley:
On Thu, 03 May 2018 13:55:36 -0700
Joseph Herlant <notifications@github.com> wrote:

> The changes look good to me but I'd be more comfortable if a 2nd pair
> of eyes could double-check.
> 

I went to look, but I don't seem able to check out the branch
fix-dbg-segfault. It is not listed as a remote or local branch.

-- 
"When we talk of civilization, we are too apt to limit the meaning of
the word to its mere embellishments, such as arts and sciences; but
the true distinction between it and barbarism is, that the one
presents a state of society under the protection of just and
well-administered law, and the other is left to the chance government
of brute force."
- The Rev. James White, Eighteen Christian Centuries, 1889
http://www.charlescurley.com

aerostitch:
@charlescurley it's because the branch comes from a fork.
But you can use https://help.github.com/articles/checking-out-pull-requests-locally/ to check it out tho.
That would be:
```
git fetch origin pull/564/head:PR_564
git checkout PR_564
```

charlescurley:
On Thu, 03 May 2018 15:49:10 -0700
Joseph Herlant <notifications@github.com> wrote:

> @charlescurley it's because the branch comes from a fork.
> But you can use
> https://help.github.com/articles/checking-out-pull-requests-locally/
> to check it out tho. That would be: ```
> git fetch origin pull/564/head:PR_564
> git checkout PR_564

Much better, thanks.

A few thoughts...

What an unholy mess of code! Even after lains' work. I probably didn't
get everything.

Library functions have return values for a reason. Either use them or
cast them to void to make it clear you are deliberately not using them.

More specific...



In debug_timestamp, is it necessary for val to be a double float?
Making it an int of some sort will save you all sorts of floating point
maths. E.g.:

val=counter.HighPart * 4294967296.0 + counter.LowPart;

becomes

val=counter.HighPart<<32 + counter.LowPart;

The stickler is whether you can get the precision you need where you
divide val by the frequency.

in debug_vprintf:

The documentation for g-strlcpy says that the return value is the size
of the *attempted* copy, i.e. including any overflows. So you could use
that to check for overflows before going any further.
https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strlcpy


Without knowing the code, I am dubious of the value of having a call to
dbg_assert in debug code.


Line 353:
			*end++ = ':';
			len=strlen(message_origin);

NO!!! Some (most? all?) implementations of strlen rely on the string
being null terminated. The first line there steps on what should be a
null terminator. But it does nothing to append a new null terminator.

			*end++ = ':';
                        *end = '\0';
			len=strlen(message_origin);

(I think this is what lains tried to do in commit
831d89e0e6a09f59f1081a724be0e98ee848d0e7. If so, he missed this one,
which is easy enough to do in this code.)

This is done correctly down at line 359 and elsewhere.

And, rather than calling strlen there, how about simply

                        len++;



-- 
"When we talk of civilization, we are too apt to limit the meaning of
the word to its mere embellishments, such as arts and sciences; but
the true distinction between it and barbarism is, that the one
presents a state of society under the protection of just and
well-administered law, and the other is left to the chance government
of brute force."
- The Rev. James White, Eighteen Christian Centuries, 1889
http://www.charlescurley.com

charlescurley:
On Sat, 05 May 2018 13:20:15 +0000 (UTC)
lains <notifications@github.com> wrote:

> > In debug_timestamp, is it necessary for val to be a double float?
> > Making it an int of some sort will save you all sorts of floating
> > point maths. E.g.:
> > 
> > val=counter.HighPart * 4294967296.0 + counter.LowPart;
> > 
> > becomes
> > 
> > val=counter.HighPart<<32 + counter.LowPart;  
> 
> I'm not confident in changing this code because I have no way to test
> under Windows (I actually did a few changes on the Windows-specific
> code and I don't like to do blind coding, so I'd rather have someone
> that has a way to compile the Windows version of the code to do the
> change).

OK, then leave it alone. If it works, don't fix it.

Which leads to the broader question, do we still have WinCE users and
maintainers?

> 
> > in debug_vprintf:
> > 
> > The documentation for g-strlcpy says that the return value is the
> > size of the *attempted* copy, i.e. including any overflows. So you
> > could use that to check for overflows before going any further.
> > https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strlcpy  
> 
> Yes, this can be done, sure. It is low risk in my opinion because we
> are copying a few characters almost at the beginning of a 4096
> byte-buffer

OK. Your call, then.

> 
> > 
> > Line 353:
> > 			*end++ = ':';
> > 			len=strlen(message_origin);
> > 
> > NO!!! Some (most? all?) implementations of strlen rely on the string
> > being null terminated. The first line there steps on what should be
> > a null terminator. But it does nothing to append a new null
> > terminator.  
> 
> Yes, the string contructed at `*end` is actually not terminated at
> line 353 but at line 354 `strlen()` is not applied to this
> unterminated string but to `message_origin` which is \0 terminated
> (at line 315)
>
> ` *message_origin = '\0';	/* Force string termination of message_origin (last byte of debug_message) */
> `
> This is far from obvious though, I agree!

OK. I'll take your word for it. Perhaps this is where efficiency should
give way to maintainability. I did not catch the fact that the code is
building two strings up in one buffer. *Shudder!*

<rant>Another example of the utter failure of "The code is the
documentation".</rant>

>
> What is not \0 terminated at line 353 is the C-string pointed by
> `debug_message` (at the end of which `end` points at that
> moment). But line 356 continues this string a few lines later by
> using `memmove()` on `end` Line 359 adds a ':' and finally line 360
> terminates the string with '\0' as you noticed!

OK, fair enough. It's a possible trap for a later maintainer, but let
that go.

> 
> > And, rather than calling strlen there, how about simply
> > 
> >                         len++;  
> 
> Actually, `strlen()` gets the length of `message_origin` while using
> `len++` would just take into account the fact we added 1 character to
> `debug_message` so it is a different string.
> 
> I agree this code is hard to read, mostly because it stores two
> strings in one char[] buffer, and just to save a bit of space (4kB).

I have no problem saving 4 kB, having worked in environments where 4
kB would have been utter luxury. But, even under WinCE, are we under
that much of a constraint? Would 2 2 kB buffers do the job just as
well, and be easier to understand. Or is that another WinCE ism of
which I have remained cheerfully ignorant?

I'm not sure what the Navit project customs are, but I would document
this in the code.

-- 
"When we talk of civilization, we are too apt to limit the meaning of
the word to its mere embellishments, such as arts and sciences; but
the true distinction between it and barbarism is, that the one
presents a state of society under the protection of just and
well-administered law, and the other is left to the chance government
of brute force."
- The Rev. James White, Eighteen Christian Centuries, 1889
http://www.charlescurley.com

lains:
> `
> val=counter.HighPart * 4294967296.0 + counter.LowPart;
> `
> becomes
> `
> val=counter.HighPart<<32 + counter.LowPart;
> `
>
> I'm not confident in changing this code because I have no way to test
 under Windows (I actually did a few changes on the Windows-specific
 code and I don't like to do blind coding, so I'd rather have someone
 that has a way to compile the Windows version of the code to do the
 change).

I think it would be a good thing to apply your proposal if it still behaves the intended way.
Is there anybody that can try the Windows build and check that this code is OK (by forcing timestamps on)?

> I have no problem saving 4 kB, having worked in environments where 4
kB would have been utter luxury. But, even under WinCE, are we under
that much of a constraint? Would 2 2 kB buffers do the job just as
well, and be easier to understand. Or is that another WinCE ism of
which I have remained cheerfully ignorant?

Good point, shall we really optimize for size and speed when code becomes harded to understand?
This is really the compromise.
To be honest, I just followed the track of the previous code (or at least what I believed it was): optimize for size and speed... and this means avoid strlen, allocs on the heap, and multiplication of buffers.
I added `dbg_assert()` clauses because it is obvious that, indeed, there are potential memory leaks. However, in my opinion, they could be caused more by a rework on the source code for `debug_vprintf()`, rather than by misusing calls to `dbg()`.
The only big risk for overflow when calling `dbg()` is to provide a log message that is too long for the buffer, and this is handled by using `vsnprintf()`
I really need review of this code by other people just to make sure I didn't mess up.
@charlescurley, did you have any chance to compile and to run step-by-step in a debugger?

> I'm not sure what the Navit project customs are, but I would document
this in the code.

Yes, your replies made me understand it is quite unreadable, so maybe this can be (at least patly fixed by documentation in the code).
Let's see what I can do by adding some verbosity to the code...

pgrandin:
> Which leads to the broader question, do we still have WinCE users and maintainers?

We still have some users yes. 

> I'm not sure what the Navit project customs are, but I would document this in the code.

Absolutely. 

> To be honest, I just followed the track of the previous code (or at least what I believed it was): optimize for size and speed...

Some of this code is over 10 years old, so this kind of optimizations might not be as relevant anymore?

Talking about optimizations, buffer overflows and things like that, we are also using coverity static code analysis. If you guys are interested, I can send you an invite.

It gives us this kind of reports :

coverity

(this is from https://github.com/navit-gps/navit/blob/221f783ea1caaaab2f5ceadc6b0fb3e720aac3df )/navit/vehicle/file/vehicle_file.c#L600 )

lains:
@pgrandin, no problem to participate to the effort of code analysis if I can find enough spare time. Do you have a report that covers the whole code?
I agree with @charlescurley on the fact that once all warnings are covered or marked as irrelevant, any new issue will likely get more review and care, so the big part of the work is the initial analysis.

pgrandin:
@charlescurley 
> OK, then let us continue to support them. But I am not in any position to do so.

No problem. This was intended as a "let's not drop support for it for now" :)

Regarding Coverity, we have access to their free tier for open source projects. I didn't notice that you also contributed to gpsd, really nice!

@lains unfortunately I am not aware of a way to export the report to publish it somewhere, but if you and/or @charlescurley are interested I can send you an invite to get access to the dashboard. I'll just need your email. I can use the email from your git commits, or maybe stop by on IRC real quick? http://webchat.freenode.net/?channels=navit

pgrandin:
> My opinion: If we are going to use coverity, we should deal with all of its complaints properly. That way any new complaint will stand out for immediate inspection and disposal. Same with compiler warnings.

Completely agree. We do have some backlogged work there, but if we get more people involved it would be easier to get rid of it. Hopefully soon we will be able to prevent new warnings/errors from reaching trunk. There's no existing integration for Github/Coverity today but we can probably build one ourselves, and get the results in the pull requests.

pgrandin:
> For coverity, it seems they have an api to pull the reports.

For the paid version of the product yes. I'm not aware of an endpoint for the version that we use. For paying customers there's also a binary cov-manage-im that allows you to pull reports for example.

> It seems there are already some tools like https://github.com/philippegabriel/coverity which could help.

I gave it a shot, but it seems to be quite outdated, and trying to guess the endpoint didn't yield any results.

It would be really, really useful to be able to publish these results.

aerostitch:
@pgrandin There's a button in the UI to export a view as CSV or XML.
Example of export: https://github.com/aerostitch/navit-attachments/blob/master/20180524_outstanding_defects_coverity.csv

lains:
Yes, you can use the email in my commits (@caramail.com). Thanks.

lains:
I have updated the PR to line-up with the indentation settings proposed in #569.

mvglasow:
Any chance of this getting merged soon (what would it take for that)? If this is going to take longer, I would propose committing the quick fix I mentioned in my comment to 4231cac (zero out the whole `debug_message` buffer rather than just its first byte on initialization) so that at least we get clean debug output again.

Right now we’ve had broken debug output for a couple of weeks. Applying the quick fix locally rectifies this but means always having an uncommitted change, which disrupts a couple of git workflows (such as `stash`, `reset`, `checkout` a different branch or `merge`). Committing it to trunk would alleviate this, and if we decide we don’t want to keep the quick fix in the long run, we can always revert the change upon merging this PR.

aerostitch:
@mvglasow did you check the code? If you approve it and that @lains can resolve the current conflicts with trunk I'm ok to merge.
I don't think @charlescurley or @pgrandin still have any open remarks on the current code modification proposal right?

charlescurley:
On Fri, 01 Jun 2018 17:35:47 -0700
Joseph Herlant <notifications@github.com> wrote:

> I don't think @charlescurley or @pgrandin still have any open remarks
> on the current code modification proposal right?

I do not. Whatever of my comments are worth pursuing should be pursued
in separate PRs. And I leave that to our WinCE maintainers.

-- 
"When we talk of civilization, we are too apt to limit the meaning of
the word to its mere embellishments, such as arts and sciences; but
the true distinction between it and barbarism is, that the one
presents a state of society under the protection of just and
well-administered law, and the other is left to the chance government
of brute force."
- The Rev. James White, Eighteen Christian Centuries, 1889
https://charlescurley.com

aerostitch:
@lains can I ask you one last modification so the circleci builds can run please?
You discarded the changes from trunk on debug.c as expected but some changes still need to be done for the style to be respected on this file.
Could you just break the lines longer than 120 chars as shown in this git diff please?
```
$ git diff
diff --git a/navit/debug.c b/navit/debug.c
index b0daa7db..54ae7a7a 100644
--- a/navit/debug.c
+++ b/navit/debug.c
@@ -409,14 +409,16 @@ void debug_vprintf(dbg_level level, const char *module, const int mlen, const ch
     }
 }
 
-void debug_printf(dbg_level level, const char *module, const int mlen,const char *function, const int flen, int prefix, const char *fmt, ...) {
+void debug_printf(dbg_level level, const char *module, const int mlen,const char *function, const int flen, int prefix,
+                  const char *fmt, ...) {
     va_list ap;
     va_start(ap, fmt);
     debug_vprintf(level, module, mlen, function, flen, prefix, fmt, ap);
     va_end(ap);
 }
 
-void debug_assert_fail(const char *module, const int mlen,const char *function, const int flen, const char *file, int line, const char *expr) {
+void debug_assert_fail(const char *module, const int mlen,const char *function, const int flen, const char *file,
+                       int line, const char *expr) {
     debug_printf(lvl_error,module,mlen,function,flen,1,"%s:%d assertion failed:%s\n", file, line, expr);
     abort();
 }
```
It's the last change before merging.
Thanks for your patience,
Joseph

lains:
It's OK @aerostitch, I've performed the changes.
Curiously, it didn't show up in the sanity_check output.

aerostitch:
Give me just a few hours to get #609 merged. It fixes the check on files that come from the merge commits.
Once that's done it won't be failing anymore and you should be good to go.
Thanks for your patience.
Joseph

mvglasow:
I realized only today that this has been merged in the meantime, and I can confirm I am now getting clean debug output again. Thanks a lot!

PR #568, opened 2018-05-02
Add:cleanup:Add a config for indent

aerostitch:
This PR adds a config that we would use for indent.
If we all agree on this config I'll:
* run indent on all `*.c`, `*.cpp` and `*.h` files.
* add a script to the CI that verifies (as a prerequisite to all the builds) that running indent don't modify the (`*.c`, `*.cpp` and `*.h`) files modified by the PR. If it does it will fail.
* add a note in the PR template (and maybe in `contributing.md`)

Would that work for everybody?
I'm totally open to any changes.

### DIFF
--- a/.indent.pro
+++ b/.indent.pro
@@ -1,0 +1,5 @@
+// Configuration file for indent
+--procnames-start-lines
+--k-and-r-style
+--indent-level 8
+--line-length 120


pgrandin:
Thanks for tackling this @aerostitch 

Guidelines are defined here : https://wiki.navit-project.org/index.php/Programming_guidelines 

The formatting aspect can be summarized by : ` indent --procnames-start-lines --k-and-r-style --indent-level 8`

aerostitch:
Ok, sorry for that.
PR updated.

aerostitch:
We might just drop this one if we go with astyle.

aerostitch:
Yep, probably.

Issue #575, opened 2018-05-03
Issues with navit/support/win32/ConvertUTF.c


aerostitch:
Hi,

We currently ship `navit/support/win32/ConvertUTF.c` which is a file that is non-free from the Unicode Inc company.
This causes issues as I have to repackage the tarball to be able to ship in Debian.

There are several software that used to ship it, but, as explained in https://bugs.debian.org/823100, it is being dropped by the different softwares because it's been retired/unmaintained (see http://unicode.org/forum/viewtopic.php?f=9&t=90) and is considered buggy.

The header is only included in:
* `navit/graphics/win32/graphics_win32.c`, there I see 1 usage of `ConvertUTF8toUTF16` [here](https://github.com/navit-gps/navit/blob/92768553ffbd17b2f9a994ecacdb5d746cb36395/navit/graphics/win32/graphics_win32.c#L1210). Can anyone check if it's really needed?
* `navit/vehicle/wince/vehicle_wince.c` but I can't find so I think it could be safely removed from there.

Thanks
Joseph

lains:
@aerostitch, did you try building after removing thoses references?
If it builds fine, I can test it on a WinCE device if you want.

PR #576, opened 2018-05-03
cleanup:wince:remove unused ConvertUTF.h header

aerostitch:
In #575 I noticed that this header seems unused, so cleaning it up.


jandegr:
thx

PR #586, opened 2018-05-10
Android lint check

jandegr:
This adds lint error checking to the Android builds, the result is uploaded in the artefacts, sample https://7955-30791823-gh.circle-artifacts.com/0/reports/lint-results.html
The CI will fail if errors are found, warnings pass.

Also fixes the last Lint errors appearing in the above report (API mismatches) and now CI succeeds with this as report https://7964-30791823-gh.circle-artifacts.com/0/reports/lint-results.html

The fix for the <19 basetheme will make it look just like other apps on your device, from the 19+ values the entries requiring 21+ are removed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,6 @@ version: 2
 defaults: &defaults
   docker:
     - image: ubuntu:14.04
-
 jobs:
   build_linux:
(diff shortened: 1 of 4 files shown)


pgrandin:
Nice!

PR #589, opened 2018-05-12
Android translations

jandegr:
Moves all UI strings from java code to strings.xml.

@pgrandin can you verify the pot file and point me to any errors or shortcomings ?

thx

https://8109-30791823-gh.circle-artifacts.com/0/translations/navit.pot


pgrandin:
Looks like this solves the strings missing from the .pot.

PR #599, opened 2018-05-24
add:doc:addcodefactor.io

jkoan:
Add codefactor.io to Readme since it is used now and has good hints ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
 [![Build Status](https://img.shields.io/circleci/project/github/navit-gps/navit/trunk.svg)](https://circleci.com/gh/navit-gps/navit)
+[![CodeFactor](https://www.codefactor.io/repository/github/navit-gps/navit/badge)](https://www.codefactor.io/repository/github/navit-gps/navit)
+
 
 Navit on Android tablet:
 


jandegr:
Hi,

I'd like to ask the following 

Is it configurable ?
Is there a list somewehere indicating what it investigates.

I have the impression it does not find a lot for java, giving more or less a false OK.

thx

pgrandin:
@jandegr worst case we can start by adding an ignore pattern for the java files : https://www.codefactor.io/repository/github/navit-gps/navit/ignore and then dig to see if we can have it parse more than one language at once?

pgrandin:
@jandegr actually we can use a checkstyle.xml file for java : http://checkstyle.sourceforge.net/style_configs.html

jandegr:
thx, why not give it a go with the checkstyle config from https://github.com/navit-gps/navit/pull/595

https://github.com/navit-gps/navit/blob/Android_checkstyle_2/navit/android/navit_checkstyle.xml

and we can compare with https://8419-30791823-gh.circle-artifacts.com/0/reports/checkstyleMain.html

I don't have access to codefactor, so if one of you could...

pgrandin:
CodeFactor support replied :

>>> Both file name and location are important forcheckstyle. So e.g. https://github.com/navit-gps/navit/checkstyle.xml can be recognized by CodeFactor. You can also check https://www.codefactor.io/repository/github/navit-gps/navit/settings to see whether or not you are using the custom configuration. If your checkstyle.xml file is in the right place, Settings Page will show "Custom" configuration tooltip next to it.

jkoan:
@jandegr @pgrandin so we would need to move the file or set a link?

PR #569, opened 2018-05-02
Cleanup:global:Use astyle to reformat everything

aerostitch:
This PR uses the config file in #568 to reindent everything.
Commands used:
```
find . -iname '*.c' -o -iname '*.cpp' -o -iname '*.h' -exec indent -gnu -l200 {} \;
# indent generates backup files that you need to delete
find . -iname '*~' -delete
```

I put that in a separate PR to allow more clarity.
As we update the other PR I'll update this one too.
1st iteration is using `-gnu --line-length=200`

Note that when I can that I got:
```
indent: ./navit/data.h:38: Warning:old style assignment ambiguity in "=*".  Assuming "= *"
                                                                                
indent: ./navit/data.h:50: Warning:old style assignment ambiguity in "=*".  Assuming "= *"
                                                                               
indent: ./navit/data.h:77: Warning:old style assignment ambiguity in "=*".  Assuming "= *"
                                        
indent: ./navit/data.h:89: Warning:old style assignment ambiguity in "=*".  Assuming "= *"
                                         
indent: ./navit/gui/qml/searchProxy.h:30: Warning:old style assignment ambiguity in "=&".  Assuming "= &"
                                                      
indent: ./navit/gui/qml/ngqpoint.h:268: Warning:old style assignment ambiguity in "=-".  Assuming "= -"
                                          
indent: ./navit/gui/qml/ngqpoint.h:272: Warning:old style assignment ambiguity in "=-".  Assuming "= -"
                                                         
indent: ./navit/gui/qml/ngqpoint.h:319: Warning:old style assignment ambiguity in "=*".  Assuming "= *"
                                                         
indent: ./navit/gui/qml/ngqpoint.h:341: Warning:old style assignment ambiguity in "=*".  Assuming "= *"
                                                     
indent: ./navit/gui/qml/routeProxy.h:18: Warning:old style assignment ambiguity in "=*".  Assuming "= *"
                                                       
indent: ./navit/gui/qml/routeProxy.h:23: Warning:old style assignment ambiguity in "=*".  Assuming "= *"
                                                  
indent: ./navit/support/glib/galias.h:0: Error:Zero-length file ./navit/support/glib/galias.h
indent: ./navit/support/espeak/speak_lib.h:145: Warning:old style assignment ambiguity in "=-".  Assuming "= -"
                                                    
indent: ./navit/debug.h:44: Warning:old style assignment ambiguity in "=-".  Assuming "= -"
                                                    
[...]

PR #603, opened 2018-05-29
add:helper:Add a pre-commit hook script for users to use locally

aerostitch:
Related to #569: This PR adds a pre-commit hook example script that removes trailing spaces, control-M and uses astyle to reformat files automatically.
Instructions are in the header.
This is not something mandatory and we'll probably need to mention it in the doc when we update it.

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+set -eu
+
+# #############################################################################
+# This is a pre-commit hook that allows you to respect the code formatting
+# chosen by the Navit team.
+#
+# To install it, create a folder named "hooks" in the .git directory of this
+# project and just symlink this script inside it. Or run the following script
+# from the root directory of the project:
+#  mkdir .git/hooks
+#  ln -s ${PWD}/scripts/pre-commit ${PWD}/.git/hooks/
+#
+# This will not work on Windows platform unless you are using cygwin.
+# #############################################################################
+
+files=$(git diff --cached --name-only --diff-filter=ACM)
+git_dir=$(git rev-parse --show-toplevel)
+
+# check for common misspells if misspell is installed
+[[ -n "$(which misspell)" ]] && misspell
+if [[ -z "$(which astyle)" ]]; then
+  echo "Unable to find the astyle executable. Please install it to have automatic formatting of your files."
+fi
+
+# Only work on the files that are part of the commit
+for f in $files; do
+  if [[ -e "${git_dir}/${f}" ]]; then
+    if [[ "${f: -4}" != ".bat" ]]; then
+      # Makes sure to not commit ^M
+      [[ -n "$(which dos2unix)" ]] && dos2unix -s -S -q "${git_dir}/${f}"
+      # Removes trailing spaces
+      [[ "$(file -bi """${git_dir}/${f}""")" =~ ^text ]] && sed 's/\s*$//' -i "${git_dir}/${f}"
+      git add "${git_dir}/${f}"
+    fi
+    # Formats any *.c and *.cpp files
+    if [[ "${f: -2}" == ".c" ]] || [[ "${f: -4}" == ".cpp" ]]; then
+      if [[ -n "$(which astyle)" ]]; then
+        astyle --indent=spaces=4 --style=attach -n --max-code-length=120 -xf -xh "${git_dir}/${f}"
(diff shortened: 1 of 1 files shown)


aerostitch:
I don't have a language that I'm sure to be installed everywhere. Bash is the most commonly installed on *nix platforms, that's why it's used here.
I have the chance not to have to touch a windows platform in years, so maybe someone with windows knowledge can add a windows-specific script.

aerostitch:
Hi @jandegr,

> On line 36 it suggests it will check java too, but did not find it in the code.

I forgot to remove the java part from the comment. Good catch. I wasn't sure what was the final tool that you wanted to use for java formatting.

> I have the impression nobody took an in-depth look at wat astyle did ok or not ok for C files but a very quick check revealed it did not a lot better there either.

I think that statement is not really fair. Several people went over the changes and yes, it seems you caught some weirdness afterwards that had not been caught before. It's no reason to diminish other people's work on change reviews.

> Why not install checkstyle on CI for C first like I already did for Java ?

The initial idea was to use one tool for everything, and in #569 we started using astyle. You came up later with the checkstyle project and it's still not clear to me waht you want to use it for. Originally I thought you only wanted it for some java best practices, now you're sayign it's our linter for java and for C?

Also you're suggesting that people using the pre-commit should also have java installed and run checkstyle?

I would really like for us to be coherent in the tools we use for code formatting. Can we make sure everybody agrees on 1 tool and stick to it?

jkoan:
Generally i like the idea of hooks, but i think we should not do it like that. The CI test should only reject the PR and we can handle it like we want.
I know that this brings a bit more overhead but is better for not so advanced users that don't know how to fix such problems. 
Other opinions? @pgrandin

aerostitch:
Note:
The hook is just a helper that you can install only if you want to **on your computer** (nothing to do with the CI), it's not mandatory at all (in case you thought so). Having it in our repo would just allow us to reference in the doc and not having people reinvent the wheel if they want to add such helper.

It will just let you avoid to worry about the formatting and the trailing spaces for you (when you execute your `git commit` command.

jandegr:
Lets all look in the mirror,

In https://github.com/navit-gps/navit/pull/569 you suggest linelength will be limited to 120,
as you stated there
**100% of the responses indicated 120 chars as preferred line length**

I already stated in https://github.com/navit-gps/navit/pull/569 that it ignores the preceding spaces (from the indents)


You ran astyle to limit the codelength to 120 instead of linelength, resulting in lines with up to 120 chars of code + any number of indents.

If that was the intended result you should have been clear about it.

I think it is questionable and I am not alone
https://sourceforge.net/p/astyle/bugs/230/

Excuse me very much for running checkstyle on it before and after the change, I know of no better than to check first, then modify followed by an after check, and then proceed with actions like this PR to enforce it in future commits.

aerostitch:
Hi @jandegr,

> I already stated in #569 that it ignores the preceding spaces (from the indents)

Indeed, with what you added in your last comment it's much clearer Thanks for that.

> You ran astyle to limit the codelength to 120 instead of linelength, resulting in lines with up to 120 chars of code + any number of indents.
>
> If that was the intended result you should have been clear about it.

Not it was not. But now that you've noticed that issue we can say that astyle is not good for length enforcement but it could still be useful for making sure that the function declaration is attached to its opening bracket and its return type no?

> Excuse me very much for running checkstyle on it before and after the change

It's already been stated several times that the work you did on checkstyle is awesome.
The disagreement in the previous response was on the aggressiveness of the comments around the work of the reviewers after the merge.
The other PR was open for more than a month and easily reproducible but a lot of changes at the same time often lead to misses and this was no exception.
Maybe if we would have checked the checkstyle reports before the merge we would have caught it earlier.

> then proceed with actions like this PR to enforce it in future commits.

Again, this is just a helper that you can install locally to make some minor code style change while you commit. It's not meant to enforce anything and is not mandatory.

The enforcement:
* of the trim of line endings (using sed) and the attachement of return code and opening brackets (using astyle) is be done with https://github.com/navit-gps/navit/pull/606
* of the use of checkstyle will be done with https://github.com/navit-gps/navit/pull/608

I think both could be complementary but if you have a better solution, we can switch to it. I'm totally open to change.

jkoan:
I will merge that now because it's only locally and does not conflict with anything. Even though there are two ways to c hack the code style, I think both have there advantages and there for we should use both to check the style. Anyone can optimize this check script if he wants to

PR #605, opened 2018-05-29
Cleanup:global:force the re-attach the return type to the function declaration

aerostitch:
It seems that there was a miss in #569: the `-xf` and `-xh` options of astyle have been forgotten, so the attachement of the return types was not done.
This PR corrects this for C and Cpp files.
Command-line used:
```
find . -type f   \
  \( -not -path './navit/support/*' -a -not -path './navit/fib-1.1/*' \)  \
  -a \( -iname '*.c' -o -iname '*.cpp' \) \
  -exec  astyle --indent=spaces=4 --style=attach -n --max-code-length=120 -xf -xh {} \;
```

### DIFF
--- a/navit/android.c
+++ b/navit/android.c
@@ -67,8 +67,7 @@ int android_find_static_method(jclass class, char *name, char *args, jmethodID *
 }
 
 JNIEXPORT void JNICALL Java_org_navitproject_navit_Navit_NavitMain( JNIEnv* env, jobject thiz, jobject activity,
-        jobject lang, int version,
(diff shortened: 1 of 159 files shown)


aerostitch:
Note that the already broken function declarations don't get reattached, resulting in stuffs like:
```
JNIEXPORT void JNICALL Java_org_navitproject_navit_NavitAddressSearchActivity_CallbackCancelAddressSearch( JNIEnv* env,
        jobject thiz,
        jlong handle) {
```
Cleaning that up.

PR #583, opened 2018-05-07
fix:config:Fix dtd to allow real xmlint

jkoan:
This PR fixes the navit.dtd

Test Command:
xmllint --noout --dtdvalid navit/navit.dtd navit/navit_shipped.xml

Output before:
> `navit/navit_shipped.xml:36: element navit: validity error : Element navit content does not follow the DTD, expecting (graphics , gui+ , log* , osd* , vehicle* , tracking , vehicleprofile* , route , navigation , speech , mapset+ , layout+)*, got (graphics gui gui log osd osd osd osd osd osd vehicle vehicle tracking vehicleprofile vehicleprofile vehicleprofile vehicleprofile vehicleprofile vehicleprofile vehicleprofile vehicleprofile route navigation speech mapset mapset mapset mapset layer layout layout layout layout layout layout layout )
> Document navit/navit_shipped.xml does not validate against navit/navit.dtd`
RC: 3

After the Patch:

> ` `
RC: 0

PR #606, opened 2018-05-29
add:ci:Add sanity check script on circleci

aerostitch:
The idea of this is to add a sanity check before even the builds start.
This just remove the trailing spaces and runs astyle with what we did in #569 and #605.
It could be, for example, extended with the nice work @jandegr is doing on checkstyle with Java.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,16 @@ defaults: &defaults
   docker:
     - image: ubuntu:14.04
 jobs:
+  sanity_check:
+    <<: *defaults
+    steps:
+      - checkout
+      - run:
+          name: Setup requirements
+          command: apt-get update && apt-get -y install git astyle sed
+      - run:
+          name: run sanity check script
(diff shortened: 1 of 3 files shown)


aerostitch:
We should probably add something that enforce @jkoan 's work from https://github.com/navit-gps/navit/pull/583 in there.

.circleci/config.yml
```diff
@@ -3,6 +3,16 @@ defaults: &defaults
   docker:
     - image: ubuntu:14.04
 jobs:
+  sanity_check:
+    <<: *defaults
+    steps:
+      - checkout
+      - run:
+          name: Setup requirements
+          command: apt-get update && apt-get -y install git astyle sed
```
jkoan: Please use setup_common_requirements.sh for that

aerostitch: Done.

aerostitch: Note that this step doesn't need most of what's in `setup_common_requirements.sh`. Only git, sed and astyle are needed there.

PR #609, opened 2018-06-02
update:ci:Refactor the sanity check for better output and add xmllint check

aerostitch:
This PR:
* Adds a check for the xmllint if modified in the PR as a follow-up from #583
* removes the `set -x` so that it has a more understandable output
* makes sure that the diff is done for each check and each file, not just one big diff that would make it less understandable
* only run the check on each file once (using `| sort | uniq`)
* show all the errors before exiting (so I had to remove `set -e`)
* add a warning in case people try to run it locally

### DIFF
--- a/scripts/ci_sanity_checks.sh
+++ b/scripts/ci_sanity_checks.sh
@@ -1,11 +1,27 @@
 #!/usr/bin/env bash
-set -x
-set -eu
+set -u
 
-# #############################################################################
+# #####################################################################################################################
 # This script runs on circle CI to verify that the code of the PR has been
 # sanitized before push.
-# #############################################################################
+#
+# WARNING: make sure you commit ALL your changes before running it locally if you ever do it because it will run a git
+# checkout -- which will reset your changes on all files...
+# #####################################################################################################################
+
+return_code=0
+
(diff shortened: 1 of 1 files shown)


scripts/ci_sanity_checks.sh
```diff
@@ -1,38 +1,62 @@
...
+}
 
 # List the files that are different from the trunk
 from=$(git rev-parse refs/remotes/origin/trunk)
 to=$(git rev-parse HEAD)
 interval=${from}..${to}
 [[ "${from}" == "${to}" ]] && interval=${to}
 
-for f in $(git show -m --pretty="format:" --name-only ${interval}); do
+for f in $(git show -m --pretty="format:" --name-only ${interval} | sort | uniq); do
```
jkoan: sort | uniq --> sort -u

aerostitch: Good catch!  I always forget that this switch is available! :)
Updated.

aerostitch:
In the last commit I switched from `git show` to `git diff` so that we only check files that differ from the trunk and avoid checking files that are a result of a branch merge commit. Let me know if I should revert that.
Basically the result is that it only checks that show as changed in the PR in the end.

PR #612, opened 2018-06-04
update:style:optimize the usage of cat in shell scripts

aerostitch:
This PR removes the warnings in codefactor about "Useless cat usage": https://www.codefactor.io/repository/github/navit-gps/navit/issues?category=Performance&groupId=1692
Note that it also remove unnecessary copy and redirect of temporary file by using the "in-place" capabilities of sed.


aerostitch:
Thanks! :)

PR #621, opened 2018-06-04
Revert #612 for sailfish build script

metalstrolch:
Whatever the 'enhancement' done in #612 should bring, it causes grep
to complain for non existing file in Sailfish's MER SDK.
Most probably because the escaping of the parenthesis was wrongly
considered. Never mind - revert.

### DIFF
--- a/contrib/sailfish/build_on_sailfish_sdk.sh
+++ b/contrib/sailfish/build_on_sailfish_sdk.sh
@@ -8,9 +8,9 @@
 if [ -z ${VERSION_ID+x} ]; then echo "VERSION_ID not set. Forgot to export VERSION_ID?"; exit 1; fi
 
 #arm devices
-sb2 -t SailfishOS-${VERSION_ID}-armv7hl -m sdk-install -R zypper in $(cat navit-sailfish.spec | grep "^BuildRequires: " | sed -e "s/BuildRequires: //")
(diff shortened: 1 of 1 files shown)


aerostitch:
Hi @metalstrolch,
Why a cat into grep?
I actually just noticed the issue I've done in #612 which is put the file name in 1st argument instead of the 2nd one.
Could you change it in this way instead of a `cat |` please?
So that would look like:
```

aerostitch:
Nevermind, I'll push the change to the branch.

metalstrolch:
Why? Using cat that way is something that was always done. I think the old way of doing it is way more readable.
But you see the Problems arising with all these fancy checkers trying to fix things that aren't even broken.

aerostitch:
It's to avoid a unnecessary pipe.
`grep` is able to read files directly and it has been for years.
Error is human. Sometimes swapping 2 argumetns happen...

aerostitch:
Would https://github.com/navit-gps/navit/pull/621/commits/6148aec3a90a51fdf294cb5d50c7a894b2853339 be acceptable for you?

metalstrolch:
Seems to be OK, if that makes you happy.

aerostitch:
Thanks for the fix!
Sorry for the issue introduced.

PR #624, opened 2018-06-05
WIP: update in-tree espeak to espeak-1.48.04

metalstrolch:
Update the C-ified version of espeak to espeak-1.48.04.
Currently builds and links for Sailfish. Untested for
Windows builds yet.
Throws a lot of warnings still.

DO NOT MERGE YET


metalstrolch:
- Can I somehow disable all those coding style checks for this in-tree 3rd party package?
 Don't think it's a good ide to mess around with espeak's data files.

- Can I temporarily disable this new sanity check? I want to know of this builds for Windows prior being picky about some silly style guide.

aerostitch:
Give me a min,  I'm working on a PR to disable the sanity check on the files in `navit/support`.

aerostitch:
Done in #625. Will be active once it's reviewed and merged.
Note that the sanity check is only failing because of trailing whitespaces right now. astyle itself is not running on the files.

metalstrolch:
OK, now somebody should test the result on Windows. Unfortunately I don't have any windows.

jkoan:
Only to understand: Shouldn't we remove external things in the future and link against library's? And for Win we should build externally and link all together? or do i only understand everything totally false?

mdankov:
@jkoan, IIRC, espeak copy is here for two reasons:
- there were no support for espeak on Windows CE platform. Unsure if wince is supported now by espeak.
- we decided to port espeak to C to keep memory usage low. It's critical for WinCE platform as these devices are way outdated now, having 64Mb of RAM on board is typical. 

@metalstrolch, 
We did not use builtin espeak on any platform besides WinCE a couple years ago, so testing on desktop Windows isn't going to make sense anyway.

I will try to reanimate my WinCE device and give you some feedback in a week.

metalstrolch:
I already use in tree espeak for my sailfish os builds since some time, as sailfish doesn't have libespeak and sailfish disallows library packages. So since espeak is in tree I used it.
But of course the speech/qt5_espeak builds against external libespeak as well.

Windows (and WinCe) builds would NOT build against stock libespeak currently because speech/espeak does not use espeak the usual way.

I just refreshed the espeak port to be based on latest espeak version which removes a lot of warnings on x86 and x86_64 builds and besides that comes with more languages.

lains:
@mdankov, I have played with a WinCE device recently, so I can confirm that:
* espeak-data is missing from zip, and this it not related to your change.
* navit.xml could be tuned much more for WinCE than it is right now.

I don't have a WinCE build environment setup right now, so I prefer not to make changes to WinCE build.

metalstrolch:
My intention was **not** to fix WinCE package, config or behaviour, but to update in tree espeak for Sailfish use.
It is currently used for sailfish builds as there are no espeak packages in Sailfish, and Sailfish's Harbour rules would not permit having library packages as dependencies. And since espeak is in tree already I just used it.

All Navit Qt5 code dealing with espeak can as well do so against stock out of tree espeak libraries. Default build system will do so if espeak is found on build system. So in tree espeak is not required for QT5 speech builds.

What to do for me?
- Merge this as it works for Sailfish ignoring possible WinCE problems?
- Block this until someone finally decides to drop WinCE as it seems nobody cannot even test it anymore? Or will somebody finally care?
- Fork Navit for Sailfish to include this, not caring about wince?

lains:
@metalstrolch, I have recently managed to install navit on a WinCE device. This is not my main device though, and I don't have it right now, but I should be able to test when I get it back.
I am currently submitting patches for the WinCE build to pass again (https://github.com/navit-gps/navit/pull/632), this will fix the issues on WinCE's espeak that are not related to your changes.
When I have my WinCE device back with me, I can try you changes to tell you if it works as with the current espeak.

lains:
I have tested the compilation on WinCE and I builds fine with this new version of espeak (although I don't have my WinCE device to test it right now).

lains:
Does this solve ticket 997 opened on trac?
http://trac.navit-project.org/ticket/997

metalstrolch:
Most probably not, or at least not directly, as the problem described in http://trac.navit-project.org/ticket/997 points more to missing locale problem in Windows CE. "Old" espeak could happily speak German ever since. At least I used it in sailfish before as well, and I'm German native speaker.

But maybe @pgrandin fixed it by accident in #680

metalstrolch:
I recently bought a Windows CE palmtop on flea market. I'll give it a try once I find a suitable power supply :)

PR #625, opened 2018-06-05
update:scripts:change the sanity script to exclude fib and support

aerostitch:
As requested in #624, removes checking for trailing spaces and other style-related elments on `navit/support`. Also added `navit/fib-1.1` to it as it is a 3rd party lib.
Note: the indentation change is just to make sure we're also using 4 spaces indentation in the script (I was using 2 spaces before).

You can use the diff settings button when reviewing to get rid of the white spaces diffs which makes the review way easier.


aerostitch:
Thanks!

PR #627, opened 2018-06-19
Fix latlon2bookmark

jkoan:
This PR Fixes two minor Bugs wit latlon2bookmark. Now its possible to compile it again


pgrandin:
Thanks @jkoan !
Maybe we should add a step in the CI to test the different scripts / extra tools that we have?

jkoan:
Or add it to cmake build?

pgrandin:
Yeah we could have an "extra" or "tools" section there.

jkoan:
I will merge this now and add other things like the cmake integration later

PR #631, opened 2018-07-14
Refactor:gui/internal: Add pointer and label when using function "view on map"

lains:
When selecting GPS coords or a town/street, and choosing "View on map", the display is centered on the selected location, but there is no visible pointer on this location.
In order to implement this, I have am re-using the code that is already existing to handle action "Show results on the map", and that is already able to display multiple result pointers on the map.

Here is an example of the result when selecting "View on map" for "An der Hauptfeuerwache" in Mûnchen.
example_view_on_map

I added some code to split long strings to multilines, in order to avoid having the text spread too much accross the map. This is also useful when using GPS coords:
example_view_on_map_coords

### DIFF
--- a/navit/graphics.c
+++ b/navit/graphics.c
@@ -1977,24 +1977,24 @@ static int limit_count(struct coord *c, int count) {
 }
 
 /**
- * @brief Draw a multi-line text close to a specified point
+ * @brief Draw a multi-line text next to a specified point @p pref
  *
  * @param gra The graphics instance on which to draw
- * @param gc The graphics color to use to draw the text
- * @param gc The graphics background color to use to draw the text
+ * @param fg The graphics color to use to draw the text
+ * @param bg The graphics background color to use to draw the text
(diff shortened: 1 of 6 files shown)


navit/graphics.c
```diff
@@ -1991,7 +1967,70 @@ static int limit_count(struct coord *c, int count) {
...
+        gra->meth.draw_text(gra->priv, fg->priv, bg?bg->priv:NULL, font->priv, label_lines[label_linepos],
+                            &pref, 0x10000, 0);
+        pref.y+=line_spacing;
+    }
+    g_free(input_label);
+}
+
 
+/**
+ * @brief Draw a dispayitem element
```
pgrandin: small typo here : dispayitem

lains:
I've fixed the typo.

lains:
PR #729 submitted to add a similar feature on the GTK GUI.

PR #632, opened 2018-07-14
Fix:scripts:WinCE build script (remove sample map, add espeak-data)

lains:
Remove inclusion of sample map in zip package (which, in addition, fails currently because map server is down)
Allow running build_wince.sh script twice by performing cleanup on directories if already created
Adding missing espeak-data to zip package (the content of this directory is currently missing in WinCE output package and makes navit speech synthesis fail unless manually copied on the WinCE device)


lains:
xslt processing is not done on the WinCE environment, on the cmake logs, we can see
```
No XSLT processor available.  You have to configure navit.xml yourself, or
  install an XSLT processor (supported:
  saxonb-xslt;saxon;saxon8;saxon-xslt;xsltproc;transform.exe).
```
... and the Ubuntu on which the WinCE is running is quite old (Ubuntu 8.04.4 LTS), so even the saxon lib that was available then does not allow xslt processing.
xsltproc is available however, so with xsltproc installed, the required windows.xslt now applies.

PR #633, opened 2018-07-15
Fix Android apk VersionCode 

jkoan:
This PR Fixes the version code of the resulting apk to get back the old behavior. 
In a further PR I will also update the circleci build so it is more clean and can publish signed apks again to google play.


lains:
Tested... works well on my android build.

pgrandin:
>  to get back the old behavior

Can you share an example of the result?

jkoan:
screenshot_20180718-160014
That's what it look like if you view the installed apps in a app overview app (in this example es file explorer)

lains:
Just a thought, about Version Name, https://semver.org/ recommends to use + before the build number, so version name could be `0.5.1-876+3fbbaab`. Is this the very same value displayed in the "About" dialog?
On version code, maybe it would be good to display the full year `201807150412` instead of `1807150412` (if it's a free string... I don't know if there's a format to respect). I think it's quicker to spot the fact that this string is a date this way.

jkoan:
hi @lains 
thats exactly what it will look like:
0.5.1-876+g3fbbaab 
We only have one digit more at git Hash and the version 0.5.0 in my screenshot is only because my fork is somewhat broken, but the main repo is ok and will give 0.5.1 as release tag.
And regrading the versioncode it need to be counted onwards. So YYYY instead of YY will work but this wont give a gig benefit.

lains:
OK, this is fine. My comment was just about the `+`, I was not really worried about the `0.5.1`, I actually made a typo while copying the version because I took the one displayed on my phone.

jkoan:
I will merge this now since it's important to get updates out on Google Playstore

lains:
Does this fix [issue 1366 opened in trac](http://trac.navit-project.org/ticket/1366)?

PR #636, opened 2018-07-27
Fix:gui/internal:Handling "spacing" attribute properly for tag <gui type="internal">

lains:
Feature was introduced in SVN r1456 (commit 5dea1c6d5ea177a39ba0a9ce29cd247a30005960) but was not applying tag properly from config file

### DIFF
--- a/navit/gui/internal/gui_internal.c
+++ b/navit/gui/internal/gui_internal.c
@@ -525,7 +525,8 @@ void gui_internal_apply_config(struct gui_priv *this) {
     if(this->config.spacing == -1 ) {
         this->spacing = current_config->spacing;
     } else {
-        this->spacing = current_config->spacing;
+        this->spacing = this->config.spacing;
+        dbg(lvl_info, "Overriding default spacing %d with value %d provided in config file", current_config->spacing, this->config.spacing);
     }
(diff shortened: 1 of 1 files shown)


jkoan:
Hi @lains 
Could you please describe what the feature is for? I don't have any idea where a "spacing" is needed.

lains:
This seems to be an old feature of the internal GUI, but that seems to be broken maybe since the beginning.
It allows to override the default spacing of icons in the internal GUI menus, but is is also used to decide the vertical spacing of many internal GUI windows.

Example with the default value:
spacing_default

When overriding using <gui type="internal" spacing="30">
spacing_30

When overriding using <gui type="internal" spacing="3">
spacing_3

Setting the `spacing` attribute in navit.xml set the value `spacing` in `struct gui_priv` see https://github.com/navit-gps/navit/blob/trunk/navit/gui/internal/gui_internal_priv.h#L55

lains:
Note: this has impact (for internal GUI) on menus and submenus, and spacing attribute is set to different values for the various screen definitions on Android (ldpi, mdpi, hdpi, xhdpi, xxhdpi, xxxhdpi), so there will be an impact on how menus display on Android.

PR #646, opened 2018-08-30
cleanup:scripts:Fix condefactor complains "String concat is not required here; use a single string instead"

aerostitch:
Not even sure we still use this script but it seems a little cleanup would make codefactor happy, so here it is.

Does anyone know how it's used?


pgrandin:
It's probably outdated. But fixing the lint won't break anything, so let's do it.

Issue #629, opened 2018-06-27
Bug: error while loading shared libraries: libfib.so


naggety:
I get this error when installed navit in a custom device (using Buildroot).

Doing `find / -name libfib.so` I get that the library is located in `/usr/lib/navit/libfib.so`. The problem is that the directory `/usr/lib/navit` is not in the shared libraries search path, `/usr/lib` is.

Moving `libfib.so` to `/usr/lib` navit runs fine.

I don't have problems with other shared libraries like plugins, support, etc. I guess that they are loaded in a different way.

Possible solutions:

- Make libfib.so be installed in `/usr/lib` (or better said, in `${CMAKE_INSTALL_PREFIX}/${LIBDIR}`). Now libfib installation is made from `navit/fib-1.1/CMakeList.txt`
- Make navit executable to have `/usr/lib/navit` in its rpath. I've done this with the patch I post below, but I don't know if this solves the problem in all cases (for example, it seems that this wouldn't apply if building for Android, but I don't know how building for Android differs from building for Linux, so I don't know if this problem also happen in Android).

Note: this patch should work, but I haven't been able to test it because Buildroot seems to be removing the RPATH property from the executable.

```
diff -Naur navit-v0.5.1/navit/CMakeLists.txt navit-v0.5.1-2/navit/CMakeLists.txt
--- navit-v0.5.1/navit/CMakeLists.txt	2018-06-27 14:54:05.808344052 +0200
+++ navit-v0.5.1-2/navit/CMakeLists.txt	2018-06-27 14:40:37.000000000 +0200
@@ -81,6 +81,7 @@
    endif()
    add_executable(navit ${NAVIT_START_SRC})
    target_link_libraries (navit ${NAVIT_LIBNAME})
+   set_target_properties(navit PROPERTIES INSTALL_RPATH ${CMAKE_INSTALL_PREFIX}/${LIB_DIR})
    if(DEFINED NAVIT_BINARY)
       set_target_properties(navit PROPERTIES OUTPUT_NAME ${NAVIT_BINARY})
    endif(DEFINED NAVIT_BINARY)
```

aerostitch:
Hi @naggety,
Thanks for reporting this.
Which version are you using?
There has been a PR (#574) since last release to fix the lib path when overwriting `LIB_DIR`. Do you overwrite your `LIB_DIR`?
Thanks
Joseph

naggety:
[...]

PR #647, opened 2018-08-30
[On_HOLD] fix:build:Fix an issue with buildroot unable to find libfib in its rpath

aerostitch:
Closes: #629
This is a patch from @naggety, see mentionned issue.
Needs to be tested, but I want to verify that it doesn't break any of our builds first.


pgrandin:
While the patch might fix the issue for the OP, I am wondering if the root cause isn't somewhere else. I've used buildroot extensively myself  ( https://github.com/pgrandin/navit-buildroot/tree/master/packages/navit ) and haven't faced that issue. I need to dig into the OP's package code. We might want to wait before merging that.

jkoan:
The difference between your two implementations are that @pgrandin set the LIB-DIR but @naggety didn't set it.

metalstrolch:
There seems to be something else strange as well around fib.

On buildroot setup i try to build navit against qt5, but without plugin support. Still libfib gets built seperately. (I get libfib.so)
Doesn't happen on my Gentoo-PC and obviously not on sailfish MER build environment. Because there is no libfib.so then.

What's different on building fib?

metalstrolch:
OK, found the libfib oddity with buildroot:
libfib is added to the project by simple "add_library" cmake function without explicitly stating how to link it. So the default behaviour of CMAKE in this case depends on the presence of the BUILD_SHARED_LIBS variable.

Buildroot sets BUILD_SHARED_LIBS environment on cmake to "ON". If unset (like on sailfish or my PC) the behaviour defaults to "NO".

For buildroot this can be fixed by explicitly stating BUILD_SHARED_LIBS in cmake run.
**But I suggest to alter supportlib_add_library macro to honour "USE_PLUGINS" setting as well.**

Or does anyone have better ideas?

jkoan:
Please continue the dicussion within #629

PR #650, opened 2018-09-02
Refactoring:navit/maptool:build with external protobuf-c

bignaux:
depend on #651 : need target to be updated.


navit/maptool/CMakeLists.txt
```diff
@@ -1,16 +1,44 @@
+find_package(Protobuf-c REQUIRED)
```
jkoan: This should be in the if statement, so that navit without maptool can build even without protobuf

cmake/FindProtobuf-c.cmake
```diff
@@ -0,0 +1,86 @@
+#  PROTOBUF-C_FOUND - System has libyang
```
jkoan: Why Libyan?!

navit/maptool/CMakeLists.txt
```diff
@@ -1,16 +1,41 @@
-
 if(BUILD_MAPTOOL)
+
+	find_package(Protobuf-c REQUIRED)
```
jkoan: This should be in the main CMakeLists.txt and it it is not found disable the maptool build, and it should even fail if manually specified that maptool should be build

navit/maptool/CMakeLists.txt
```diff
@@ -1,16 +1,41 @@
...
+
+  # cmake_policy(SET CMP0026 OLD)
```
jkoan: Old comments that are not needed anymore

bignaux: in fact they will need later.

navit/maptool/CMakeLists.txt
bignaux: i could look that but it's bad idea to centralise in the main CMakeLists.txt, things should be closest to what they interfere with, the big big main make things harder to read and improve.

cmake/FindProtobuf-c.cmake
bignaux: i pickup from them and i would like to not modify it since it seems to work that way. Best effort for the moment, no need to clean stuff that could be change very soon.

navit/maptool/CMakeLists.txt
jkoan: This is a policy we should not need to set, we should only fix the problem, but okay
for now

cmake/FindProtobuf-c.cmake
jkoan: This file is apache licensed... @pgrandin what do you think?

navit/maptool/CMakeLists.txt
bignaux: yes i've to remove the function and call the implicit mecanism but for now i don't have understand that well, see that later.

cmake/FindProtobuf-c.cmake
pgrandin: Mixing ASL with our GPL2 is tricky indeed. We need to look if we already did it somewhere, and we need to check what's defined as linking (as this seems to be a way out of the issue).

https://opensource.stackexchange.com/questions/1357/can-i-link-a-apache-2-0-library-into-software-under-gplv2

bignaux: The entire protobuf this patch removes was ASL. This is one file used only at build time.

navit/maptool/CMakeLists.txt
jkoan: #666 fixes this problem

cmake/FindProtobuf-c.cmake
jkoan: according to the Doc cmake itself has FindProtobuf: https://cmake.org/cmake/help/v3.0/module/FindProtobuf.html

jkoan: Okay, the support is only for c++ compile, not for c, so we need to rewrite this probably. Hopefully we can reuse some of the cmake sources.

navit/maptool/CMakeLists.txt
jkoan: line 5, yes

pgrandin:
@jkoan looks like @bignaux addressed your concerns. Want to update your review?

jkoan:
I don't see any new commits. And the issue with the find protobuf-c has end addressed as well.

pgrandin:
:confetti_ball:

PR #651, opened 2018-09-02
Refactoring:cmake:from 2.6 to 3.2

bignaux:
Need container navit/tomtom-build-image to be upgrade :
  CMake 3.1 or higher is required.  You are running version 2.8.12.2

Need wince targets and others to be fixed to by theirs maintainers or be drop from default CI build. we can't be stuck to very old version of software because others parts are not playing the game of upgrade software time to time. Please up trusty at least to xenial ...


jkoan:
The problem with expachially WinCE is that it is WinCE and we have no way to update the OS because the Toolchain is so old that its only running on such old OSs. This is a Problem we encountered because the WinCe Toolchain is not updated anymore, but we dont want to drop support for it because most of the common navigation devices are WinCE devices (especially those cheap ones from ebay, and devices which are build into cars). So we dont want to drop a large userbase.

But i denenetly agree that we should update cmake and probably also everythink that is possible on OS side for every build, but we may not drop device support and there for targets

bignaux:
Anyway , for the moment i think we have to update the old trusty and stretch env to latest ubuntu LTS. I try in another PR.

PR #652, opened 2018-09-03
[WIP] refactoring:switch build to latest ubuntu LTS

bignaux:
* CMake is progressively dropping support and we HAVE to update it.
* Some buildenv seems not maintained for _years_ (or month), if we can't spend time on them, use available and saves time to reinvent the wheel.
* unify on latest ubuntu LTS would save time fixing packages dependencies, configuration and scripts. 

Please avoid to split code between dockerfile, circleci , and bash scripting. Do the maximum in the scripts, so we don't have to fix on upper software. 

UPDATE : with tomtom-ng , oldest cmake is V3.2.0 from wince. This image could be update later thanks to http://max.kellermann.name/projects/cegcc/ . 

@pgrandin @jkoan Should i continue this work or useless ?

### DIFF
--- a/po/bg.po.in
+++ b/po/bg.po.in
@@ -1,5 +1,5 @@
 # Bulgarian translations for navit
-# Copyright (C) 2006-2015 The Navit Team
+# Copyright (C) 2006-2018 The Navit Team
 # This file is distributed under the same license as the navit package.
 # Many thanks to the contributors of this translation:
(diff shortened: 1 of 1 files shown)


pgrandin:
> Should i continue this work or useless ?

Definitely useful to update our base image. Thanks for looking into this!

PR #661, opened 2018-09-10
[WIP] cleanup:graphics:Make sure we use g_strcmp0 instead of strcmp

aerostitch:
Everywhere else in that file (especially for this variable), we use the `g_*` functions to manage the memory. It would only make sense that we use the `g_strcmp0` instead of `strcmp` there too.
Plus, `g_strcmp0` handles `NULL` better.

### DIFF
--- a/navit/graphics.c
+++ b/navit/graphics.c
@@ -135,7 +135,8 @@ struct displaylist_icon_cache {
 
 };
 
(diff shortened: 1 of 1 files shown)


aerostitch:
@pgrandin maybe we should update our glib support?
Do you know how we do that? Are there restrictions about it?

aerostitch:
Closing for now as it requires a refactor of the way we embed glib, which we don't really have the bandwidth for.

PR #662, opened 2018-09-10
[WIP] cleanup:navigation:Make sure we use g_strcmp0 instead of strcmp

aerostitch:
Moving `strcmp` to `g_strcmp0` in `navit/navigation.c`.


aerostitch:
Maybe we should update our glib support?
Do you know how we do that? Are there restrictions about it?

pgrandin:
Yes it looks like an upgrade is needed as per the discussion on IRC.
Going to mark this PR "WIP" for now.

aerostitch:
Closing for now as it requires a refactor of the way we embed glib, which we don't really have the bandwidth for.

PR #668, opened 2018-09-16
[WIP] Switch to CMake ExternalProject_Add instead of support for glib

aerostitch:
This is a first tentative to use cmake's external projects instead of including glib via support.
There are a bunch of cleanup to do later on, but let's see if we manage to build everywhere with that.


aerostitch:
Working on this locally for now to avoid hitting circleci quotas, please don't push anything to this branch for now.

CMakeLists.txt
```diff
@@ -260,7 +261,28 @@ if (Glib_FOUND)
 	include_directories(${Glib_INCLUDE_DIRS})
 	list(APPEND NAVIT_LIBS ${Glib_LIBRARIES})
 else(Glib_FOUND)
-	set_with_reason(support/glib "Glib not found" TRUE ${INTL_LIBS})
+	set(EXTERNAL_GLIB_SORT_VERSION 2.58)
+	set(EXTERNAL_GLIB_TAG 2.58.0)
+	ExternalProject_Add(project_glib
+		URL http://download.gnome.org/sources/glib/${EXTERNAL_GLIB_SORT_VERSION}/glib-${EXTERNAL_GLIB_TAG}.tar.xz
+		URL_HASH SHA256=c0f4ce0730b4f95c47b711613b5406a887c2ee13ea6d25930d72a4fa7fdb77f6
+		# Pulling from the repo forces us to use autogen and autoreconf
```
jkoan: What would be the problem with this?

CMakeLists.txt
```diff
@@ -260,7 +261,28 @@ if (Glib_FOUND)
...
+		# URL https://gitlab.gnome.org/GNOME/glib/-/archive/${EXTERNAL_GLIB_TAG}/glib-${EXTERNAL_GLIB_TAG}.tar.gz
+		# URL_HASH SHA256=e1b8e10da740da41260e5178346d57f4a5458d6bf300af91cdf4edee76d3437e
+		# git is to old on the wince image to be used here :(
+		# GIT_REPOSITORY https://gitlab.gnome.org/GNOME/glib.git
+		# GIT_TAG ${EXTERNAL_GLIB_TAG}
+		PREFIX ${CMAKE_CURRENT_BINARY_DIR}/glib-${EXTERNAL_GLIB_TAG}
+		CONFIGURE_COMMAND "./configure"
+		BUILD_COMMAND make
+		INSTALL_COMMAND make install
+		PREFIX=${CMAKE_CURRENT_BINARY_DIR}/glib-${EXTERNAL_GLIB_TAG}
```
jkoan: This Line occurs twice

CMakeLists.txt
aerostitch: The issue is t that lmza is too old on the image and is unable to extract the xz as it is in a too new format apparently. That's why I was working on the wince image refactoring (the image is still not working as I was explaining in https://github.com/navit-gps/Dockerfiles/pull/4, so as long as it's not fixed, I cannot move forward on this PR.

jkoan: but why we cant use the git repo and compile with autoreconf and autogen?

aerostitch:
I haven't had the time to go much further lately so if you got more time to look at it, knock yourself out. Just put a note in the PR so that we don't both work on it at the same time.
Thx!

CMakeLists.txt
aerostitch: Because I didn't want to re-add autoconf & co back in the build process. Seems tedious.

aerostitch:
> seems a bit over complicated to try to build glib2 from navit source instead of use binary from http://win-builds.org/doku.php/1.5.0_packages#glib2_2421-1_-_library_of_c_routines for example and install them via circle-ci (to avoid distro specific stuff in script directory).

Sounds interesting. Do they support wince too?
Note that from the discussions I had with @pgrandin it seems the glib version in support was a patched version and nobody knows which version it was and what the patch was! :)

> keep the build system as simple as possible and let distro packager provides the right dependancies is for me a better way to do.

Agreed, this is only used for platforms that don't have glib available, like the other libs that are present in support/

I don't have time to have a look further right now but if you want to give it a shot and use win-builds, feel free to do so @bignaux

bignaux:
for win32 plateform, we need to enable gtk ui (currently only internal is available) so we need at least atk cairo gtk2 gdk_pixbuf pango imlib2 . So ExternalProject_Add for each would transform our cmake to a package manager ... Have you try to build for wince with this 2.58 version ? I'd really like to clean stuff (ezxml for example) and we need this glib update.

jkoan:
@aerostitch let's cose this PR for now but open a issue to rember that we want to update some build things later (like glib in build images and so on)

PR #669, opened 2018-09-16
update:ci:deprecate setup_wince.sh script

aerostitch:
Follow-up of https://github.com/navit-gps/Dockerfiles/pull/3

### DIFF
--- a/scripts/setup_wince.sh
+++ b/scripts/setup_wince.sh
@@ -1,4 +1,10 @@
 #!/bin/bash
+
+# #############################################################
+# This script has been deprecated. If you need to add packages,
+# add it to the wince/Dockerfile on the navit/Dockerfiles repo.
+# #############################################################
+
 set -e
 
 mkdir -p /var/lib/apt/lists/partial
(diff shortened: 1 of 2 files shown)


aerostitch:
Looks like installing `git-core` brings a bad version of git that doesn't properly support to git clone in `.`.
The upgrade of the image starts to be a blocker for me. I guess I'll have to make it a priority :disappointed:

scripts/setup_wince.sh
```diff
@@ -1,4 +1,10 @@
 #!/bin/bash
+
+# #############################################################
```
jkoan: please put this in echo so it also print out when this script runs

aerostitch: Done.

aerostitch:
Scanning dependencies of target git_version
-- git not found, cannot record git commit
-- Git commit: xdevxgitxnotxfound
-- Git   date: 7001010000
[  0%] Built target git_version

jkoan:
@pgrandin yes. But I head the idea to add a fake gut program in the path witch handles the "." directory by checking out to a temp directory and then move everything once finished

aerostitch:
I'll close this one until I get more time to work on the image upgrade for wince as it's quite dependent on it to work properly.

PR #672, opened 2018-09-29
First merge for traffic branch

mvglasow:
I would like to do a first merge of the traffic branch I’ve been working on for the past months.

See here for what works and what doesn’t (yet): https://github.com/navit-gps/navit/projects/8

The traffic module will not do anything unless explicitly enabled by adding the respective line to navit.xml. It is currently not enabled in the default configuration but needs to be enabled manually if you want it, as it is still an experimental feature.

The first (and so far only) functional traffic plugin is the Android one. It receives traffic updates via system broadcasts from external apps; the first app that sends these broadcasts is the upcoming build of Qz.

On Linux, traffic messages can be injected via Dbus messages; this is mainly intended as a test interface, not a production-use one. You will still need to configure a traffic plugin for this to work; there is a `null` traffic plugin which provides all internal functionality but will never send any traffic messages of its own.

Without the traffic module enabled, the main change is that Navit now uses LPA* instead of Dijkstra for routing. This change was necessary in order to do dynamic re-routing as segment costs change. It also allows for a couple of future routing improvements, as we can now add segments to the route graph any time, even after calculating a route. This would allow us to do the following:
* If the vehicle deviates from the route, we no longer have to do a full recalculation (which can take a long time for longer routes, and may leave drivers without instructions just when they need them most). Instead, we can simply add the local roads around the current position (which may not be in the route graph yet) and recalculate the parts affected by these segments.
* Rather than first building a complete route graph and then flooding it, we could first add the surroundings of the destination, start calculating, figure out which area we need to add next, add it and continue calculating, and continue doing this until we have a route. This approach could solve our long-standing route depth issue (Navit ignoring roads lower than trunk on long routes) with fewer resources than the current workaround. Performance implications would need to be tested. Another upside is that this kind of routing is fairly easy to move to an idle loop, so the UI remains responsive while Navit is calculating that 1000-km route.

The new routing is the main reason I would like to do a first merge now. I have run my suite of 78 routing test cases (originally created for Project HighFive) against both the current and proposed implementation, and both produced the exact same results.

Work on the traffic branch will, of course, continue after the merge.

I’ve marked this as WIP to allow me to test this in real life while the change is under review; I will remove the marker when the branch appears stable under real-life conditions.


mvglasow:
Did some real-world tests today, without anything out-of-the ordinary. To the user, LPA* does not look any different from Dijkstra. (I should note that, for the last three months, I have already been running a hybrid, with Dijkstra being used for initial routing and LPA* for partial recalculations after traffic situations changed—no issues with routing.)

mvglasow:
I’ve used this branch for the past few days, and routing works as expected, no changes for the end user. I’ve removed the WIP from the title.

mvglasow:
Done. I’ve also fixed navit.dtd, added support for lane closure and contraflow events, and fixed a bug regarding supplementary information classes.

mvglasow:
I notice the Android build failed, due to some toolchain issue. In the meantime, I’ve successfully built this PR for Android, hence the code is not at fault.

mvglasow:
Re-ran the build, Android seems OK now.

lains:
@mvglasow, it is definitely a nice feature to get real time traffic taken into consideration for routing decisions.
I'm looking forward to using this feature.

Issue #433, opened 2018-04-05
Wrong behaviour on Screen rotation or Windows resize


ChibuikeOruche:
Navit app is a free mobile application which offers an offline navigation system on your mobile device.

#### Expected behavior
When I click on a menu item (or option and rotate the screen, I should remain on my current view and continue with whatever I intended to do with the app.

#### Actual behavior
Unlike the expected behavior, rotating my device after clicking on a submenu reverts me back to the menu thus making me start all over with what I want to do in the app. *It occurs on all submenu items*. 

#### How to reproduce
Open the Navit app. Click on the menu button(bottom right of screen). A menu will be displayed. Click on any menu option of choice. It should display the options in that submenu. Then, rotate your screen.

* Operating system: Android 4.4.2(KitKat)
* Device: Tecno Y6 Build version:1.7.1

#### Recording Of The Bug
![Navit_bug.mp4](https://cdn.utopian.io/posts/3a492cebe4ebca72bc11fdcc2bda70735fdaNavit_bug.mp4)

    

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@goodluckleo/wrong-behaviour-on-screen-rotation">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

metalstrolch:
Same happens on resizing Navit on desktop.

So not directly related to rotation nor related to a special graphics backend / OS build. I think it's only related to resizing, as rotation usually imposes resizing the screen as well.

lains:
This forced return to the main menu is indeed quite annoying on the current internal GUI.
This can happen frequently while manipulating the device, and while entering GPS coords or selecting a street in a town, this makes the user loose time going again through the GUI a second time.

As far a code is concerned, this behaviour comes from function `gui_internal_resize()` in https://github.com/navit-gps/navit/blob/trunk/navit/gui/internal/gui_internal.c
In that function, when resizing the screen (or rotating a portable device), a variable named `changed` will be set to 1, which later on will make `gui_internal_html_main_menu(this);` be called. This is what forces the return to the main menu, and from the code, it seems it is the desired behaviour.
This has the effect of cancelling the current menu (prune) and reload the main menu.

Fixing this is not trivial though...

The behaviour can be slightly improved by reloading the current menu (submenu or main menu) using the following code instead of calling `gui_internal_html_main_menu()`:
```
        gui_internal_prune_menu(this, NULL);
        char *href=g_strdup(this->href);
[...]

PR #682, opened 2018-10-06
Fix:gui/internal:Handling resize/rotation

lains:
This change allows dynamically resizing windows using the internal GUI.
Initially, I guess this GUI was targeted for embedded touchscreen devices, but it is also used for Android (probably until the Qt5 GUI is available).
The issue with the internal GUI is that it does not really handle resizing when navigating inside menus and submenus, it merely moves back to the main menu each time there is a resize (and rotation on Android implies resize which is annoying).
Issue #433  has been open to track this.
This PR aims to solve this resize problem, allowing to continue wherever the user is (inside the menus/submenus) in the menu after a resize/rotation.


lains:
Resize/rotate is now working on HTML menus and on widget-constructed menus as well (these are handled differently).
It still needs work on virtual keyboard resizing (and this is maybe only due to the fact tables resizing is not implemented yet).

hoehnp:
Will it also be possible to get a signal about the rotation for use in changes of the osd layouts?

lains:
@hoehnp: yes, we should be able to get a generic notification about resize (thus also handling rotation for fixed-size displays).
From the test I made, map view resize can also be catched by `gui_internal_resize()` but I think there should be a better place (not related to internal GUI code).
Do you have another example of a signal that is send to OSD items from other events?
How do you catch such signals in OSD items?

lains:
Yes, definitely, if there is a hook to implement, it should be outside of internal GUI, because the way the map is displayed (and thus OSD items on top of it), is really separate from the GUI implemented in my opinion.

lains:
I still have some work to do on handling table and box resize properly, this is impacting virtual keyboard that is currently not rescaled properly.

lains:
I'm almost there but the code around widget packing is quite complex.
However, I am getting small deltas between an initially small rendered window resized to fullscreen, and an originally fullscreen rendered one...
Is there any internal GUI expert that could help out?

pgrandin:
I'm not an expert, but if you share code/screenshots maybe we can figure it out.

lains:
I had to actually write some code to debug the internal GUI boxes (highlight them in red).
If this is of interest, I can share this very small patch to the code or add a preprocessor define to make this accessible to others.
Once I highlight boxes using a red border, and if I initially display the keyboard, then maximize the window, the last (bottom) line of the virtual keyboard _was_ cut as per the screenshot below:
navit_coord_small_resized_maximized_broken

I wrote "_was_" because, in the meantime, I have made progress on this, and it now seems solved by my last commit 91e6a05 (now resizing while displaying the virtual keyboard seems to work).

However, I think handling resize properly would actually need to be developped on a per-GUI menu basis. And it seems to me that I could extend the issue with virtual keyboard to all cases where a box is not aligned to the top left (which was the case for the virtual keyboard that is aligned with the bottom of the window).

Fortunately, this may be caught properly by the resize handler I wrote (for box widgets), because this handler actually recursively repacks all children of a widget.
I have written no handler for tables though, so the result is uncertain.
... and to be honest, I have no good control about the side effects of the code I have added so testers are welcome...

lains:
In any case, it seems that the code in this PR is actually working as expected and solves the issue initially targetted (resize/rotation).
I will give it a try on Android to check how well it performs...
Please let me know what you think about this fix.

hoehnp:
nice work. I think for OSD designers the code with the red lines is for sure a useful patch.

lains:
I'll submit the debugging patch (red lines) in a separate request but this is only active on menus (drawn by the internal GUI), not for OSD items on the map.

lains:
I have tested on Android, and it is actually fixing the rotation also (you don't return to the main menu, or loose anything that you might start typing).
There is one bug that I have identified: if you start in a menu with a given size or orientation, then enter a submenu, then change size or orientation, and finally go back to the previous menu, it is not drawn properly (it has not completely applied the orientation change itself).
I know where it comes from, I will submit a patch when I have some spare time.

lains:
I confirm menus display is broken when going back after rotating, so I still need to work on this.

pgrandin:
Nice work @lains ! This rotation issue was quite annoying. I also agree with @hoehnp the patch to highlight items would definitely be useful.

lains:
@hoehnp and @pgrandin, no problem, I will submit another PR to allow for visually debugging the internal GUI.

lains:
I still need to do some testing on table widget resizing and this PR should then be ready.

lains:
Development is finsihed for this PR. Please test and tell me if there are corner cases of menus in the internal GUI.
Virtual keyboard, top menu bar and standard HTML-like menus should resize/rotate without issues.
The risky parts are the content zone of other menus, for example layout lists, GPS coords input.
Testing can be done directly under Linux by using the internal GUI and resizing the window while on a menu.

> I also agree with @hoehnp the patch to highlight items would definitely be useful.

I have submitted the visual debugging in PR #715

lains:
This patch should handle the issue covered by trac ticket 1149:
http://trac.navit-project.org/ticket/1149

metalstrolch:
Is this ready to be merged? Would be cool if so, because this fixes something really annoying.

lains:
@metalstrolch, tes, it should be ready, I have tested this on Android and Linux builds.
You can already build this code and test it either on the desktop version (resizing the window will show you the result), or by installing it on android which uses the internal GUI. You should get rotation and resize working even while typing some search text for countries, towns or streets.

Please let me know if this works correctly for you.

metalstrolch:
Thank you, good work.

PR #687, opened 2018-10-13
fix: cmake url

hoehnp:
downloading the base map failed with recent changes of the navit server. The PR changes the URL to a working one.


pgrandin:
@hoehnp it looks like you based your PR from the wrong branch, making it difficult to read.

A better approach would be this:
- create a new branch from trunk
- apply your map download change (hint : you can fetch it as a patch here : https://github.com/navit-gps/navit/commit/410a4624bd777307916884385435d794062cf7b5.patch and apply it easily )
- submit a PR.

Once that PR will be merged, we can import trunk into your WIP branches to fix the download there.

PR #688, opened 2018-10-15
fix:i18n:remove extra whitespaces when importing translations

pgrandin:
This ensures that we don't commit translation updates with extra spaces at the end of the line. It should avoid having the CI fail on the sanity checks.


pgrandin:
Tested via https://github.com/navit-gps/navit/pull/686

PR #710, opened 2018-11-01
Refactoring:cmake:simplify version management

bignaux:
Currently , cmake relies on git to provide version variable. This is a bug (and a bad practice) since git data are removed from tarball (and distribution prefers archives to git for lot of good reasons) . So official releases have all ugly version like on ubuntu : 
```
# navit -v
navit 0.5.0+git:0000-

or nixos : 
 navit 0.5.3+git:xdevxgitxnotxfound-
```
So let packager setup  -DGIT_VERSION at cmake invocation (if needed so only for non official tagged releases). I made many improvement in the intention to fix that well, move from a global variable to a build time macro, but i need information about how you want to manage release.

I try to keep the things as they were. NAVIT_VARIANT seems unused, could we remove it ? GIT_VERSION=$(git rev-parse --short HEAD) in the build script is easy to add, how do you want to  ? Can we remove the "+git:" in the NAVIT_VERSION string ?

Proposal : 
* NAVIT_VERSION will have PACKAGE_VERSION+GIT_VERSION(if provided):NAVIT_VARIANT(if provided)
* remove PACKAGE_NAME var.


pgrandin:
> NAVIT_VARIANT seems unused, could we remove it ?

indeed.

> Can we remove the "+git:" in the NAVIT_VERSION string ?

so use directly the sha here? I don't have an objection. what's bothering you with the `git` prefix?

metalstrolch:
The git prefix is misleading. Sailfish packages are for example versioned differently. So the version used for building in GIT_VERSION is actually not pointing to anything git. It's not even a sha1 hash most probably. Same for e.g. ubuntu I think. So it would be better to remove the hardcoded git: from the string.

jkoan:
I would Promote to remove all of the automatic version detection stuff and define the version in the version.h in the future. Of cause this can yield to human error, but if we document it in the Release Process: https://wiki.navit-project.org/index.php/Release_process

Future more i would change only the cmake scripts so that whe Sha sum is passed to the build it will be appended. But dont require this. If a sum is not specified it will be threaded as a dev build (like in IDE), If it is passed and CMAKE_BUILD_TYPE is not given or not Release its also a dev build. But when CMAKE_BUILD_TYPE==Release and sum is not given the build should fail since we don't want to release a build without a sum because this would make Problems in the future, especially debugging and searching for errors.  

I also want to say sorry for my rough tone in the irc chat.

jkoan:
I will merge this now, but we need at least to add the hand over of the Git version to our CI

Issue #584, opened 2018-05-08
Graphics resize image


jandegr:
This line https://github.com/navit-gps/navit/blob/trunk/navit/graphics.c#L721
prevents icon resizing to work properly. 
If a usable size is found it tries to use the next one in the array, often pointing to a nonexisting png.
If I want an icon in 96x96 on Android and I did not provide it but it finds 128x128 it will choose that size to size down, but then ultimatly move to the next one in the array (192) which does not exist and if fails to load the icon.
Since I don't have a clue at all why or for which usecases that worked and only one line of code is involved, I did not make a PR but just report it.
Removing that line in Android makes it work as expected.

jandegr:
Found an even better solution for Android.

pgrandin:
I'm reopening as we might want to make sure that this isn't impacting other platforms as well.

lains:
The code mentionned by @jandegr in May is the following:
https://github.com/navit-gps/navit/blob/f661c8f3f259ed10b6e79477d8125a1f7350d2a2/navit/graphics.c#L721

I think the issue is still there, as I ended up with very poor definitions on a display where 192px wide icon where needed, before https://github.com/navit-gps/navit/pull/716 was merged.

Steps to reproduce is to revert https://github.com/navit-gps/navit/commit/a18841b31e92b5a1dbe7d1025ef12b80a2abc973, and try to dispay the internal GUI on a display that requires 192 or 256px icons (in the main menu for example). It will not use a scaled up version of 128px icon as one would expect, but a scaled up 24px icon, it seems, which makes the icon barely understandable.

hoehnp:
@lains, to me it sounds with the merging of #716  we can close this issue. Is this correct?

PR #716, opened 2018-11-28
circleci:Android: adding high resolution icons and improving rename steps

lains:
In Android build by circleci, icons for 192 and 256px are not built, which, altogether with https://github.com/navit-gps/navit/issues/584 makes icons be displayed with the worst available resolution.
This PR adds high resolution icons during build.
It also fixes successive builds on circleci for android where a rename command applies to all files, including Makefiles, and makes subsequent builds fail.


lains:
@jkoan, no problem.
I am not very used to editing gradle config so I am not sure I can help.

PR #720, opened 2018-12-19
Traffic: Improve handling of situations where no segments are found

mvglasow:
Fixes a crash in the traffic module if no matching segments are found for a traffic condition.

For three-point locations (from–via–to), if no segments are found for from–via, then via–to is used.


mvglasow:
All errors were related to formatting of files not touched by this PR, merging anyway.

Issue #436, opened 2018-04-11
Logo Proposal


redmusic27:
Hello sir, I'm a graphic designer and I would love to contribute to your project with a logo. This is completely free, take it as a contribution. In case you agree I would be presenting sketches and you will be choosing/modifying the aspects you like more so we could work on the perfect logo for your project together.

Let me know what do you think? best regards :)

jkoan:
Hi @redmusic27,
thanke you for your offer, i think we would really like to have a new and up to date logo.
Our current logo looks like this:
![1024px-navit_logo svg](https://user-images.githubusercontent.com/3747958/38617975-a546c1d8-3d98-11e8-8348-6a02f604e1f4.png)

And our app Icon on Android looks like this:
![icon](https://user-images.githubusercontent.com/3747958/38618092-ec9b21f0-3d98-11e8-9752-8454800b20f4.png)

Feel free to ask any question you need to :)

Many thx in advance
Regards Jkoan

redmusic27:
Hello sir @jkoan , I have created a concept logo for you. If you like it, I'll send you all files and pr.
Is there anything you want ?

**Concept**
![concept](https://user-images.githubusercontent.com/37577374/38692611-7b000314-3eae-11e8-8909-2b7dc8600d2a.jpg)
**Horizontal**
![horizontal](https://user-images.githubusercontent.com/37577374/38692618-7ef3df04-3eae-11e8-8adb-866b8c07a56a.jpg)
**Vertical**
![vertical](https://user-images.githubusercontent.com/37577374/38692621-812f262a-3eae-11e8-9916-75ebd411da0e.jpg)

hoehnp:
Thanks for picking up the work. A new logo would be really nice. Also I like the idea of a kind of compass shape. Only criticism from my side is that intuitively I would not recognize center symbol as the characteristic "N" of navit.

zintor:
Nice to see the work on a new logo.
I also like the compass shape, but also agree with hoehnp: The "N" is not easy to recognize. Could you try to highlight the "N" inside the compass?

hoehnp:
Looks great. I like it a lot better.

jkoan:
Hi @redmusic27,
i also really much like the second logo you posted, only one question, is it possible to have the red corner of the icon the same as the blue one?

![navit_new_logo](https://user-images.githubusercontent.com/3747958/38819606-1d90ce52-419c-11e8-9871-39608a98d901.png)

[...]

PR #724, opened 2018-12-30
Update:Navit logo according to proposal in issue #436

lains:
Updating navit logo icon as proposed in issue #436.
Changing GTK GUI to use this logo on window.
Using this logo as the icon for the Action menu in the internal GUI:
navit_new_transp


mvglasow:
Looks good, except that I personally would leave the Actions icon as is. (In fact, the menu/UX is in need of a redesign as well, but that is way out of scope here.)

zintor:
It looks also good to me. I am also with mvglasow, that it sould be a logo replacement and not a redesign of menu/UX.
As pointed, there are more ressources of changing the logo. There is a [wiki page](https://wiki.navit-project.org/index.php/Logo), which we could use to summarize where the logo should be exchanged. Please feel free to use the page. When we do not forget a location.

lains:
I've excluded the update in the internal GUI from this PR.
I agree that a more global redesign of the Internal GUI wuld be required, various actions are using the same icon, and this created some confusion to me when I first used this software.

I also switched to a black logo for the GTK UI for better readability (orange over grey was not very good).

navit_new_logo_gtk

jkoan:
> Looks good, except that I personally would leave the Actions icon as is. (In fact, the menu/UX is in need of a redesign as well, but that is way out of scope here.)

I totally agree with this.

But the GTK UI Icon in black looks good. 
We need to put much more work into UX/UI anyway.

I like the Logo itself (both) but i think we need to redesign even more and meke the look and feel unique across the application.

lains:
I have reverted the icon used for the Action button to the original one [here](https://github.com/navit-gps/navit/pull/724/commits/4b0f5ebd85b16756e4cf78f1acb7d0269ce76f1b), but, yes, I think the design of UI is a wider work.
Can't we first tackle the update of logo/application icon rework, before starting the whole UI rework? This would maybe help to split this huge work into smaller chunks.

mvglasow:
> Can't we first tackle the update of logo/application icon rework, before starting the whole UI rework? This would maybe help to split this huge work into smaller chunks.

Absolutely. Let’s stick with the application logo for this PR, and tackle other UX items separately. I have a few ideas of my own to contribute there. We might want to turn UX redesign into a project to allow us to keep an overview over everything going on, as it involves a bunch of things.

lains:
Hello all,

What shall we do with this PR?
Shall we just merge this new logo, [initially designed by redmusic](https://github.com/navit-gps/navit/issues/436) and reworked+implemented as svg by @zintor?
This PR not only adds the new color svg logo for Navit but also a derived black and white logo for application, and also adds code to make the GTK GUI Windows use that B&W logo as icon.

lains:
Rebased code on trunk.
Also, I intergated to sources, [the svg logo catalog](https://github.com/navit-gps/navit/blob/563b86189811764993ad5add6261610e3e5e520c/logo_catalog.svg) that was last posted in https://github.com/navit-gps/navit/issues/436#issuecomment-450587935

lains:
Anyone against merging this? If not, I will rebase and merge (probably around Wed 18th).

mvglasow:
Looks good to me.

PR #730, opened 2019-01-11
Fix:circleci:Temporarily disable Coverity as it is currently down

mvglasow:
As Coverity is currently unavailable (see https://community.synopsys.com/s/article/Coverity-Scan-Update), this PR disables the Coverity scan to allow CI to complete successfully.

Right now, automatic merges of trunk into master are impossible if the Coverity scan (i.e. generating the scan and uploading the results) fails, as this will result in CI failing. Conversely, as I understand it, scan results do not affect the outcome of CI. That is:

* If the scan cannot be executed or uploaded, CI fails and no merges into master take place. (Which is the case if Coverity is down.)
* If the scan runs successfully and the upload goes through, but several bugs are found in the code, that doesn’t prevent CI from reporting success and merging the results.

On the long run, we should IMHO choose one of the two:
* Ensure CI fails if the scan finds anything suspicious, or
* Try scanning and uploading the results, possibly generating a warning if that fails, but still report success for CI.

If we go for the first, we might want some kind of override, i.e. being able to manually tell CI to ignore what we deem a false positive.

For the time being, this PR disables Coverity altogether, to allow CI to complete until Coverity is back online.

### DIFF
--- a/scripts/build_linux.sh
+++ b/scripts/build_linux.sh
@@ -9,38 +9,38 @@ cmake_opts="-Dgraphics/qt_qpainter:BOOL=FALSE -Dgui/qml:BOOL=FALSE -DSVG2PNG:BOO
 [ -d $BUILD_PATH ] || mkdir -p $BUILD_PATH
 pushd $BUILD_PATH
 
+# Build everything
+cmake ${cmake_opts} ../
+make -j $(nproc --all)
+make package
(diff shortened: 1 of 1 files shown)


mvglasow:
PS: refer to today’s IRC logs for the discussion with jkoan.

Issue #733, opened 2019-01-20
Redesign of POI icons on map


zintor:
As there are a bunch of map POI icons from different sources with various styles and without an cooperate design, it would be nice to have a POI Icon package for Navit.

So I used the package of https://github.com/gmgeo/osmic to create some variations of icons:
![fuel](https://user-images.githubusercontent.com/22443449/51445269-87aed480-1d03-11e9-996a-43cfdc75bc7f.png)
I know, that it is a lot of work, but we could start with this package of osmic to get the most relevant icons.

The menu use the same icons. So there are some more work.

Some Screenshotd in SFOS 3 Emulator with a "SFOS"-style 96x96px:
![fuel_marker_pin1_96_96](https://user-images.githubusercontent.com/22443449/51445279-95fcf080-1d03-11e9-89d6-44ff26d9cfc8.PNG)
![fuel_marker_pin1_96_96_3d](https://user-images.githubusercontent.com/22443449/51445280-95fcf080-1d03-11e9-9773-00d7ed98cbc5.PNG)
![fuel_marker_pin2_96_96](https://user-images.githubusercontent.com/22443449/51445281-95fcf080-1d03-11e9-8cac-9fd25993e148.PNG)
![fuel_marker_pin2_96_96_3d](https://user-images.githubusercontent.com/22443449/51445282-96958700-1d03-11e9-91fa-c84c1564cded.PNG)
![fuel_marker_rect_96_96_sfos_2d](https://user-images.githubusercontent.com/22443449/51445283-96958700-1d03-11e9-9f29-33b07c4852c1.PNG)
![fuel_marker_rect_96_96_sfos_3d](https://user-images.githubusercontent.com/22443449/51445284-96958700-1d03-11e9-9e02-a829d581d737.PNG)

mvglasow:
Looks great and is indeed much needed. While we’re at it, we might also want to think of a redesign of the default OSD layout; my suggestion would be this one: https://wiki.navit-project.org/index.php/File:Screenshot_Navit_20180831-141920.png

hoehnp:
I am also currently working on a redesign. This should then include an auxmap with a simplified layout to avoid a map too much with POIs. However, I would guess this code still take a while as it also requires some changes to the code. What do you think?

zintor:
Well, the layout is not my Intention to bring forward. The layout section on the [wiki](https://wiki.navit-project.org/index.php/Layout) shows good examples. Maybe the Mapnik layout could be the stanard layout.
[...]

Issue #471, opened 2007-11-06
Roundabout detection


aerostitch:
will be really needed soon :\)

Reported by: kazer_

Original Ticket: "navit/feature-requests/6":https://sourceforge.net/p/navit/feature-requests/6

PR #737, opened 2019-01-25
New default OSD layout

mvglasow:
Inspired by #733, I am proposing a gentle redesign of the default OSD layout. Its main features are:

* Consistent design for all OSD touch controls
* Default transparency changed to 0x0a for better legibility, while still allowing the map to shine through
* Touch controls with white icons on semitransparent black background, to improve visibility on the map background
* Variable-width items (street names) occupy the full screen width
* OSD items are hidden when they are not meaningful, and are grouped so that blocks of items are shown or hidden together
* Items related to the current position are near the bottom of the screen, those related to routing are near the top of the screen (in a fashion similar to traffic signs).
* Designed to work with any screen size (minimum 240×240 LDPI)

Currently this is implemented for Android and iOS (which already shared a minimum OSD), as well as for WinCE and Openmoko, but I am proposing to extend this to all mobile platforms.

osd redesign 1

osd redesign 2

Different use cases (no location, location but not routing, map not following vehicle, calculating route, routing):

navit-1-nolocation navit-2-location-notrouting navit-3-notfollowing navit-4-calculating navit-5-routing


hoehnp:
@mvglasow: thanks for the work. It is a very nice design. Does the design also work well in landscape mode? I would also suggest one more improvement which i worked on in the past but didn't complete yet. I worked on implementing opening an auxmap with a map layout  excluding POIs to give a better guidance in crowded city areas. In particular this covers a large number of parking POIs.

mvglasow:
@hoehnp that’s why I designed it for a 240×240 screen—I assumed the minimum available to be a QVGA screen (320×240) and ensured nothing takes up more than 240 pixels in either direction. (On platforms that support higher densities, these sizes go for LDPI, thus the screen needs to be 320 dp×320 dp.) All controls are aligned to the closest screen corner; the bands along the top and bottom take up the entire screen width. So yes, the layout should work for portrait and landscape mode alike.

As for your auxmap solution, do you have any screenshots to share? Might be a separate pull request, though.

mvglasow:
…and here’s a screenshot in landscape mode:

navit-6-landscape

hoehnp:
@mvglasow: I don't have a screenshot of the complete solution yet. Otherwise I had already integrated it ;-) but I can show you a bad picture of what i have in mind:
img_20171101_165150

hoehnp:
The code part I am missing is to give auxmap the option to use a different layout compared to the main screen.

mvglasow:
@metalstrolch @aerostitch @jandegr I see you have created platform-specific OSD configurations for Tomtom and Sailfish respectively, would you be open to switching to a common default layout?

navit/osd/core/osd_core.c
```diff
@@ -2291,10 +2292,10 @@ static struct osd_priv *osd_nav_toggle_announcer_new(struct navit *nav, struct o
         char **array;
         we = file_wordexp_new(attr->u.str);
         array = file_wordexp_get_array(we);
-        this->icon_src = g_strdup(array[0]);
+        this->icon_src = graphics_icon_path(array[0]);
         file_wordexp_destroy(we);
     } else
-        this->icon_src = graphics_icon_path("%s_32.xpm");
+        this->icon_src = graphics_icon_path("%s_48_48.png");
```
metalstrolch: Is there a better way than had coding .png with fixed size? Sailfish aims to be .svg only.

navit/xslt/osd_minimum.xslt
```diff
@@ -5,27 +5,35 @@
...
 
-      <osd type="compass" enabled="yes" x="{round(-60*number($OSD_SIZE))}" y="{round(-80*number($OSD_SIZE))}" w="{round(60*number($OSD_SIZE))}" h="{round(80*number($OSD_SIZE))}" font_size="{round(200*number($OSD_SIZE))}" enable_expression="vehicle.position_valid"/>
+      <osd type="text" label="${{navigation.item[1].name}}" x="0" y="0" w="100%" h="{round(14*number($OSD_SIZE))}" font_size="{round(200*number($OSD_SIZE))}" enable_expression="navigation.nav_status&gt;=3"/>
       <xsl:text>&#x0A;        </xsl:text>
-      <osd type="text" label="${{navigation.item.destination_length[named]}}\n${{navigation.item.destination_time[arrival]}}" x="{round(-60*number($OSD_SIZE))}" y="0" w="{round(60*number($OSD_SIZE))}" h="{round(40*number($OSD_SIZE))}" font_size="{round(200*number($OSD_SIZE))}" enable_expression="navigation.nav_status>=3"/>
+      <osd type="compass" enabled="yes" x="{round(-60*number($OSD_SIZE))}" y="{round(-94*number($OSD_SIZE))}" w="{round(60*number($OSD_SIZE))}" h="{round(80*number($OSD_SIZE))}" font_size="{round(200*number($OSD_SIZE))}" enable_expression="vehicle.position_valid"/>
```
metalstrolch: Again, is there a better way. Need to think how to convert this to .svg for Sailfish using xslt. At least it features "w" and "h", which is mandatory for pure .svg

mvglasow: Probably not something that will work across all platforms, as not every platform supports SVG. The best bet is probably to use XSLT to convert this to SVG on platforms which support it. Question—will SVGs be scaled to match the size specified? PNGs appear to get cropped or padded, at least for `toggle_announcer`.

navit/osd/core/osd_core.c
mvglasow: This is just the fallback in case `icon_src` is not specified. The XSLT does specify an icon, though.

metalstrolch:
Simply copied over openmoko.xslt to sailfsh_osd.xslt to enable the new OSD layout, but fails badly on x86_64 configured the same way as sailfish is. Configure using the following cmake line:
``
cmake -DCMAKE_INSTALL_PREFIX:PATH=/tmp/navit -DPACKAGE:STRING=harbour-navit -DNAVIT_BINARY:STRING=harbour-navit -DSHARE_DIR:PATH=share/harbour-navit -DLOCALE_DIR:PATH=share/harbour-navit/locale -DIMAGE_DIR:PATH=share/harbour-navit/icons -DLIB_DIR:PATH=share/harbour-navit/lib -DBUILD_MAPTOOL:BOOL=FALSE -Dfont/freetype:BOOL=FALSE -Dbinding/dbus:BOOL=FALSE -Dbinding/python:BOOL=FALSE -Dgraphics/gtk_drawing_area:BOOL=FALSE -Dgraphics/null:BOOL=FALSE -Dgraphics/opengl:BOOL=FALSE -Dgraphics/sdl:BOOL=FALSE -Dspeech/dbus:BOOL=FALSE -Dvehicle/gpsd:BOOL=FALSE -Dvehicle/gpsd_dbus:BOOL=FALSE -DUSE_PLUGINS=n -DUSE_QWIDGET:BOOL=FALSE -DXSLTS:STRING="sailfish_disable;sailfish_qt5;sailfish_osd;sailfish_cursor;sailfish_mapset;sailfish_svg;sailfish_gui" ../navit/
``
Fails badly not finding some callbacks and segfaulting. What's the cause? Any ideas?
``
error:navit:command_register_callbacks:could not resolve object in cs=0x55e87e857700:
error:navit:command_register_callbacks:	navigation.nav_status>=3
error:navit:command_register_callbacks:	           ^
error:navit:command_register_callbacks:could not resolve object in cs=0x55e87e857ff0:
error:navit:command_register_callbacks:	vehicle.position_valid
error:navit:command_register_callbacks:	^
error:navit:command_register_callbacks:could not resolve object in cs=0x55e87e8587b0:
error:navit:command_register_callbacks:	navigation.nav_status>=3
error:navit:command_register_callbacks:	           ^
error:navit:command_register_callbacks:could not resolve object in cs=0x55e87e8590f0:
error:navit:command_register_callbacks:	navigation.nav_status==-1 || navigation.nav_status==1 || navigation.nav_status==2
error:navit:command_register_callbacks:	           ^
error:navit:command_register_callbacks:could not resolve object in cs=0x55e87e859910:
error:navit:command_register_callbacks:	navigation.nav_status>=3
error:navit:command_register_callbacks:	           ^
error:navit:command_register_callbacks:could not resolve object in cs=0x55e87e85a130:
error:navit:command_register_callbacks:	navigation.nav_status>=3
error:navit:command_register_callbacks:	           ^
error:navit:command_register_callbacks:could not resolve object in cs=0x55e87e85cd80:
error:navit:command_register_callbacks:	navigation.nav_status>=3
error:navit:command_register_callbacks:	           ^
error:navit:command_register_callbacks:could not resolve object in cs=0x55e87e85d640:
error:navit:command_register_callbacks:	vehicle.position_valid
error:navit:command_register_callbacks:	^
error:navit:command_register_callbacks:could not resolve object in cs=0x55e87e8608d0:
error:navit:command_register_callbacks:	follow>1
error:navit:command_register_callbacks:	^
Segmentation fault
``

mvglasow:
About proximity of buttons, I have used a similar setup (with the same button size and distance) on Android for years now and personally did not have issues with the size.

The buttons are designed to be 48 px at 100–120 dpi (which includes most PC screens and WinCE devices), with 8 px spacing. For Android they are scaled according to pixel density and end up being 64 dip (display-independent pixels) with ~10 dip spacing, which is what Google recommends for UI design.

For Openmoko I have assumed the pixel density to be 240 dpi (taking the Freerunner screen as a basis) and scaled things accordingly, i.e. doubling pixel dimensions.

The Jolla with its 4.5", 960 × 540 px display would be quite close to 240 dpi, therefore things should be OK here if we scale the items by a factor of 2 (which would make the buttons 96 px, with 16 px spacing). Though I am not particularly familiar with the Sailfish platform, neither the devices available (and how much pixel density varies) nor with how the platform addresses pixel density. I understand SVG solves some of the issues, but how would I specify that I want a particular particular item some 0.4 inches wide (0.5 still being OK but not 0.2), regardless of the screen’s pixel density?

mvglasow:
As for the error messages, I am seeing the ones about irresolvable objects as well. Seems to be a pre-existing bug in Navit that some objects are not yet visible when the OSD starts up. Other than generating a few error messages on startup, it is harmless and the OSD will work properly once everything is loaded.

I have successfully tested this OSD layout on Linux and Android (the screenshots were made on Linux with the Android LDPI OSD copied in). You will also see that code changes are minimal and mostly limited to constants; any bugs are probably long-standing ones that are coming to light only now.

I never had a segfault, though. Can you get me a stack trace of where the segfault occurs? Then I can look into it.

navit/xslt/osd_minimum.xslt
metalstrolch: About file types: Doesn't the graphics subsystem try to "guess" the filename anyway? I remember something in navit/graphics.c (image_new_helper). So it should be ok to give "w" and "h" properly and then remove the file extension and the size from the file name. Graphics subsystem will pick what's there, starting with %name_%width_%height.png Next try would be %name.svg and so fourth. That's what I think we should use for everybody to be fine.

About scaling: Unfortunately this is up to the graphics plugin to handle correctly. And obviously not all do right. Qt5 does. If you see that cropping, then the graphics used by you does not. But bitmap scaling should only be the last resort. If using bitmaps, then they should be pre scaled as you do as well.

About svg scaling: Svg's as all other file formats do include a size hint, So if no size is explicitly given, at least QT5 uses the size of their size hint to render them to bitmap. Unfortunately this size hint is often bogus, as nobody cares for that on svg usually. So one has to provide the size to the renderer explicitly. That's why using "w" and "h" properties is so important and should be preferred over implicitly setting the size by the bitmap icon size. The one svg file can be rendered to all required sizes on the fly.

mvglasow:
Good to hear the OSD seems to work.

Re saxon: build chain fragmentation is something that should be avoided as much as possible—once a uniform build chain is in place, development becomes so much easier. Is the CI environment also affected by this? Would installing saxon be an option, or is there anything preventing it?

If not, we would have to explore the path of porting the XSLTs affected to 1.0, though I am by no means an expert on XSLT and would need some help with that.

About image file names and size: I will look into specifying the OSDs as you said—give `w` and `h` attributes and strip dimensions and format from the file name. If that works, I’ll look into the same for `toggle_announcer`, for which the logic is slightly different (due to the need to swap bitmaps to reflect current state). On Sailfish, how would I specify dimensions—physical pixels on a 240dpi screen? Or something else?

metalstrolch:
Adding saxon as a build requirement is quite hard, as it is a java tool and would pull in entire java into the build system. Think that isn't even available as a official package on sailfish. Don't know if CI is affected as I don't know what package feed is used there. Xsltproc has always been valid option in navit build, as is windows "convert.exe" which is most probably even more restrictive. Albeit I don't know if it is still used. So currently the xslt files doesn't fit to that.

For me the best way would be to port (most) xslt files to 1.0, but I'm no expert here as well.

Dimensions on sailfish are given the same way as on every navit platform, in percent or pixels. Would be cool if navit could honor DPI values and then understand millimeters. Would solve the "osd for dpi" thing for good if one could say the icon should be 6mm x 6mm. But this would be a quite big change to navit's OSD and other modules.

Now I have the problem that the OSD doesn't really work on qt5 graphics, but I more and more think this is an old qt5 graphics bug since I remember disappearing overlays and such from development.
Seems It's time to fix qt5 here.

mvglasow:
Now that I am looking at it again, I see that osd_minimum.xslt is version 1.0, and this is the only XSLT needed for the OSD layout. So we should be fine here.

metalstrolch:
OK, I made some progress here. I fixed graphics_qt5 to finally handle overlays properly. See #741 .
By porting openmoko.xslt to sailfish_osd.xslt and converting osd_enable_zoom.xslt to version 1.1 I'm finally able to build this for qt5 with svg only. Now I've 2 problems left
:
1. osd_enable_zoom.xslt enables the zoom buttons out of navit_shipped.xml as well. How did you fix this for your build?

2. The enable_expression's don't work. So I always see all elements, causing some elements to overlap. Try to investigate why. I get plenty of error messages that could pint to that problem.

metalstrolch:
Btw. saxon is a problem, as sailfish doesn't have packages for it. So you could not simply install it.

mvglasow:
Re saxon, as the only xslt involved is version 1.0, this should no longer be an issue.

You might not need the full openmoko.xslt, it should be sufficient to include osd_minimum.xslt and make sure the variables it uses are defined. I’ll come up with something.

As for enable_expressions, just a wild guess (as I don’t know the inner workings), but this might have to do with the graphics driver. Things to try:
* When you touch an OSD button which should be disabled, does it trigger the desired action? (You might want to build a custom button for that, with an action that is noticeable on the screen (such as zoom in/out) and an enable_expression that should evaluate to false.)
* As I understand, Sailfish supports multiple graphics drivers. Do they all exhibit the same problem, or is it limited only to some?

mvglasow:
3c51ff8 applies the OSD changes to Sailfish. Icon sizes are adapted to match those in sailfish_gui.xslt (buttons follow medium icon size, route status/next turn follows big icon size).

metalstrolch:
You must enable the zoom buttons for Sailfish/qt5, as zoom gestures are not supported(yet).
The osd_enable_zoom.xslt can be ported to 1.1 by just replacing the version number. But it enables buttons brought from navit_shipped.xml as well.

Sailfish uses always graphics_qt5, but this can use QWidgets or QML for drawing. Both modes share most of the code and (mis)behave the same with your layout. The default btw. is to use QML, as QWidgets is disallowed on Sailfish store. After all I was the one who wrote that. But you're right, it is graphics dependent, albeit I don't know why and how. But It's not drawing. I think this is somehow a race condition.

The Problem:
The buttons always work, albeit funny things happen if you press overlapping ones.
I think the cause is indicated by:
``error:navit:osd_std_config:Error in saved command: not_ready, item=0x5624516b2a00.
error:navit:osd_std_config:Error in saved command: not_ready, item=0x562451681cf0.
error:navit:osd_std_config:Error in saved command: not_ready, item=0x562451680b60.
error:navit:osd_std_config:Error in saved command: not_ready, item=0x56245168cf30.
error:navit:osd_std_config:Error in saved command: not_ready, item=0x56245168c570.
error:navit:osd_std_config:Error in saved command: not_ready, item=0x56245168bbd0.
error:navit:osd_std_config:Error in saved command: not_ready, item=0x562451695680.
error:navit:osd_std_config:Error in saved command: not_ready, item=0x562451693f00.``

metalstrolch:
Oops, wrong button. of course stays open...

mvglasow:
Added osd_enable_zoom. As for my build, I did not touch that part, other than changing the button images and moving the buttons to a new location.

metalstrolch:
OK, were getting closer. Fixed graphics_qt5 so the enable_expressions are evaluated right. Again see #471

Still the additional set of zoom buttons. present in the sailfish rpm out of CI artefacts as well.

Remaining questions valid on qt5 and gtk:
1. What should the "follow" button do? I didn't manage to make it appear. I think it should react on "follow_cursor" instead of "follow". Additionally the "set_center_cursor()" call needs to be guarded or fixed, as it segfaulted on my box having no active vehicle. I use "demo" vehicle which must be activated manually before it has a "fix". Maybe I have a look in set_center_cursor() next.

2. I would remove the "cmd_interface" line as this makes everything behave really strange. At least for me this doesn't "feel" right when using navit. Maybe it's because I don't like the tilted view (and autozoom, but that's a different story)

3. When should the compass appear? At least with the "demo" vehicle it did never.

mvglasow:
The follow button should appear when dragging the map; tapping it should reattach the map to the vehicle position (so it will follow the vehicle).

The cmd_interface is something I took over from the existing config unmodified, but I have disabled it in the setup I use. I will look into it.

The compass should appear whenever the vehicle has a valid position. On Linux (the setup on which I generated the screenshots) with the demo vehicle this works as soon as we have a position and a destination, and the compass remains when navigation is stopped. I use the same definition on Android, and that works as well. Might be another graphics issue, or (less likely) something about vehicle position or related callbacks on Sailfish.

mvglasow:
The cmd_interface sets the pitch (enabling/disabling 3D view) and orientation when autozoom is active and the map is set to follow the vehicle. This was introduced in #306. I’ve disabled it in my personal config—maybe it could be improved by adding some setting that toggles between 2D and automatic 2D/3D switching.

metalstrolch:
Now I got it working. The trick is that the active="1" property must be set on vehicle. If I add this to the demo vehicle it works. If the property is not set and the vehicle is enabled using the menu, the OSD fails.
Can somebody confirm this?
If true, how can we fix that.

Still I think the automatic pitch / orientation setting is plain stupid and renders the menu items about the same useless. But if the majority thinks this is an improvement I will only disable this on my personal device and keep it for the default sailfish package.

mvglasow:
I also have doubts if the cmd_interface is the best way of doing it; a proper option would probably be better. Anyway I have disabled it in my personal config as I do not use 3D at all. But I guess this feature should be discussed in a separate ticket.

About the vehicle, IIRC the vehicle must be both enabled and active in order for Navit to use its position. The internal menu has a menu item for points, “Set as position (and deactivate vehicle)”, which will make Navit assume that position as the current one and ignore position updates from the vehicle, until the vehicle is set as active again. With the vehicle disabled, reattaching the map view to the vehicle is not really meaningful.

With the demo vehicle disabled, if you set a position and a destination, I suspect Navit would not even start routing or at least would not update the map view to move along the route.

metalstrolch:
If using the demo vehicle without active="1" , you can manually enable it via the vehicle menu. The same way you can re enable your vehicle after "Set as position (and deactivate vehicle)". That's how I used demo vehicle up to now, as navit_shipped.xml does not contain active="1".
Need to try what happens to the OSD if "Set as position (and deactivate vehicle)" and later on activate it again with a "real" vehicle.
Right about the pitch discussion. Should be done elsewhere. Personally I will disable that as well. Don't like 3d view at all.

mvglasow:
> demo vehicle [...] navit_shipped.xml does not contain active="1".

Kind of makes sense, as it is seldom needed for production use. I have seen some commercial sat-navs which offer a simulation of the route, but cannot think of a real use case other than presentation to a potential buyer—or testing the product during development. Thus, making it the default would make sense only for development, test or demonstration setups, not for production ones.

Anyways, what is left to be done before this OSD on Sailfish is ready for prime time?

metalstrolch:
I've still the issue with the double zoom buttons, but I think here we have a misunderstanding.
Should the zoom buttons of the OSD layout added by osd_minimum.xslt be visible by default or not.

Because including osd_minimum.xslt and osd_enable_zoom.xslt in the same.xslt file and therefore the same xslt run doesn't make sense. By including both rules are applied together to the basis, so osd_enable_zoom.xslt will never apply to the zoom buttons brought by osd_minimum.xslt. The nodes do not yet exist when the enable rules are applied.

I think we must: remove the include of osd_enable_zoom.xslt from all .xslt's that include osd_minimum.xslt as well (sailfish, openmoko, windows). The zoom buttons are currently always available from osd_minimal.xslt.

If we want to control the zoom button presence from osd_minimum.xslt, the I would simply do it using a xslt variable like done for use_overlay.

mvglasow:
I get what you mean now, will look into it

mvglasow:
OK, I’ve removed osd_enable_zoom.xslt from anything that includes osd_minimum.xslt. That will make them show on any platform that uses that OSD. This maintains existing behavior on Android and iOS. On WinCE and Openmoko the zoom buttons are needed as devices are not guaranteed to have other means of zooming.

I understand we need the buttons on Sailfish as well, correct?

On the long run, it might help to have a property like `has_hardware_zoom` which could then be used in an `enable_expression`, as this can vary between devices of even the same platform, and may even change at run time (imagine unplugging a wheel mouse, and falling back to the two-button no-scroll touchpad). Then again, despite having pinch-to-zoom on Android, I prefer having zoom in/out buttons. Again, different issue for a different PR.

metalstrolch:
Yes we need zoom buttons on sailfish, as qt5 is not (yet) enabled for zoom touch gestures.
But, as is, the zoom buttons are always enabled! On any platform I think.

osd_enable_zoom.xslt is only good for enabling the zoom buttons already present in navit_shipped.xml that show up on the lower right and left corners of the screen. If done next to osd_minimal.xslt., two pairs of buttons appear.

Have a look on the generated navit.xml. You see two sets of zoom buttons, but the one coming from navit_shipped.xml are usually disabled.

But yes, the "has_hardware_zoom" is indeed a good idea.

mvglasow:
Great! Now I’d just like an opinion from the Tomtom folks…

mvglasow:
@jkoan could you maybe take a look at the latest CircleCI artifacts? Seems the XML config is missing entirely from it, and has been for a while already.

An older Tomtom file I found had a config file named `tomtom480.xml` in `/navit/share`, the CircleCI artifacts do not. If I SSH into the machine, I can find `./project/build/tomtom480.xml` but none of the XSLT transforms seem to have been applied. In both cases, that goes for the current build, the last build before I changed the Tomtom XSLT, and the latest build on master.

While I am not an expert on this platform, to me that looks like the Tomtom build is broken…

lains:
@mvglasow, this OSD looks nice to me.
If the XML config parsing is broken on tomtom, this is maybe because of my previous work in PR https://github.com/navit-gps/navit/pull/559. I don't know if anyone managed to test it on tomtom builds before the code was merged.

mvglasow:
@lains any ideas where to look? A lot of things changed in this PR…

mvglasow:
@lains it looks like Tomtom had the same issues before. I just created a branch based on b077321 (just had to doctor the CircleCI config to force the Tomtom builds to run) and ran it through CircleCI. Same thing, no XML file in `/navit/share`. If that is an error (which I can only suspect), it has been there before.

mvglasow:
I went back in CircleCI history as far as it would go (i.e. September 5, 2017) and already back then, the XML file was absent from the Tomtom builds. In other words, this file has been absent from the packages since we moved to CircleCI V2, that is, for almost a year and a half.

@jkoan @lains can you shed some light on this? Is this an error at all? If not, where does the Tomtom build get its config from? If yes, any idea what might be causing it and how to fix it?

mvglasow:
Yes, I’d seen that line already. It explains why the XML file does not show up in the output, but not why the XSLs are not getting applied.

The wiki links to an archive which has the XML file (https://wiki.navit-project.org/index.php/TomTom#Using_pre-compiled_binaries), dated December 3, 2016. As of September 5, 2017, the file has been missing; I’ll examine the commits in that timeframe. History is a bit hard to trace for the build scripts as they were moved to a different directory after that date.

mvglasow:
@jkoan Thanks for looking into this, very much appreciated!

lains:
> the zoom in and zoom out buttons are quite close to each other. Quite a pain to use. Even wrong with the toggle auto zoom next to them

I have compiled and tried on Android (hdpi display).
This OSD looks nice and modern to me. Congratulations.
I tend to agree with what @metalstrolch wrote: I often miss the zoom in, out and manual icons when I click on them on that specific phone.

mvglasow:
@lains the buttoms use the medium icon size for the platform, which is also used for the title bar of the internal GUI menu. There’s a bit of variation in screen densities; particularly when a device is in the border area between two density bins (say, almost halfway between HDPI and XHDPI), but reporting the lower of the two densities, buttons may end up being a bit smaller. If that is an issue, we would need to increase the medium icon size.

mvglasow:
Any news on the Tomtom build?

mvglasow:
See #751 for Tomtom build issues; I will merge this one now. If the OSD does not work properly on Tomtom after fixing the build, this will need to be looked at separately.

PR #741, opened 2019-01-29
Fix:graphics_qt5:fix overlay handling

metalstrolch:
Do not clear overlay contents if resize is called for them, but the actual geometry doesn't change.
This happens on resize if the overlays are repositioned. Now OSD's do not loose elements any more.

Apply astyle to comply with coding style and to tame CI


metalstrolch:
Merget because required for properly testing new osd.

PR #745, opened 2019-02-04
Make Navit work in split screen mode

mvglasow:
According to blame, this particular chunk of code was added by Rikky (who seems to be no longer active) in 2011, and seemed to be a precautionary measure to prevent unclean shutdowns.

However, this prevented Navit from resizing properly, as resizing destroys the activity and then recreates it—the latter being prevented by Navit exiting. This explains why Navit would not work in multi-window mode. (Tested on Anbox, where this change will make Navit resize properly rather than exiting on any attempt to change the window size.)

Most likely we do not need that precautionary measure any longer: 85b9d5da already replaced the call to `finish()` in the `Navit#exit()` method with a call to `NavitDestroy()` (to prevent unclean shutdowns), and any non-Android quit operation should cause the process to exit anyway.

Still I would appreciate it if someone familiar with the Android port could take a look at this, as I might have overlooked something.

### DIFF
--- a/navit/android/src/org/navitproject/navit/Navit.java
+++ b/navit/android/src/org/navitproject/navit/Navit.java
@@ -20,6 +20,7 @@
 package org.navitproject.navit;
 
 import android.Manifest;
+import android.annotation.TargetApi;
 import android.app.Activity;
 import android.app.AlertDialog;
 import android.app.Dialog;
@@ -75,11 +76,6 @@ public class Navit extends Activity {
     private NavitActivityResult[]      ActivityResults;
     public static InputMethodManager   mgr                             = null;
     public static DisplayMetrics       metrics                         = null;
-    public static int                  status_bar_height               = 0;
-    private static int                 action_bar_default_height       = 0;
-    public static int                  navigation_bar_height           = 0;
-    public static int                  navigation_bar_height_landscape = 0;
-    public static int                  navigation_bar_width            = 0;
     public static Boolean              show_soft_keyboard              = false;
     public static Boolean              show_soft_keyboard_now_showing  = false;
     public static long                 last_pressed_menu_key           = 0L;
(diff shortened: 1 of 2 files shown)


mvglasow:
After testing in true split-screen mode, it seems that the change per se works. However, the transparent system bars (in fact, the way Navit handles them) needs to be adapted to work with split screen properly.

mvglasow:
Reworked the system bars, which should now fix a couple of bugs (requires API 23+, though; earlier APIs still retain the previous behavior).

There’s one bug remaining: the navigation bar is not transparent in split screen, but when it borders on the Navit window, the system reports an inset on that edge, equal to the height/width of the navigation bar. As a consequence, Navit will prevent OSD items from occupying that area. Dunno why this happens (or even if transparent navigation bars are supposed to work in split screen).

Help is appreciated—either to make the navigation bar transparent in split screen, or reliably tell whether or not it is obscuring parts of the window. Right now window insets will always claim it is on top of the Navit view even when it isn’t…

mvglasow:
This is getting weirder and weirder. Spoiler: there seems to be a solution, albeit a hacky one.

Apparently transparency in split screen mode is supported for the notification bar but not for the navigation bar, and no content will be drawn behind the latter. Nonetheless, querying the root window insets always reports an inset for the navigation bar, which would result in unnecessary padding along it.

This whole mechanism does not seem to be the preferred way of querying window insets anyway, and is only supported on API 23+ (whereas window insets first appeared in API 19). Apparently, the preferred way is to override `onApplyWindowInsets()` and examine the `insets` argument. (This method can also alter the insets, specifying how much of them it will consume with padding, resulting the altered values to be passed down to child views.)

Sound cool? Not quite so, for that mechanism has a bug of its own: the top inset is somewhere between 24 and 64 dip too big. It will report an inset for a notification bar when there is none (as on Anbox, where each app runs in a window), or when it is not obstructing any part of the app (e.g. in split-screen mode when the app is being displayed in the lower half). Or it will report an inset much bigger than the area occupied by notification bar. Or a combination of both.

We thus have two ways of querying insets, each with its own bugs:
* Querying root window insets will give an incorrect inset for the navigation bar.
* `onApplyWindowInsets()` will give an incorrect top inset.

As long as the navigation bar never gets displayed at the top, these two bugs are complementary. As a workaround, we determine the top inset using `onApplyWindowInsets()` and the others by querying root window insets. This seems to work so far, will conduct more tests to be sure and then merge the changes.

mvglasow:
Tests look good, seems like we’ve finally nailed window insets. I am still getting a warning that the `LocationListener` is never unregistered; will look into that.

Also, we are currently discarding and recreating the activity whenever it changes (rotation, resize, split screen toggle), which is very expensive. Might want to look into keeping the activity, as we do on other platforms.

hoehnp:
do you have a screenshot of t he split screen?

mvglasow:
btw, the `LocationListener` leak has been there all the time, not something introduced by the changes in this PR. Most devices don’t log it but one of the emulators I tested on did, which likely is why it went unnoticed. Also it is not critical, as we would release the `LocationListener` only upon closing Navit (and at that point the OS does that for us anyway).

Recreating the activity is also nothing new, except that split-screen mode exacerbates the issues (especially with a lot of cached traffic reports). I would therefore propose merging this as-is, and look into reusing the activity in a second step.

mvglasow:
Taking a closer look, I see the activity is set to be preserved, and adding extra debug output reveals that Android appears to honor that. However, I see that cached traffic messages are getting re-read when the configuration changes. Possibly the activity unnecessarily goes through startup stuff on a config change.

mvglasow:
Actually… not quite. `onConfiguration()` gets called twice, but almost immediately after, I see a call to `onDestroy()`, followed by a call to `onCreate()` during which `getChangingConfigurations()` returns 0.

That is, the activity is notified of the configuration change, but then gets destroyed and recreated for no apparent configuration change.

That leaves us with two options: either dig deeper and figure out why this activity is getting destroyed, or simply let it happen and ensure we preserve its state across instances. The latter would also help should there be some configuration changes which we can’t handle properly.

mvglasow:
I had another look at the initialization stuff on Android. On Anbox I see the main activity gets destroyed and recreated already during startup, so I am inclined to just accept this is going to happen (basically, this seems to be what the Android docs say: be prepared for your activity to get destroyed and recreated any moment).

Most `Navit` members are already static and would survive recreation; those few which are not are related to the activity itself, and it makes sense to reinitialize them. Most of that happens in `Navit#onCreate()`; the method body contains only calls which are either required after an instance recreation (as they requery parameters which may have changed) or are very quick.

Ultimately, the method calls through to `Navit#NavitMain()`, a native method. As above, most calls in the method body are fairly quick, the only exception being the call to `main_real()`.

`main_real()` is the all-platform entry point to Navit. Among others, it initializes all plugins and eventually starts the main event loop. This is the stuff that takes long and shouldn’t be necessary twice on Android.

I then did the following:

* initialize the static member `Navit.navit` to null
* in `Navit#onCreate()`, check if `navit` is null: if yes, we are starting from scratch; else we are just recreating the activity.
* add a boolean argument to `Navit#NavitMain()` which is true on launch, false when called for a recreated activity
* in `Navit#NavitMain()`, skip the call to `main_real()` unless we are starting from scratch
* in the same place, call `event_main_loop_run()` instead of `main_real()` for activity recreation
* in `Navit#onActivityResult()`, add a check to ensure `ActivityResult` is non-null before invoking its event handler

Alas, this does not work. I just get a black window, but I don’t know what I missed. Any help would be appreciated.

mvglasow:
I’m merging the changes as they fix something that is definitely broken. This does exacerbate the startup bug, but the latter bug has been there all along, and is only an issue in those use cases which, as of now, are completely broken—that is, while the result is still not perfect, nothing will change for the worse.

jkoan:
Thanks for taking care of this and the other Android stuff like F-Droid :-)

mvglasow:
@jkoan I am just working on a separate issue for the startup bug and will post the link here. If you have anything to contribute, that would be much appreciated, as I currently feel I’m at the end of my wits…

mvglasow:
#750 deals with the Android startup bug.

PR #754, opened 2019-02-21
Android: Do not reinitialize everything when the activity is recreated

### LINKED ISSUE #750: Android: Startup is repeated unnecessarily when activity is recreated
### To reproduce

1. Start Navit for Android
2. Watch the logcat
3. Rotate the device, switch to split screen mode and back, repeat as needed

### Observed behavior

* Recreating the activity takes a long time, i.e. Navit shows a black screen for some time before finally displaying the map view (more evident when the traffic plugin is enabled and a large number of messages is in memory)
* If a recreation operation is still in progress when another one is triggered, Navit may stop responding altogether
* The logcat shows that the activity was destroyed and recreated, despite the manifest indicating that Navit will handle resize operations by itself. (Basically, this seems to be what the Android docs say: be prepared for your activity to get destroyed and recreated any moment.)
* Furthermore, the logcat shows that Navit goes through its entire startup sequence again, including initialization of all plugins, although that should not be necessary and costs extra time. It also complains about Navit leaking its `LocationListener`, which is registered as part of the startup sequence. Subsequent startup operations will register another one and throw away the old one without unregistering it properly.

### Expected behavior

Changing the dimensions of the Navit window should be quick, and startup operations not related to the activity should run only once on app startup, not whenever the activity is recreated.

### Versions tested

f4d8faa on Anbox

### Additional information

Digging through the code, the startup sequence seems to be the following:

* `Navit#onCreate()`
* `Navit#NavitMain()`
* `main_real()`

Most `Navit` members are already static and would survive recreation; those few which are not are related to the activity itself, and it makes sense to reinitialize them. The first two methods do nothing that is both time-intensive and unnecessary to repeat. `main_real()` is the all-platform entry point to Navit. Among others, it initializes all plugins and eventually starts the main event loop. This is the stuff that takes long and shouldn’t be necessary twice on Android.

I then did the following:

* initialize the static member `Navit.navit` to null
* in `Navit#onCreate()`, check if `navit` is null: if yes, we are starting from scratch; else we are just recreating the activity.
* add a boolean argument to `Navit#NavitMain()` which is true on launch, false when called for a recreated activity
* in `Navit#NavitMain()`, skip the call to `main_real()` unless we are starting from scratch
* in the same place, call `event_main_loop_run()` instead of `main_real()` for activity recreation
* in `Navit#onActivityResult()`, add a check to ensure `ActivityResult` is non-null before invoking its event handler

Alas, this does not work. I just get a black window, but I don’t know what I missed. Note that Anbox destroys and recreates the activity already on startup, thus we cannot even tell if the first activity instance works properly. Any help would be appreciated.

mvglasow:
Fixes #750. This significantly improves performance when the window dimensions change (as happens when entering or leaving split-screen mode, or when rotating the device), and also fixes some resource leaks and potentially a race condition.

Developed and tested on Anbox, as well as some brief tests on a real device.

### DIFF
--- a/navit/android.c
+++ b/navit/android.c
@@ -22,7 +22,8 @@
 #include "track.h"
 
 JNIEnv *jnienv;
-jobject *android_activity;
+jobject *android_activity = NULL;
+jobject *android_application = NULL;
 int android_version;
 
 struct android_search_priv {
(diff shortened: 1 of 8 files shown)


mvglasow:
Dunno why CircleCI isn’t running on the PR, but the run on my branch has been successful.

mvglasow:
Tested on a real device today, Navit keeps running and receiving location updates after switching to split screen and back. Response is almost instant, certainly faster than before. Ready to merge from my side.

mvglasow:
Last commit is for the pedestrian plugin (if anyone still uses that…), not tested beyond a quick glance at the code and ensuring the thing builds

lains:
Nice work, I will soon rebase my current workspace, build and try this!

PR #756, opened 2019-02-25
Fix:port/Android:Create notification channel on API 26+

### LINKED ISSUE #640: Notification icon missing again on Android Oreo
### Steps to reproduce

1. Install a recent build of Navit on Android Nougat (7.x).
2. Start Navit and observe the notification icon.
3. Install the same build on Android Oreo (8.x), or upgrade the OS on the device.
4. Start Navit again and observe the notification icon.

### Expected behavior

Notification icon appears while Navit is running, regardless of OS version

### Actual behavior

Notification icon appears on Android 7 but not on Android 8.

### Versions tested

Navit d421350 (actually a branch of it, with no changes regarding the notification), on LineageOS 14.1 and 15.1 respectively.

mvglasow:
Fixes #640. This restores the notification icon on API 26+ (Oreo and up).

I had to change the `NotificationCompat.Builder` in `Navit.java` to `Notification.Builder` as the latter allows us to specify a notification channel on API26+. `NotificationCompat.Builder` would require us to use version 26+ of the compatibility library (we currently use version 24), requiring us to increase the minimum SDK version.

`Notification.Builder` seems to work (tested on LineageOS 15.1), so I would stick with that unless someone can give me a compelling reason why we need `NotificationCompat.Builder`.

### DIFF
--- a/navit/android/res/values/strings.xml
+++ b/navit/android/res/values/strings.xml
@@ -7,6 +7,7 @@
   <string name="cancel">Cancel</string>
 
 	<!-- NOTIFICATION -->
+	<string name="channel_name">Navit</string>
 	<string name="notification_ticker">Navit started</string>
 	<string name="notification_event_default">Navit running</string>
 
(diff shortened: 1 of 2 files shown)


mvglasow:
Found the compelling reason for `NotificationCompat.Builder`: it is called API 9/10, which does not support `Notification.Builder`. Will look into it.

mvglasow:
Solved it by using `NotificationCompat.Builder` on API <= 25 and `Notification.Builder` on API 26+, should work everywhere now.

Issue #760, opened 2019-03-09
Multithreading


mvglasow:
As far as I can tell, Navit currently does not (or hardly) use multithreading. The standard mechanism for doing things in the background is to run them in an idle loop, on the main thread, with the following implications:

* We cannot harness the full power of a multi-core CPUs. When Navit has some heavy work to do, the UI may appear irresponsive. In the background, we are maxing out one processor core while the others may be idle. (And I might have contributed to that with the traffic module—when it has a large feed of messages to read on startup, there’s a lot of load in the background, which slows down the UI.) With multi-core systems becoming more and more common (by now, even smartphones and the current Raspberry Pi generation have multiple cores), this is becoming more of an issue than it was in the early days of Navit.
* Background loops are essentially cooperative multitasking: each task can keep the CPU until it voluntarily lets go. If anything runs into an endless loop, Navit will freeze.
* Time slices (which directly impact UI responsiveness) are often defined in units of work rather than in milliseconds, which means a time slice will be longer on a slower CPU, further degrading CPU responsiveness. (The traffic module uses milliseconds, but see the next point.)
* Granularity for time slices is limited, as each job needs to be chopped into units of work, and an interruption can only take place at the end of a unit. (For the traffic module, these units are rather large and a single unit of work usually takes longer than the time slice.)

I understand that multithreading is no small thing and needs to be worked out carefully. Specifically, we will need to synchronize accesses to our data structures to avoid data corruption.

However, the first question is: **is multithreading available on all platforms we support?** Or are there any that do not support multithreading? Even the latter would not be a show stopper, but would need to be accommodated in our design: for example, we could keep the code structure with idle loops but run them in a background thread where supported.

jkoan:
I did a debug session in eclipse lately and noticed that there where 3 threads running while I was debugging. I only remember that one was Sdl. Because I was debugging the Sdl graphics driver.

mvglasow:
That appears to be something happening behind the scenes, i.e. triggered by SDL, not Navit, and likely happens with a few platform-specific libraries. Unfortunately that doesn’t help us here—what I meant was: are there any instances where Navit itself (as opposed to some third-party library) launches additional threads?

mvglasow:
A quick scan of the core reveals references to the gthread (GLib thread) and pthread (POSIX thread) libraries.

[...]

PR #766, opened 2019-04-03
FEATURE:Read traff messages from a traff file

OLFDB:
This plugin uses a traff file located in NAVIT_USER_DATADIR to read traff messages. The whole file is read once and then monitored for new messages.

### DIFF
--- a/navit/traffic/traff_file/traffic_traff_file.c
+++ b/navit/traffic/traff_file/traffic_traff_file.c
@@ -48,7 +48,8 @@ char *traffic_read_traff_file(char *filename);
 
 char *traffic_read_traff_file(char *filename) {
 
-    char *data = 0;
+    char *data;
+    data = 0;
 
     if (traff_file_ptr == 0) {
         if (filename) {
@@ -59,16 +60,25 @@ char *traffic_read_traff_file(char *filename) {
     if (traff_file_ptr) {
 
         fseek(traff_file_ptr, 0, SEEK_END);
-        long end = ftell(traff_file_ptr);
+
+        long end;
+        end = ftell(traff_file_ptr);
+
         rewind(traff_file_ptr);
-        long start = xml_last_msg;
+
+        long start
+        start = xml_last_msg;
 
         fseek(traff_file_ptr, start, SEEK_SET);
 
         data = g_malloc(end - start + 1);
 
         if ((sizeof(data) - 1) > 0) {
-            while (fread(data, 1, (long) (end - start), traff_file_ptr));
+
+            while (fread(data, 1, (long) (end - start), traff_file_ptr)) {
+                ;
+            }
+
             data[(end - start)] = '\0';
             xml_last_msg = end;
         }
@@ -101,15 +111,12 @@ struct traffic_message ** traffic_traff_file_get_messages(struct traffic_priv * 
     struct traffic_message ** messages;
 
     dbg(lvl_debug, "enter");
-
     dbg(lvl_debug, "processing traffic from file: traff.xml");
 
     char *filename = g_strdup_printf("%s\\%s", getenv("NAVIT_USER_DATADIR"), "traff.xml");
-
     char * xml;
 
     xml = traffic_read_traff_file(filename);
-
     messages = traffic_get_messages_from_xml_string(traffic, xml);
 
     if (xml)
--- a/navit/traffic/traff_file/traffic_traff_file.c
+++ b/navit/traffic/traff_file/traffic_traff_file.c
@@ -112,7 +112,7 @@ struct traffic_message ** traffic_traff_file_get_messages(struct traffic_priv * 
 
     dbg(lvl_debug, "enter");
     dbg(lvl_debug, "processing traffic from file: traff.xml");
-    
+
     char *filename;
 
 #ifdef HAVE_API_WIN32_BASE
--- a/navit/traffic/traff_file/traffic_traff_file.c
+++ b/navit/traffic/traff_file/traffic_traff_file.c
@@ -66,10 +66,9 @@ char *traffic_read_traff_file(char *filename) {
         fseek(traff_file_ptr, start, SEEK_SET);
 
         data = g_malloc(end - start + 1);
-//           char data[end - start + 1];
+
         if ((sizeof(data) - 1) > 0) {
-            while (fread(data, 1, (long) (end - start), traff_file_ptr))
-                ;
+            while (fread(data, 1, (long) (end - start), traff_file_ptr));
             data[(end - start)] = '\0';
 
             xml_last_msg = end;
@@ -92,21 +91,8 @@ char *traffic_read_traff_file(char *filename) {
     /**
      * @brief Returns a traff_file traffic report.
      *
-     * This method will report two messages when first called: The messages indicate queuing traffic on the
-     * A9 Munichâ€“Nuremberg between Neufahrn and Allershausen, and slow traffic on the A96 Lindauâ€“Munich
-     * between GrÃ¤felfing and MÃ¼nchen-Laim.
-     *
-     * The 10th call will report an update message for the A9 (with a recent timestamp but otherwise the same
-     * data) and a cancellation message for the A96.
-     *
-     * They mimic TMC messages in that coordinates are approximate, TMC identifiers are supplied for the
-     * locations and extra data fields which can be inferred from the TMC location table are filled. The
-     * timestamps indicate a message that has just been received for the first time, i.e. its â€œfirst
-     * receivedâ€� and â€œlast updatedâ€� timestamps match and are recent. Expiration is after 20 seconds for
-     * messages in the first feed and 10 seconds for messages in the first feed (far below the lowest
-     * expiration timespan permitted in TMC).
-     *
-     * All other calls to this method will return `NULL`, indicating that there are no messages to report.
+     * This method will report all messages from a traff file when called first.
+     * All other calls to this method will return only new messages or NULL if none are available.
      *
      * @return A `NULL`-terminated pointer array. Each element points to one `struct traffic_message`.
      * `NULL` is returned (rather than an empty pointer array) if there are no messages to report.
--- a/navit/traffic/traff_file/traffic_traff_file.c
+++ b/navit/traffic/traff_file/traffic_traff_file.c
@@ -141,7 +141,7 @@ static struct traffic_methods traffic_traff_file_meth = { traffic_traff_file_get
  * @return A pointer to a `traffic_priv` structure for the plugin instance
  */
 static struct traffic_priv * traffic_traff_file_new(struct navit *nav, struct traffic_methods *meth,
-    struct attr **attrs, struct callback_list *cbl) {
+        struct attr **attrs, struct callback_list *cbl) {
     struct traffic_priv *ret;
 
     dbg(lvl_debug, "enter");
--- a/navit/traffic/traff_file/traffic_traff_file.c
+++ b/navit/traffic/traff_file/traffic_traff_file.c
@@ -113,7 +113,12 @@ struct traffic_message ** traffic_traff_file_get_messages(struct traffic_priv * 
     dbg(lvl_debug, "enter");
     dbg(lvl_debug, "processing traffic from file: traff.xml");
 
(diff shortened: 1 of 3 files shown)


navit/traffic/traff_file/traffic_traff_file.c
```diff
@@ -0,0 +1,177 @@
...
+    };
+
+    struct traffic_message ** traffic_traff_file_get_messages(
+            struct traffic_priv * this_);
+
+    /**
+     * @brief Returns a traff_file traffic report.
+     *
+     * This method will report two messages when first called: The messages indicate queuing traffic on the
+     * A9 Munichâ€“Nuremberg between Neufahrn and Allershausen, and slow traffic on the A96 Lindauâ€“Munich
```
jkoan: Encoding problem

navit/traffic/traff_file/traffic_traff_file.c
```diff
@@ -0,0 +1,177 @@
...
+     * between GrÃ¤felfing and MÃ¼nchen-Laim.
+     *
+     * The 10th call will report an update message for the A9 (with a recent timestamp but otherwise the same
+     * data) and a cancellation message for the A96.
+     *
+     * They mimic TMC messages in that coordinates are approximate, TMC identifiers are supplied for the
+     * locations and extra data fields which can be inferred from the TMC location table are filled. The
+     * timestamps indicate a message that has just been received for the first time, i.e. its â€œfirst
+     * receivedâ€� and â€œlast updatedâ€� timestamps match and are recent. Expiration is after 20 seconds for
```
jkoan: Again

navit/traffic/traff_file/traffic_traff_file.c
```diff
@@ -0,0 +1,177 @@
...
+#include "xmlconfig.h"
+#include "traffic.h"
+#include "plugin.h"
+#include "debug.h"
+#include <stdlib.h>
+
+long xml_last_msg = 0;
+FILE* traff_file_ptr = 0;
+
+char *traffic_read_traff_file(char *filename);
```
jkoan: Would it make sense to make a header?

navit/traffic/traff_file/traffic_traff_file.c
```diff
@@ -0,0 +1,177 @@
...
+        rewind(traff_file_ptr);
+        long start = xml_last_msg;
+
+        fseek(traff_file_ptr, start, SEEK_SET);
+
+        data = g_malloc(end - start + 1);
+//           char data[end - start + 1];
+        if ((sizeof(data) - 1) > 0) {
+            while (fread(data, 1, (long) (end - start), traff_file_ptr))
+                ;
```
jkoan: Why Semicolon in own line?

jkoan:
Also please take a look at the last output from ci, it hints you to fixes.

navit/traffic/traff_file/traffic_traff_file.c
OLFDB: Copy paste from dummy plugin. Fixed.

navit/traffic/traff_file/traffic_traff_file.c
OLFDB: Other plugins here doesn't have one, so I declared the function in the same way.

navit/traffic/traff_file/traffic_traff_file.c
OLFDB: Eclipse formatter used. Fixed

OLFDB:
CircleCI gives a comment "Formatted  navit/graphics/android/graphics_android.c", but that file was not changed in this CR.

navit/traffic/traff_file/traffic_traff_file.c
jkoan: Okay, I'm fine with that

jkoan:
I'm fine with this PR, before merging we should look at the ci and make sure that everythings okay. Even if the ci is red, I'm fine to merge this PR because it is one of the cleanest PRs we had in the last time

lains:
> CircleCI gives a comment "Formatted navit/graphics/android/graphics_android.c", but that file was not changed in this CR.

Yes, this seems to be a recurrent issue with sanity check on PR. There are formatting and indentation that are rejected by sanity check, while they have actually not been modified by the PR. If someone knows how to fix this... it's probably something related to the changed files detection:
https://github.com/navit-gps/navit/blob/7523c112affbf4f15fb04b5130b93cfc0a1d46fa/scripts/ci_sanity_checks.sh#L26

jkoan:
> > CircleCI gives a comment "Formatted navit/graphics/android/graphics_android.c", but that file was not changed in this CR.
> 
> Yes, this seems to be a recurrent issue with sanity check on PR. There are formatting and indentation that are rejected by sanity check, while they have actually not been modified by the PR. If someone knows how to fix this... it's probably something related to the changed files detection:
> 
> [navit/scripts/ci_sanity_checks.sh](https://github.com/navit-gps/navit/blob/7523c112affbf4f15fb04b5130b93cfc0a1d46fa/scripts/ci_sanity_checks.sh#L26)
> 
> Line 26 in [7523c11](/navit-gps/navit/commit/7523c112affbf4f15fb04b5130b93cfc0a1d46fa)
>  # List the files that are different from the trunk

Yes, i will look into it, last time i was not able to reproduce it

OLFDB:
OK. I have adpated the Eclipse formatter to match the guidelines required by CircleCI and traffic_traff_file.c is fine now.

mvglasow:
Great to see someone else is taking up work on the traffic module! Apparently you didn’t see my comments in the IRC logs a few days back—never mind, at least that gives me time to sort out what I need to get across :-)

TraFF is based on XML, which doesn’t lend itself too well to streaming information. For some in-depth information about TraFF, you might want to check out https://gitlab.com/traffxml/traff, which explains the ideas behind TraFF and what I had in mind for message exchange. (Which is also how the `traff_android` plugin is implemented.)

Message exchange is based on feeds. There can be multiple sources sending feeds, and multiple consumers receiving them. Maybe tomorrow OSMand will process TraFF as well. Or we’ll have another provider over which drivers can create and share share traffic reports, or one which performs image recognition on variable-messsage signs and extracts traffic information from them.

Each feed is a complete XML document with (only) the new messages in it. A feed is sent by the source (e.g. Qz) on its broadcast channel to all interested consumers (e.g. Navit). A consumer also can (and usually needs to) send a poll request (“show me everything you have”), to which the source will reply with the complete list of currently active messages.

Future versions may need to include additional capabilities, such as applications being able to subscribe to certain regions and getting updates only for that region.

Feeds can be implemented with a variety of transport media. Between two local applications, my idea was to use the request broker architecture of the respective platform. On Android that means intents. On Linux this would be DBus. There is, in fact, a rudimentary DBus implementation in `binding_dbus.c` (look at `request_navit_traffic_inject()`). It is not perfect, as it relies on temporary files for data storage—a better option would be to send the XML data over the DBus link. Also, it uses directed messages (only Navit will receive those); a broadcast would be the better option. (Admittedly, I do not know my way around DBus very well, therefore I’m not sure if and how these things can be done.) On Windows the mechanism of choice would probably be DCOM (which I haven’t used in decades, though). Feel free to make a proposal over at TraFF, this part is not yet worked out in detail, although much needed.

As an alternative mode of transportation, bare TCP/UDP sockets would probably work as well, but for anything that is not a broadcast mechanism, we would need a subscription mechanism so sources would know where to send their data. An HTTP API would also be an option for the future, but this would require consumers polling the source for updates and indicating the timespan they are interested in. (Again, the TraFF project would happily welcome a proposal—feel free to open a ticket or draft a spec.)

A file is probably not the best way to do that—monitor mode might work, but still feels a bit hacky to me. For example, we’ll always be dealing with malformed XML—either the root element is still open, or we have multiple root elements. Also, once we introduce subscriptions (or service status), we might need to add attributes to the root element, and these might change from one feed to the next.

Sorry if that means extra work, but I can assure you that you are not alone in this… traffic took lots of iterations to get right (and about a year to get anywhere near usable), and multithreading is very similar in that.

Regarding the poll interval for TraFF, bear in mind that I am currently working on multithreading (#760). Traffic is the first candidate I am moving into a separate thread (essentially `traffic_process_messages_int()` and everything it calls), so it will not block the UI any more (unless both are trying to access the same data structure at the same time). The 1s interval will then become the duration for which the traffic thread sleeps when it is done processing all new messages.

I'd be interested in seeing your Qz port, though. I had something like that in mind but never got around to actually implementing it (I use Android for production, for Linux tests I rely on canned data injected through the test interface).

OLFDB:
The poll interval is not an issue any longer, as with the monitor only new messages being read. I agree a file is hacky. Maybe an idea is using JMS for transport. That could be used on Linux, Windows, OSX.

My Qz port is a mess currently. I mocked all the android classes in a new package and changed the imports of all the files I needed to get the TMCReceiver up and running. It writes into the traff file with each message wrapped in <feed> </feed>. This is working fine. I gave it a first live try in Berlins traffic yesterday on Windows. Do you have an idea how to import the mocked or real android classes based on the build target? As in Java I can't use preprocessor defines this is a problem currently. The TMCReceiver needs to get some private->protected changes so it can be a super class. There is no GUI currently and I'm not sure if it is worth to create one. Would be better to make Navit showing more details on the map like icons for the type of issue, times etc. A big deal would be to make Qz TPEG capable. e.g. using the Monkeyboard DAB+ receiver to get the raw data.

mvglasow:
About the Qz port, I have created an issue at https://gitlab.com/mvglasow/qz/issues/53.

For mocked Android classes, you could place them in a JAR and include that, or turn the mock classes into a Git submodule (as I did with the other Qz dependencies).

> Maybe an idea is using JMS for transport. That could be used on Linux, Windows, OSX.

Sounds good, but would this also work with native applications (such as Navit)?

> Would be better to make Navit showing more details on the map like icons for the type of issue, times etc.

Absolutely. See https://github.com/navit-gps/navit/projects/8. The graphics driver is currently very limited. On the other hand, I am somewhat concerned about cluttering the map (which is also why I use a map theme derived from 0606.at, rather than the default one). It’s all too easy to fall for the fat-and-dirty approach, but deciding what **not** to display can be just as important. Therefore I would keep the map display somewhat minimalistic but implement a list of active traffic messages in the GUI.

> A big deal would be to make Qz TPEG capable. e.g. using the Monkeyboard DAB+ receiver to get the raw data.

That would in fact be a new application. Just to give you an idea, it took me roughly two years to extend RDS Surveyor with all the functionality I needed (extra tuner support, TMC location parsing), draw up the TraFF specification (including the list of messages and how to translate TMC to TraFF) and finally write Qz around it. Also, I am not currently exploring that option as Germany seems to be the only place that has free-to-air TPEG services (https://en.wikipedia.org/wiki/TPEG#TPEG_services_worldwide); everywhere else things seem to be encrypted and require payment. I am currently leaning more towards a crowdsourced traffic service where people can report congestions they are observing.

jkoan:
@mvglasow so what do do with this PR? As i understood, since this would read xml and xml is not streamable this whould not make much sense to merge, right?

jkoan:
Cloasing this as xml files are not really streamable. But feel free to open new PRs at any time

mvglasow:
@OLFDB I have recently started working on exchanging TraFF via HTTP. It’s still in its infancy, but maybe this is something that might work for you.

Source code: https://gitlab.com/traffxml/traff-server.

Documentation of the subscription mechanism (still in draft status): https://gitlab.com/traffxml/traff/-/blob/dev/TraFF%20Subscription%20Management.pdf

mvglasow:
TraFF server is somewhat usable with the available sources, and the TraFF 0.8 spec (which introduces HTTP transport) is now final.

I am now considering a TraFF HTTP plugin, which would be available on most (ideally on all) platforms to which Navit has been ported.

What we would need is HTTP client functionality in Navit. The only thing we have so far is the map download code in navit/file.c, most of which seems to have been written back in 2010/2011 by Martin Schaller. However, it is quite rudimentary and limited to plain HTTP (no TLS) and GET (no POST). Therefore, I would either have to implement all of this from scratch (which feels like reinventing the wheel) or introduce some third-party HTTP client library.

Hence the first question: has anyone ever considered using an external HTTP client library for Navit?

If we go down that route, we could also replace the code in file.c with calls to the library—would be cleaner than having two independent HTTP implementations. That would require the help of someone who is familiar with this portion of the code and would know how to port it to another support library.

PR #769, opened 2019-04-04
Zip code search

OLFDB:
FEATURE:internal gui command 'postal_code' to search destination using postal codes.


OLFDB:
Too many changes.

PR #771, opened 2019-04-04
Fix:core:Fix item_def.h when used whithout defined macros (like in IDE)

jkoan:
Hi, 

always when in a IDE like Eclipse i head the problem that all attr_* definitions where not found by the IDE and therefore where marked as error, so my approach now is to declare them as extern so the IDE does not get annoyed by this. This also fixes auto competition.

### DIFF
--- a/navit/item_def.h
+++ b/navit/item_def.h
@@ -18,6 +18,13 @@
  */
 
 /* This file is generated from http://wiki.navit-project.org/index.php/Item_def.h, do not edit it, edit the wiki page instead */
+#ifndef ITEM
+#define ITEM(x) extern ##x;
+#endif
+#ifndef ITEM2
+#define ITEM2(x,y) extern ##x;
+#endif
+
 ITEM2(0x00000000,none)
 ITEM2(0x00000001,point_unspecified)
 ITEM(town_streets)
--- a/navit/item_def.h
+++ b/navit/item_def.h
@@ -22,7 +22,7 @@
 #define ITEM(x) extern ##x;
 #endif
 #ifndef ITEM2
-#define ITEM2(x,y) extern ##x;
+#define ITEM2(x,y) extern ##y;
 #endif
 
 ITEM2(0x00000000,none)


mvglasow:
Much needed indeed (this has been bothering me as well for some time :-)

The only doubt I have is related to this comment:

    /* This file is generated from http://wiki.navit-project.org/index.php/Item_def.h, do not edit it, edit the wiki page instead */

Does that still reflect the current status? If yes, we’d need to make sure the extra declarations survive an update. If not, the comment should be removed.

jkoan:
I think this is not up to date. At least we don't do this automatically. There is a script for that but I don't even know if it is functional. So my opinion would be to remove the comment

mvglasow:
I am generally not convinced that generating a source file from a wiki page is such a great idea. However, I would suggest doing a test run of the script to see if it runs at all.

* If the script fails and is not trivial to fix, I’d suggest removing it. If the fix is easy, fix and retry.
* If the script runs but ends up removing item types (suggesting the wiki is out of date), I’d also suggest removing the script.
* If the script runs and generates essentially the same file (without the fixes), then I would suggest extending the script to generate `item_def.h` with the fixes.

With that I would give my approval. If we end up removing the script, we might consider updating the wiki from the source file—but that doesn’t have to go in this pull request.

BTW, Eclipse equally nags about stuff in `attr_def.h`—would the fix work there as well?

pgrandin:
100% onboard with @mvglasow 's approach.

jkoan:
> I am generally not convinced that generating a source file from a wiki page is such a great idea. However, I would suggest doing a test run of the script to see if it runs at all.
> 
>     * If the script fails and is not trivial to fix, I’d suggest removing it. If the fix is easy, fix and retry.
> 
>     * If the script runs but ends up removing item types (suggesting the wiki is out of date), I’d also suggest removing the script.
> 
>     * If the script runs and generates essentially the same file (without the fixes), then I would suggest extending the script to generate `item_def.h` with the fixes.
> 
> 
> With that I would give my approval. If we end up removing the script, we might consider updating the wiki from the source file—but that doesn’t have to go in this pull request.


The Script is functional and works but the wiki is outdated. So remove Script and edit wiki to say that item_defs are now handled in the item_def.h file?

Currently the state of the Wiki would breake things as Items would be added what would change places.

> 
> BTW, Eclipse equally nags about stuff in `attr_def.h`—would the fix work there as well?
We Could fix this as well, i will open another PR for this

mvglasow:
> The Script is functional and works but the wiki is outdated. So remove Script and edit wiki to say that item_defs are now handled in the item_def.h file?

Sounds good to me—and remove the comment in `item_def.h`, which is now obsolete.

jkoan:
I will do this in another pr where I also remove the script

PR #775, opened 2019-04-23
fix:graphics:Fix issue when try to close the SDL window ion non webos

jkoan:
Currenty the window created by the SDL driver cant be closed with wihe window X button.

This fixes this issue


mvglasow:
@jkoan I trust your tests, don’t feel I’m competent to make a judgement here as I don’t use SDL.

PR #778, opened 2019-05-17
Fix:core:Fix Warning in IDE for Items not found

jkoan:
As discussed in #771 and requested by @mvglasow hers the same fix for attr_def.h

### DIFF
--- a/navit/attr_def.h
+++ b/navit/attr_def.h
@@ -29,6 +29,12 @@
 /* prototypes */
 
 /* common */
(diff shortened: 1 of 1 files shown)


mvglasow:
Umm… this changes `item_def.h` (as #771), not `attr_def.h`, looks like you committed the wrong file

jkoan:
Oh yes, true. Will fix this

jkoan:
Now its better ;)

PR #782, opened 2019-05-25
Sync bookmarks 2

jkoan:
As a followup for #435 heres a version which fixes the Problem for all Plattforms. 
Original Text from @pgrandin:

> When using navit on some devices that could loose power without being properly shut down, the bookmark files was lost (content was empty).
> 
> This problem is reproduceable : running Navit on a buildroot image on the raspberry pi (for example one built using https://github.com/navit-gps/buildroot ), if the pi looses power and the bookmark file has been modified, the bookmark file will be lost everytime.
> 
> This simple fix ensures that the files get flushed to the disk when we write the bookmark files, and solves the issue according to my tests.


pgrandin:
Thanks for finishing my half-assed work :smile:

PR #791, opened 2019-06-12
update:doc: add the submit patches section from the old wiki close to the code

aerostitch:
I think it makes sense to start putting elements to contribute to the code inside the `CONTRIBUTING.md` document to centralise that a bit more like every other project do.
This PR migrate the https://wiki.navit-project.org/index.php/Submitting_patches document from the wiki to here and do some updates on it.


pgrandin:
Thanks @aerostitch 
We will probably need to rework the content at some point, but it's definitely a good start.

aerostitch:
Thanks I'll migrate other docs in the days to come.
Bisou beco! :kissing_closed_eyes:

PR #798, opened 2019-06-18
qt5: Remove unnecessary moc includes.

metalstrolch:
This commit removes moc file includes for some classes which do not
define new qt classes. This caused a long standing warning which seems
to be treated as error on some setups.

Additionally a small fix to make nit-picky gcc 8 happy.


jkoan:
@metalstrolch QAudio::InterruptedState was added in qt 5.10. 
@hoehnp can we update Sailfish build image to a newer qt version?

metalstrolch:
No, unless Jolla decides to do, we cannot update QT

PR #803, opened 2019-06-21
fix:core:Fix run when in build dir

jkoan:
This PR should fix the issue for not finding the right files when navit in run through the build directory

### DIFF
--- a/navit/main.c
+++ b/navit/main.c
@@ -350,7 +350,7 @@ void main_init(const char *program) {
     setlocale(LC_NUMERIC,"C");
 #endif
 #if !defined _WIN32 && !defined _WIN32_WCE
-    if (file_exists("navit.c") || file_exists("navit.o") || file_exists("navit.lo") || file_exists("version.h")) {
+    if (file_exists("navit.c") || file_exists("navit.o") || file_exists("navit.lo")) {
         char buffer[PATH_MAX];
         printf("%s",_("Running from source directory\n"));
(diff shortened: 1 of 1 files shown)


jkoan:
Fix for Tomtom will be addressed in another PR, its not related to this PR and occures because of outside network effects

jandegr:
Hi,
Not checking for `version.h` won't help a bit, which makes me think this proposed fix is yet another one that has not been tested properly.

Looking for `Makefile` might be an alternative and a first test shows it works, see https://circleci.com/gh/jandegr/routing-qa/582#artifacts/containers/0

It would be odd that somebody would run it from a place where something else had been built and left a `Makefile` behind, but if somebody can think of a sideeffect or propose a better, working, solution the pls. do so.

IIRC correctly, in the old days in CI, after compilation maptool would compile the sample map, navit would then be started with the produced binfile and a screenshot would be imported. That was real testing and would prevent such mishaps from beeing committed.

jkoan:
@jandegr you are right, it was my mistake while testing. I will try do do a better fix soon. But checking for Makefile is not a solution ether, since cmake cant use a few other generators. checking for lib directory's or other build results could be a solution or even create a file within BINARYDIR which is empty and its only purpose is to be detected in this case.

jandegr:
ok, the workaround I came up with is far from optimal either, that is true, but I just had to find some temporary solution to get QA back running, otherwise it was just dead in the water :)

Issue #777, opened 2019-05-10
Navit installer for Windows doesn't create maps folder


lexterror:
Navit installer for Windows doesn't create maps folder. Also won't load *.bin maps.

pgrandin:
I had a look and I can confirm the following issues :
- the maps folder isn't created by the installer
- enabling the map as part of the sample mapset doesn't work
- debug logs aren't available, regardless of whether they are enabled via the config or the command line flag. All I can get so far is 
```
��SPAGE_SIZE: 4096, SVALIGN:33554432
return dwTlsIndex = 0x4
```

pgrandin:
Progress!
The issue with logs on win10 is that they are stored in a convoluted place instead of being next to the xml files. Something like
`C:\Users\John\AppData\Local\VirtualStore\Program Files (x86)\Navit`

jandegr:
`error:map_binfile:module_map_binfile_init:error: sizeof(struct zip_cd)=zu`

There was a check for this in zipfile.h that would break compilation if the configuration of the build is not ok, but that check was removed in https://github.com/navit-gps/navit/commit/84ec04ba1ca2f36d591a297fa42a93a5f04de747#diff-9192dfcce35a26969c16e0cf0525bd39

Now you get an error at runtime instead of at compiletime and impossible to load any binfile.

Regards.

jandegr:
A little more info,
For toolchain file to build a 32bit version on a 64 bit box I have **mingw-w64-i686.cmake**

> set(CMAKE_SYSTEM_NAME`` Windows)

>set(TOOLCHAIN_PREFIX i686-w64-mingw32)


>FIND_PROGRAM(CMAKE_C_COMPILER NAMES i686-w64-mingw32-gcc)
>FIND_PROGRAM(CMAKE_CXX_COMPILER NAMES i686-w64-mingw32-g++)
>FIND_PROGRAM(CMAKE_RC_COMPILER_INIT NAMES i686-w64-mingw32-windres)
>FIND_PROGRAM(CMAKE_AR NAMES i686-w64-mingw32-ar)

> target environment on the build host system
>set(CMAKE_FIND_ROOT_PATH /usr/${TOOLCHAIN_PREFIX})

> modify default behavior of FIND_XXX() commands
>set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
>set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
[...]

PR #806, opened 2019-06-27
jandegr's fix for the windows builds issues

pgrandin:
The resulting build has been confirmed to fix the issue.


jandegr:
Tested on win10-64 and I confirm it can read binfile again and no errormessages (except for the speech issue as reported in https://github.com/navit-gps/navit/issues/777)
On IRC DuncanDo confirmed it working as well on june 4

PR #807, opened 2019-06-29
update:doc:Fix the readthedocs url for the development guidelines

aerostitch:
The readthedocs url is available for the trunk branch but not the latest tag in readthedocs so let's use that branch instead.


aerostitch:
thx!

PR #810, opened 2019-07-15
Add:osd/core:honor screen dpi

metalstrolch:
This pull request adds the possibility to specify OSD items and OSD icons in human units correctly honouring screen dpi. It allows to give for example w="15.2mm" or "x=-20.5mm" in config file for osd items displaying them the same size regardless of the device's DPI. This allows to have the same OSD configuration for quite different dpi devices.

For example on sailfish the high dpi sony xperia xa2 vs the standard Jolla1. 

Eventually I'll enhance this text sizes as well, or maybe even for map drawing.

Does contain:
- Small backward compatible interface change for graphics plug ins.
- Change in attr handling for attr_type_rel_abs eventually overcoming ATTR_REL_RELSHIFT magic.
- Change in attr handling for internal gui (icon_xs, icon_s, icon_l and spacing)

What works already:
- qt5 graphics already gives dpi value. Needs to be implemented for different graphics as well.
- gtk_drawing_area gives dpi value.
- standard OSD items x,y,w,h values. Not all tested though.
- on map icons w and h values.
- internal gui icon sizes can take human values and relative values (%) as well.

What works not:
- icon_w and icon_h for those that support this. Don't know if this makes sense at all.
- font size handling. But no regression on existing configs. 
  Need to understand the unit navit works with for font size, to learn how to convert from e.g. mm to that.
**- setting the changed attrs from bindings like d-bus may fail without porting the remote application.**

Comments are welcome, of course.


metalstrolch:
**Help wanted:**
What's the current unit of navit font sizes? How to convert to that from e.g. millimeters, inches, or pixels. Currently it doesn't seem to be pixels.

metalstrolch:
Help wanted:
What applications exist dealing with "w h x y icon_xs icon_s icon_l spacing" from remote and what purpose for?
Who knows about how the bindings are used?

mvglasow:
I may be a bit late to the party but I would suggest considering the way other platforms deal with different screen densities.

Android (being the one I am most familiar width) has the concept of the “display-independent pixel” (or was that density-independent?): 1 dp (or dip) corresponds to 1 pixel on a 160 dpi screen (the baseline resolution). When Android gets a size specified in dip, it will scale it according to the reported screen density of the device. For bitmap resources, it will pick a differently-sized bitmap if available.

To simplify things, devices do not report their precise DPI but are divided into “density bins” of 120, 160, 240, 360 and 480 dpi (LDPI, MDPI, HDPI, XHDPI and XXHDPI, respectively).

iOS, as far as I know, follows a somewhat similar approach: pixel sizes refer to the density of the original iPhone (which was around 160 dpi) and are scaled on later models with different screen densities.

Windows has had a registry setting for screen density since Windows 95/NT 4.0. Default was 96 dpi (roughly equivalent to a 24″ screen at 1920×1080). Back in the days, fiddling with the DPI did not have any effect on the dimensions of UI elements, though some applications would resize their content accordingly (print previews in Word being one case I remember). Not sure if this is the same setting by which more recent versions of Windows determine UI dimensions.

No idea how Unix-like platforms do this; I remember seeing something long ago about 75 dpi being the baseline assumption (roughly equivalent to the density common on CRT screens back in the days).

I would in fact go for a unit such as “pixels at baseline resolution” combined with density bins, thus avoiding arbitrary scaling. (A bitmap scaled to 2.851849 times its native size doesn’t look too good, especially on low-density screens.) Also, metric sized don’t lend themselves too well to this purpose, as screen measures are mostly inch-based.

For the baseline resolution, 160 dpi would be good for compatibility with mobile OSes. The original assumption of pixel density in Navit was probably around 100 dpi (as higher display densities weren’t yet a thing for another 5 years). A good compromise would be 120 dpi, sufficiently close to the density of traditional desktop displays and still compatible with the density bins mentioned above, so I would go for a baseline of either 120 or 160 dpi.

metalstrolch:
Yes that's exactly what i try to do here. I'm not quite shure to what extend.
Usually you have your dpi value. Which means how many pixels per inch your display has.

Then you need a physical unit:
I choose "mm" and "in". It's simple to add "dip" (2 lines). Because all of them are physical units.
1 in == 1 * dpi pixels
1 mm == (0,0393701) * dpi pixels
1 dip == (1 / 160) * dpi pixels

**-> you always need to scale things including bitmaps (or better use svg) to screen. Android does this as well. But usually only upscaling if higher dpi. That's barely visible.**

Navit currently uses pixels internally on interface between navit core and the graphics. The text size is in pixels as well, but a strange unit is used here due to somebody once misunderstood freetyp I think.

My concern is that I don't want to break old configs. All configs currently give everything in pixels. I try to enhance this by allowing to specify a unit in configuration: 1 (pixels), 1mm (millimeters) 1in (inches). We could add 1dip (android dip).

**But maybe you're right, We could (globally) make navit work with a virtual pixel unit, and make the unit selectable. This would best be done directly inside the graphics plugins then. But this will effectively limit the "addressable" pixels for navit, which on the other hand increases processing speed on high dpi screens. BUt this attempty may be way simpler and less intrusive...**
Think I'll roll a second attempt on that. (Now I know how text size is given...)

metalstrolch:
Closed, as #813 is the better attempt.

PR #794, opened 2019-06-14
Add/Android: Provide geo coordinates for clicked position in contextual menu

lains:
When performing a short clicking/tap on an point on the map, in Android, a contextual menu is opened.
This menu mainly allows to start navigation to the point that has just been clicked on.
This PR adds two features:
* it adds the geographical coordinates of the point clicked on the map in the contextual menu title:
Screenshot_Navit_20190613-085619

* and it fixes the displayed name of the location in the destination history (before it used screen coordinates as a bookmark names, which does not make much sense for the final user, and was partially broken, as only X coordinates were used).
The screenshot below allows to compare both the new destination name format (top), and the old destination format (second entry named "334"):
Screenshot_Navit_20190613-085808

In addition, this PR also adds a mechanism to escape and unescape double quotes (") in text files (required to store geographical coordinates, as they use character (") for seconds).

Issue #678, opened 2018-09-30
Allow QR code input for GPS coordinates


lains:
On Android, when scanning a QR code for GPS coordinates, and allow navit to handle this type of input to open a contextual menu for these GPS coords (same GUI as when typing the coordinates manually).

To generate QR codes for geo coords, use "Contents" type="Geo location" on the following website
https://zxing.appspot.com/generator

For example, the Android application "Barcode scanner", when scanning a geo QR code, allows two actions: "Display map" or "Navigate there"

Example QR code:
![eiffel_tower](https://user-images.githubusercontent.com/383502/46258260-a45e9c80-c4c7-11e8-9875-b1f42ef8c16e.png)

jkoan:
This should be easy by processing the geo intent (which will be send by the qr-scanner app of your choise) which is planned aleready ;)

lains:
I am currently implementing a similar feature , but the other way around:
In Android, when tapping on a specific location on the map, have the contextual menu allow to open the location in an external Android application (using the geo intent).

lains:
Implemented in #812

PR #812, opened 2019-07-20
Add/Android Supporting sending & receiving geo: intents

lains:
This PR adds 2 features to navit for Android:

- When a geo: intent is started on Android, for example by scanning a geo QR code, navit is able to takes these coordinates as input

- When a specific point is clicked on the map, this PR adds a new item in the contextual menu: "View". By clicking on that item (and if there are applications to handle it) a geo intent is generated in Android


lains:
This PR adds the feature described in https://github.com/navit-gps/navit/issues/678
If you scan that QR code, navit will use the geo location.

lains:
This PR is a work in progress, it should not be merged as is as it is based on PR https://github.com/navit-gps/navit/pull/794 , and it thus requires https://github.com/navit-gps/navit/pull/794 to be merged first.

navit/android/src/org/navitproject/navit/NavitGraphics.java
```diff
@@ -37,6 +40,7 @@
 import android.os.Message;
 import android.support.v4.view.ViewConfigurationCompat;
 import android.util.Log;
+import android.net.Uri;
```
jkoan: Build fails due to checkstyle is failing:
[ant:checkstyle] [ERROR] /home/circleci/code/navit/android/src/org/navitproject/navit/NavitGraphics.java:43: Wrong lexicographical order for 'android.net.Uri' import. Should be before 'android.util.Log'. [CustomImportOrder]

Could you please fix this?

lains: Yes, sure!

lains: OK, done!

lains:
@jandegr, I think [the tweak you did here](https://github.com/navit-gps/navit/pull/877/files#diff-de01e9991ecc98ddc67d26f95a4e22f3R231) needs to be added to the current PR to avoid the loop you're mentionning.
I should maybe even add code to remove navit from the proposed applications so that we never propose the intent to loop back to navit... if this can be done on Android... but I have [a track to investigate](https://stackoverflow.com/questions/28492579/give-the-option-to-always-choose-a-browser-to-open-a-link#28492633).

jandegr:
Hi,
I removed the receiving geo part in 877. Users can already enjoy sending intents while the other issues are being solved I suppose. I have a few concerns with the receiving geo part but to solve these the differences between navit-gps and https://github.com/navit-gps/navit/pull/877 become relevant, so I did not digg deeper for now.

lains:
Hi,

Good stuff. Can you please point me to some sample code so that I can filter out navit from the list of app candidates to receive the goe: intent I send from navit?
When navit is the receiving app for an intent issued by navit itself, you're right, we end up with two apps running, sometimes it even crashes, so I have to deal with this case.

In the future (in another PR), I am planning to build on this code to open a contextual menu when a geo: intent is received. Probably, this could be an internal GUI menu showing options like "set as destination", "view POIs", "add to bookmarks" etc.

jandegr:
If I read it correctly then this 

https://developer.android.com/reference/android/content/pm/PackageManager#queryIntentActivityOptions(android.content.ComponentName,%20android.content.Intent%5B%5D,%20android.content.Intent,%20int)

could provide a filtered list using the caller param, if I did not read correct then you can still remove navit from the list yourself.

For the chooser, postings like this might help
https://stackoverflow.com/questions/5734678/custom-filtering-of-intent-chooser-based-on-installed-android-package-name
You may have to dig throug a number of those to get a correct solution.


> When navit is the receiving app for an intent issued by navit itself, you're right, we end up with two apps running

I think this will happen if Navit is already running and you launch an intent from another app just the same, we just noticed it because of the loopback, or have you tested more ?


> sometimes it even crashes

PR 877 comes with a 1 year/20000km warranty for such crashes, I am very interested in scenarios that can cause one related to android code.


> In the future (in another PR), I am planning to build on this code to open a contextual menu when a geo: intent is received. Probably, this could be an internal GUI menu showing options like "set as destination", "view POIs", "add to bookmarks" etc.

ok to deal with it later, my take on it would be to handle it with a dedicated activity but feel free to think otherwise

lains:
I am going to rework the part that sends intents so that it does not propose to loop back to navit. I already did some tests that looks promising.

navit/android/src/org/navitproject/navit/Navit.java
```diff
@@ -83,7 +83,7 @@
     private static final int           NavitDownloaderSelectMap_id     = 967;
     private static final int           NavitAddressSearch_id           = 70;
     private static final int           NavitSelectStorage_id           = 43;
-    private static final String        NAVIT_PACKAGE_NAME              = "org.navitproject.navit";
+    public static final String         NAVIT_PACKAGE_NAME              = "org.navitproject.navit";
```
jandegr: Hi,
Can you tell me a good reason for this ?
and why public ?
thx

lains: Hi. I made it public because I'm using it in NavitGraphics to filter out Navit's package name from the intent list, but maybe there is a better way...
https://github.com/navit-gps/navit/blob/bf282c648b3f6648405f00a69489006ab909f9d1/navit/android/src/org/navitproject/navit/NavitGraphics.java#L242

jandegr: Hi,
public is only for interaction with the world outside of Navit, so even if you would preserve this solution, package private is more than enough. 
Almost each time we have smth. with public it is some method Android will call, like the onTouch and so. 

Then the subject itself, if you have a getter for smth. then it is preffered to use that as you did before by using getPackageName(). 
I was attracted to take a look into this commit because of the word 'guesswork' in the title of the commit, did the solution with the getter give you strange results ?

I plan to remove the hardcoded packagename in Navit.java on the first occasion, I have already done quite a bit of cleanup and that is a leftover.
Where does it all lead to ? the coupling between the classes gets lower

Does your work to filter out Navit from the list work ok ?

THX!

lains: OK, I reverted back to the previous solution: it did work, but when I saw the package name value was already directly accessible as a string, I switch to that solution.

Filtering out navit from the list does work (tested on 4 devices). Please have a look, it should work in all cases (no app for intent, 1 app for intent and multiple apps for intent), without proposing navit and thus without any crash anymore.
If so, this PR will be ready for merge.

lains:
> PR 877 comes with a 1 year/20000km warranty for such crashes, I am very interested in scenarios that can cause one related to android code.

Do you want me to try and reproduce the issue of reopening navit from navit, by re-building on a previous commit (this is when I had a crash)?

jandegr:
The point is that after PR 877 I want it to be crashfree, sure it was strange in the loopback situation so if you can make it crash on code after PR 877 I would be interested.

jandegr:
This one https://circleci.com/gh/navit-gps/navit/17716#artifacts/containers/0

on the SM-J600G of STF @pgrandin 

View leads to Googlemaps without the chooser menu, so I suppose there is only Googlemaps and Navit on the device to receive geo and with that assumption it behaves as expected and sends it to Googlemaps without showing the chooser.


520004958d47b5dd



After installing one more app to receive geo intents I now get the chooser.


520004958d47b5dd2

jandegr:
Receiving a geo intent works but I just have the impression I end up with multiple instances of Navit at the same time, especially if Navit was already running backgrounded. I compared this with the behaviour of a PDF viewer, there I can try whatever I want, no way I could get multiple instances running.

A screenshot of 2 Navit's running in splitscreen mode after sending a geo intent to Navit.

Screenshot_1570266625

jandegr:
The deault launch mode is 

> 

"standard"| | Default. The system always creates a new instance of the activity in the target task and routes the intent to it.
-- | -- | --


where you could try with singleTop, singleTask or singleInstance

lains:
Hi @jandegr, I could not reproduce the issue, but I'm sure I've noticed before this feeling of multiple instances running.
I have implemented the changes you suggested, and the singleTask is indeed a singleton, so this is probably what is needed.
Please let me know if this new version of the code fixes it.

lains:
> I could not reproduce the issue, but I'm sure I've noticed before this feeling of multiple instances running.

Actually, I just reproduced the issue on a Samsung device running Android 4.4.2.
And there, it seems fixed by the singleTask launch mode...

lains:
And I also restested the intent loopback issue (crash) after I rebased on the code from trunk including PR877 and it did not crash. So, I think you already fixed the crash in that PR.

jandegr:
I will take some time to test it soon, but even without testing I suspect it now indeed sticks to being a singleton, but since you did not implement onNewIntent(), it will just recreate navit and that is not what we want. I might be wrong, but I just think it will now show in the log it takes the onCreate() route (or maybe, but probably not, the onRestart() route) since you don't open the onNewIntent entrypoint. At least it is what the API makes me think.

THX

lains:
I have tested on this new code and receiving a geo intent now works as expected (starts navigating to the coordinates provided) even when navit is already running.
If navit was inside a menu when the intent is received, it remains inside that menu. Not sure whether this should be kept like that or if map should be displayed immediately and current menu dropped.

I am already planning to build some future work on this code base, opening a contextual menu based on the geolocation extracted from the intent, so that the exact action to perform on the input coordinates can be selected by the user, exactly the same way as when coordinates are input using the Action>Coordinates menu in the internal GUI...
But I would rather not add more code to the current PR and handle this in a future PR [as explained in a previous comment](https://github.com/navit-gps/navit/pull/812#issuecomment-535811698).

@jandegr already told me sending geo intents is working but I fixed receiving geo intents since then, so I'd be happy if someone can test and give me feedback on this feature.

Shall I merge?

Thanks!

jandegr:
Hi,
There are failing scenario's for the existing google.navigation intent in the existing codebase and your work just inherits those.
There are several ways to deal with this and the way that not turns this into an endless story for you is you commit this and I will apply the fixes afterwards together with some other rework. You will see some of your lines being shuffled around afterwards.
The other way is you change all of it into a solution working in every scenario for both the intents.

your choice.

jandegr:
@lains pls. merge this since future work depends on it.
After merging I can post another (but smaller this time) Android rework and then you can proceed with the future extensions you mentioned whenever it suits you.

THX

lains:
Great. Thanks @pgrandin for merging this.
@jandegr, yes, I will wait for #919 before doing further work on the view intent.

PR #813, opened 2019-07-21
add:graphics:dpi scaling support

metalstrolch:
This pull request adds dpi scaling to navit's graphics layer. It basically allows to convert the internal measurement unit (traditionally 1:1 pixels) into something virtual. You can specify the dpi value your layout / OSD was designed against in virtual_dpi, effectively assigning the internal pixel a human unit.
Then it will get the screen dpi value (or you can override that in config) to scale your setup to your real hardware dpi. Hardware dpi support is given for qt5 and gtk, but not widely tested yet.

Of course best results if real_dpi is a even multiple of virtual_dpi 

This can be used as a basis for supporting human units (like millimeters, inches, android dip) in config files as well, since these can be converted using virtual_dpi.

This is the second attempt after #810 . But #810 was the wrong attempt do do what is done here.

### DIFF
--- a/navit/attr_def.h
+++ b/navit/attr_def.h
@@ -200,6 +200,8 @@ ATTR(turn_around_penalty)
 ATTR(turn_around_penalty2)
 ATTR(autozoom_max)
 ATTR(nav_status)
+ATTR(virtual_dpi)
+ATTR(real_dpi)
 ATTR2(0x00027500,type_rel_abs_begin)
 /* These attributes are int that can either hold relative or absolute values. See the
  * documentation of ATTR_REL_RELSHIFT for details.
--- a/navit/callback.c
+++ b/navit/callback.c
@@ -23,6 +23,7 @@
 #include "debug.h"
 #include "callback.h"
 
+
 struct callback {
     /* func has variable number of arguments, not (void),
      * but we must declare something... */
@@ -34,6 +35,8 @@ struct callback {
 };
 
 struct callback_list {
+    callback_patch patch;
+    void * patch_context;
     GList *list;
 };
 
@@ -115,6 +118,13 @@ void callback_list_remove_destroy(struct callback_list *l, struct callback *cb) 
     g_free(cb);
 }
 
+void callback_list_add_patch_function (struct callback_list *l, callback_patch patch, void * context) {
+    if(!l)
+        return;
+    l->patch = patch;
+    l->patch_context = context;
+}
+
 void callback_call(struct callback *cb, int pcount, void **p) {
     int i;
     void *pf[8];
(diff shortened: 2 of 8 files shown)


metalstrolch:
We should agree on a proper default value for virtual_dpi, as the usual desktop 96dpi seems to be wrong for most layouts.

metalstrolch:
Ok done and working for qt5, but can be used on all other graphics as well.
What this does: It scales in all calls to the graphics backend the containing coordinates, widths, lengths and font sizes by the rounded factor given by virtual_dpi/real_dpi. It is limited to scale even factors only, as fractional factor provide really ugly results.
Additionally it hooks to the call back list provided by graphics core to the graphics backend. There it unscales all coordinates, widths, lengths and font sizes by the same factor.

This has following result:
- The displayed image get's upscaled.
- Navit uses less resources as it needs to render data for only factor of the screen size.

**To test: add virtual_dpi="100" real_dpi="200" to upscale by 2**

The diff of graphics.c / graphics.h is bigger than required due to the new "astyle" (in)sanity check on CircleCI. Sorry for that.

metalstrolch:
No comment is good comment... Will merge this as would conflict with ongoing work on multipolygons.
Blame me if this breaks anything.

jkoan:
Hi @metalstrolch,
i looked at the sources and they looked good. But until now i cant test it, maybe tomorrow.
But i can say that i really like the work you are doing :smile:

metalstrolch:
Is this because of the broken android build? Seems I'm the only one left not on android.

jkoan:
> Is this because of the broken android build? Seems I'm the only one left not on android.

is't mostly because of time on my side. I also know @hoehnp is also using Navit on non android

metalstrolch:
No stress. I merged this because I want to continue with multipolygon support. Now I'm facing the drawing side of it, and there I want to do the scaling right already. Sorry for my not being patiently.

metalstrolch:
@mvglasow It seems you pushed something direct on master, so merging this via CI from trunk to master failed. Was this on intention? How to restore "master"?

hoehnp:
As @jkoan  mentioned, I use navit both on SailfishOS and sometimes also on Android. @metalstrolch would be great to improve navit on SailfishOS, in particular the TTS. Is the scaling also working on Sailfish?

metalstrolch:
Should work on all platforms. Qt5 on Sailfish even reports hardware dpi. I did this specially for Sailfish. I don't want to require different packages for standard dpi Jolla1 and high dpi Sony Xperia devices. Should be OK now.
Next i plan to add multipolygon support to finally get rivers and lakes shown.
I don't have plans to work on tts, as espeak works for me.

Issue #802, opened 2019-06-21
Android build: libnavit.so missing from resulting apk


lains:
My recent builds for android (using the script `scripts/build_android.sh`) generate an apk that misses libnavit.so.
On a successful build, the apk contains:
unzip -l navit/android/build/outputs/apk/release/android-release-unsigned.apk | grep '\.so.*'
```
   793600  1980-00-00 00:00   lib/armeabi-v7a/libnavit.so
  1203076  1980-00-00 00:00   lib/x86/libnavit.so
```
On recent builds, the apk contains no libnavit.so and the application fails to start in Android, with an error message mentionning that libnavit.so was not found.
It seems this issue was introduced by commit https://github.com/navit-gps/navit/commit/bbff86825991abb7408abf1691118b112c83bed2.
Building succeeds until previous commit https://github.com/navit-gps/navit/commit/b2205d39d2864f4ed7fd3d931b2e6b8f06a6b38b
Or maybe that's just on my build machine, if anyone can give a try...

jandegr:
bbff868 over there seems to contain both libs
https://circleci.com/gh/navit-gps/navit/14377#artifacts/containers/0

jandegr:
but this one contains no lib folder in the apk at all
https://circleci.com/gh/navit-gps/navit/14588#artifacts/containers/0
so I would say it is da3e1e0 instead that causes it

jandegr:
aka 4e5b35a

lains:
@jandegr, yes you're right, it's my mistake.
At bbff868, the shared lib is still included
At 4e5b35a, there is no shared lib, this is easily checked using `unzip -l android-release-unsigned.apk|grep '\.so'`

During compilation, one can spot this quite quickly as one on the last lines during the execution of `scripts/build_android.sh` is:
`[158/158] Linking C shared library ../../../../build/intermediates/cmake/release/obj/armeabi-v7a/libnavit.so
`

I thought the issue was the modification in [navit/CMakeLists.txt](https://github.com/navit-gps/navit/commit/4e5b35a97ea2f2e15b9b5f361ebd3bb9094a4be1#diff-60b4c902ce6303fc9b1194fd0e76c2f3) but reverting only that part of 4e5b35a makes the build fail.

lains:
I tried a few options, but I didn't manage to get libnavit.so back inside the generated apk for Android.

Could someone that is comfortable with cmake have a look?
[...]

PR #816, opened 2019-07-24
fix:core: Correct default layout handling.

metalstrolch:
This commit corrects the default layout handling broken when splitting
the layouts to own files. It restores the layout saving of the internal
gui. So now it starts again with the last selected layout honouring
gui_internal.txt

### DIFF
--- a/navit/main.c
+++ b/navit/main.c
@@ -328,10 +328,6 @@ static void win_set_nls(void) {
 }
 #endif
 
-void main_update_default_layout(struct navit *navit) {
-    navit_update_current_layout(navit, NULL);
-}
-
 void main_init(const char *program) {
     char *s;
 #ifdef _UNICODE		/* currently for wince */
--- a/navit/main.h
+++ b/navit/main.h
@@ -35,7 +35,6 @@ void main_add_navit(struct navit *nav);
 void main_remove_navit(struct navit *nav);
 int main_add_attr(struct attr *attr);
 int main_remove_attr(struct attr *attr);
-void main_update_default_layout(struct navit *navit);
 void main_init(const char *program);
 void main_init_nls(void);
 int main(int argc, char **argv);
--- a/navit/navit.c
+++ b/navit/navit.c
@@ -2551,6 +2551,7 @@ static int navit_set_attr_do(struct navit *this_, struct attr *attr, int init) {
     case attr_layout:
         if(!attr->u.layout)
             return 0;
+        dbg(lvl_debug,"setting attr_layout to %s", attr->u.layout->name);
         if(this_->layout_current!=attr->u.layout) {
             navit_update_current_layout(this_, attr->u.layout);
             graphics_font_destroy_all(this_->gra);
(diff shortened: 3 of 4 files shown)


lains:
Thanks for spotting this.

The split of layout descriptions in their own files was done in https://github.com/navit-gps/navit/pull/559

OK, I didn't know about the existence of an active="yes"/"no" attribute on the `<layout>` tag.
Do you know how this tag is used? In which use case?

I tried to understand what the changes would mean in terms of sequence.
Is my below guess correct?
The changes would modify the way navit determines which default layout to use (whenever the user has not selected any layout yet), following the sequence:
- check active tags for layouts in the order of parsing
- check if the default layout name matches the name for layouts in the order of parsing 

At the first match of either conditions above, select the layout as the default one.
This will be overriden by the user selected layout if any.

metalstrolch:
Truly said I didn't know about the active attribute myself unless I spotted it trying to find the solution for the default layout handling breaking user selection. But looking at it, it was already a way to define the default value back when all layouts were in the same file. With this patch both ways work. Albeit not at the same time. Btw: #559 removed the "cursor" from Car-simple layout as well.

lains:
Do you mean that when the user selects a layout (different from the default layout), it is currently not kept in navit at the next startup, and navits defaults back to the default layout each time?

lains:
You're right, there is no cursor in Car-simple after #599:
https://github.com/navit-gps/navit/blob/e70a28963ff9754d3257361fad93aebf59edee08/navit/navit_layout_car_simple_shipped.xml#L2

but it seems it was already not there before the merge for #599:
https://github.com/navit-gps/navit/blob/b0773218ab67e00e1aa7303829f342dce87c46c0/navit/navit_shipped.xml#L5574

It even seems it was never there: https://github.com/navit-gps/navit/commit/1dd6966649777071595a95c2ad2072ae69f19333

In any case, it is missing, so probably there is no cursor at all displayed. I have to admit I have never used this layout during navigation...

Or maybe I understood incorrectly. Are we speaking about the `<cursor>` tag?

metalstrolch:
Yes, currently its even worse. Navit starts up drawing the first map in user selected layout, then if you move the map or force it otherwise to refresh, it partly switches to default layout keeping the cursor from previous. This reverts correct behaviour.

As for the simple layout, I didn't bisect. If there was no cursor before then sorry for blaming.

Btw. there's something wrong with night/daylayout switching as well . But that was broken before layout split already. Its also possible that this is vehicle dependent.

lains:
Sorry... I couldn't reproduce the issue on my Android build, but I'm confident your code fixes the issue.

metalstrolch:
Is Android using gui-internal at all? I don't have some Android device to check.

lains:
Yes, Android build uses gui-internal.
So far, it properly saves and restores the user-defined layout.
I will only check that it still does (just in case) after these changes, but I don't see why it shouldn't.
But I am waiting for #802 to be fixed before I can easily build for Android again.

Issue #747, opened 2019-02-12
Map display: Cursor and route minor update glitches when not in "follow route" mode


lains:
When navigating, if follow vehicle is dectivated, the current route followed is not removed from the map as the vehicle cursor progresses and there are glitches also on moving the cursor itself (when the vehicle speed is low).

Steps to reproduce:
- start navigation, as a pedestrian
- move the map around to disable "follow vehicle", and orient the map toward north (if needed zoom out a bit to observe the redraw on a longer period)
- progress on the route

The screenshot below illustrates this problem.
- The pink path is not updated to always start from where the cursor is located. If the route is recalculated, though, of course, the path is completely redrawn, as well as when the user moves the map view manually
- One can see that there as multiple (sometimes partially) drawn cursors. Every minute, roughly, all "ghost" cursors are deleted and the view gets cleaner (but the ping route does not get updated)

This is not a blocking issue, as there are ways to work around it (moving the map actually forces redrawing and cleans up the view).
This issue has been there for a while, but it was reproduced on https://github.com/navit-gps/navit/commit/845dd4092ee64bbce0939299793336decafa40b8

![screenshot_2019-02-10-16-30-23](https://user-images.githubusercontent.com/383502/52639254-2332ff80-2ed4-11e9-89f2-ef01a8ac7662.png)

metalstrolch:
What graphics system was this observed on? Drawing the cursor / drawing a layer is highly graphics dependent. #817 may fix something looking different on cursors with too small layer. But here it was observed with a simple circle. Here #817 most probably doesn't fix it.

lains:
The screenshot above was taken on a device running Navit on Android.
From the display, I assume the layout was "Bike"
I'm guessing that this likely happens more often when moving slowly (walking) and for sure, when map is not redrawn each time (thus not in follow vehicle mode, but in map browse mode).

metalstrolch:
OK, I don't have any Android device to check on android. But usually the map in the background including the highlighted route is not redrawn unless dragged around or the cursor leaves the screen. So the highlighted route not being updated is quite OK.

The cursor is draw within it's own layer which is moved on top of the map. This layer has it's own drawing context and is usually transparent except the drawn cursor. So the artefacts are most probably because the transparent layer is not cleaned properly by the graphics back end on re draw.
[...]

PR #817, opened 2019-07-24
fix: vehicle: Increase layer for cursor for rotation

metalstrolch:
This commit increases the layer used for drawing the cursor that the
cursor fits on it regardless its rotation. This is required as some
graphics (e.g. qt5) clip drawings leaving the overlay size. This
cripples the cursor if the layout creator did not mind this. This patch
adds extra space to compensate.


metalstrolch:
What about this? Anybody any remarks?

lains:
No remark from my side, except a minor typo at https://github.com/navit-gps/navit/pull/817/files#diff-e4894e6a6c9c1250eb569166a703bef1R316
cursot->cursor

I don't know this part of the code, so I trust your changes.
Could it be related to https://github.com/navit-gps/navit/issues/747 ?

PR #818, opened 2019-07-25
feature:gui_internal: include post code in town serch

metalstrolch:
This pull request includes postal code or zip code into the town search.


metalstrolch:
Why not just include post code / zip code in town search? Renders #770 obsolete probably.

-Was / is there a reason for not doing this? Performance?

metalstrolch:
Any news here? Still find this very useful, at least in Germany.

aerostitch:
Does that mean that we can't search by town/district anymore on town level and only use zip code?

metalstrolch:
No it doesn't. It just includes postal code matches into the result. So you can start to search with numbers finding postal code. Could imagine this being a problem in countries where postal codes are letters. But even there it should add them correctly to the result.
Stumbled across this by looking at #770 which tried to additionally switch the keyboard to numbers. But if you gave letters the usual name search popped up. So I created this PR for the "simple" way not requiring a HMI change.

metalstrolch:
To be honest I didn't inspect the search code much. So this might even exploit a bug in reality. But at least for me this works quite nice.

Issue #815, opened 2019-07-23
Processing large volumes of TraFF messages massively slows down Navit


mvglasow:
So far, the only real-world use case for traffic in Navit relied on TMC. Even at busy times, there would usually be no more than around 150 active messages, most of which would expire after an hour. This does cause some delays on startup, which can be annoying but is still manageable.

Now I have built another source, which gets its messages from the Internet, as many as 300 in one shot and some 600 active messages. (This can be controlled by enabling only the sources the user requires, but these values are typical for Poland and Lithuania together, needed when traveling in the border area.)

With that volume of data, Navit slows down to the point of being unusable for several minutes after startup. When a feed is received I expect similar effects, unless most of the messages are already in cache with the same location.

I see the following ways to address this:

* Cut down on the time needed for location matching; reduces time to parse a location from scratch (both on receiving a new location and when reading cached messages):
  * Reduce the geographic area of the map to the minimum needed: less data to read from the map, smaller route graph to examine
  * Ignore certain way types (e.g. if the road class is `MOTORWAY`, we may consider ramps as well as trunk and primary roads, but not service roads): smaller route graph to examine
  * Examine only the tile levels needed for that particular way type: less data to read from the map, smaller route graph to examine
* Cache traffic map, with reference to the map from which the data was obtained and to the message. On startup, check which messages have expired, and drop ways no longer associated with an active message. If ways are no longer present in the map from which they were taken, or their geometry has changed, re-parse the locations of the associated messages. This would speed up startup (unless the map has changed) but would not help with new locations.
* Implement multithreading: message processing would run in a separate thread while other things could happen on the main thread. This is currently being worked on. However, every thread will need to access map data (and these operations could result in threads locking each other up), so there are still some big question marks. Also, this will not fully eliminate all performance issues, thus having the first two items will still be beneficial.

mvglasow:
Tested with a feed of 443 messages (with events relevant for Navit so they will not get discarded immediately). Original processing time (on startup) is around 3 minutes.

The geographic area originally was 1000 m around the enclosing rectangle of the endpoints. Reducing that to 100 m shaves a few seconds off the processing time (somewhere around 15 seconds for the whole feed). The padding is necessary to match locations with a `LOW_RES` fuzziness (which is the case for TMC but not here); I therefore now use 1000 m for `LOW_RES`, 100 m otherwise. Some padding will still be needed to allow for bends in the road.

[...]

PR #822, opened 2019-07-28
traffic:Improve startup performance

mvglasow:
To address #815, this PR improves performance of the traffic module by deferring location expansion until the location is actually needed, speeding things up when we are starting up with a large number of cached messages or receiving a large feed.

On startup and when receiving a feed, items for the location are generated immediately only if we are currently routing and the rectangle for the location overlaps with the rectangle for the route, as these messages may influence the route.

Otherwise, when a map rectangle is opened on the traffic map, items are generated for locations whose rectangles overlap the queried rectangle.

When the traffic map is queried with a NULL selection (meaning everything should be retrieved), nothing is expanded and items are returned only for previously expanded locations. This is an experimental feature which became necessary since the traffic module occasionally queries its own map (all of which happens with a NULL pointer and would otherwise cause all locations to be expanded). If the current implementation is found to cause trouble, we would need to constrain these operations to a particular rectangle.

When TraFF messages are exported via Dbus, all messages are expanded as part of the operation—so the export can now take quite long when the message store holds a lot of unresolved messages.

This still needs to undergo some tests. Starting up with a large list of messages and dragging the map around works, but routing and export have not been tested yet.


mvglasow:
Rerouting still has errors when messages are added (removing messages works as before). Freshly calculated routes (after the messages were received) look OK.

mvglasow:
Rerouting works now; the error was an incorrect negation. However, freshly calculated messages now seem to have issues. Exports work.

mvglasow:
Freshly calculated messages now work as well.

mvglasow:
70071e7 fixes an out-of-bounds write. On Android, due to #802, you need to revert 4e5b35a before building, after that it will build into a functioning Android app.

I’ll take it for a spin these days to rule out any unwanted side effects in practice, and then report back.

mvglasow:
Tests for lazy location matching did not reveal any unwanted side effects. Startup is quick, at the expense of long routes taking a long time to calculate if there are a lot of traffic reports within their map selection.

The latest batch of commits adds location caching: traffic map items are written to `traffic.xml` and restored on startup. This works well in lab tests, real-life tests are still pending.

Bugs and limitations (still to be fixed): 

* Speed limits may be incorrect if an item has more than one associated message.
* If map data changes after the location was expanded and cached, this is not detected and the traffic distortion may no longer match the map data.

mvglasow:
And the latest batch of updates also fixes the two bugs and limitations:
* Speed limits now should be handled correctly even for items referenced by multiple messages
* When restoring traffic distortions, Navit now compares them to the map (i.e. all active maps in the mapset) and discards all cached data if it does not match current map data

The only open point is that restoring segmented ways currently does not work—segmented ways have to be regenerated from scratch. This would only be an issue with exotic use cases (binfiles from the map download server do not have segmented ways), and would only negate performance benefits introduces by these commits but not impact accuracy of data.

mvglasow:
A brief test on a real device with real traffic messages looked OK. I will likely be making some heavy use of the new features over the course of the next week and then report back.

mvglasow:
I did a test under real-world conditions today:
* Grabbed all active messages for Poland and Lithuania (some 600, though one-third expires after a few minutes)
* Calculated a route from Munich to Augustów, about all the way across Poland, then shut down Navit
* The next day, opened up Navit and calculated a route in town

Startup and route calculation were still quite slow, which may be in part because restoring some 200–300 locations from cache is still quite time-consuming. It might in fact make sense to defer even restoring of items until a map area containing them is queried.

Also, if a message has cached items but an update for that message is received before the items can be restored, Navit might discard the cached data, negating all performance benefits.

mvglasow:
The latest batch of commits defers restore operations until a rectangle possibly containing the items is queried. That speeds up startup and calculation of short routes; the performance penalty will only strike when calculating a long route in an area with a lot of traffic messages. Probably the best we can get without multithreading. I’ll do some more tests over the course of this week, but I am planning to merge if I don’t encounter any crashes—for a number of use cases this is already a massive performance improvement over what we had before.

mvglasow:
Real-world performance for a route all across Poland was still disastrous; route calculation with expansion of some 300 items took several minutes.

Iterating over the whole map set once for each item we restore turns out to be quite inefficient. The latest commit takes a different approach: parse everything, then iterate over the map and compare anything with matching IDs and mark items when a match is found. If all items are matched, use the data, else discard everything. In the lab, this has cut route calculation from 60 to 14 seconds (tested with a route from the Latvian-Lithuanian border to Hradec Králové, spanning most of Lithuania and Poland).

mvglasow:
Ready for merge, except that CodeFactor is preventing me from doing an automatic merge…

PR #827, opened 2019-08-14
Android build fix and update - to be tested

jandegr:
try to fix https://github.com/navit-gps/navit/issues/802
The apk built on circle now includes tha so's again, but I am a bit short in Android devices for the moment so I invite everybody to help in testing.

further testing to be done for regressions this might introduce in other platforms, I don't expect any but lets make sure. 

adds 64 bit native lib for google play store as well
https://android-developers.googleblog.com/2019/01/get-your-apps-ready-for-64-bit.html
https://developer.android.com/distribute/best-practices/develop/64-bit
> If you do not have any arm64-v8a or x86_64 libraries, you'll need to update your build process to start building and packaging those artifacts in your APK.

so I suppose adding arm64-v8a is enough and it is not mandatory to add x86_64 as well, however that would be no problem if needed.


gefin:
Tested the Build result 3951. It work on Android 7.1.2.
Tested Build 3954 The binary work with opentom.

jandegr:
thx @gefin 

I tested the windows32 build (3952) on windows10_64, found no regressions

@pgrandin can you test if it now complies with the 64 bit requirements of the google play store, thx.

pgrandin:
So it works on the 2 following devices :
 - SM-J600G, android 8.0.0, SDK 26, ABI armeabi-v7a
 - SM-J320H, android 5.1.1, SDK 22, ABI armeabi-v7a

It doesn't work on the following device :
 - Mi A2, android 9, SDK 28, ABI arm64-v8a

I think that this PR address the original intent of #802 (missing lib) and I've created #828 to keep track of this issue.

jandegr:
removed arm64-v8a to narrow this pr down to the issue of the missing so,
and just now some build issue pops up :(

jkoan:
Hi @jandegr
first of all thanks for the fix.
The build issue could be because we don't pin the Docker images of our build images, could you just restart the build on circleci? Or even better the whole pipeline. For me it looks like circleci could have got a problem with some jobs because there status is "expected" here on github.

lains:
I tried to build manually for Android, but I am getting the following compilation error:
```
  In file included from ../../../../../support/wordexp/wordexp.c:15:
  In file included from /usr/include/glib-2.0/glib.h:30:
  In file included from /usr/include/glib-2.0/glib/galloca.h:32:
  /usr/include/glib-2.0/glib/gtypes.h:423:3: error: '_GStaticAssertCompileTimeAssertion_0' declared as an array with a negative size
    G_STATIC_ASSERT(sizeof (unsigned long long) == sizeof (guint64));
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /usr/include/glib-2.0/glib/gmacros.h:232:103: note: expanded from macro 'G_STATIC_ASSERT'
  #define G_STATIC_ASSERT(expr) typedef char G_PASTE (_GStaticAssertCompileTimeAssertion_, __COUNTER__)[(expr) ? 1 : -1] G_GNUC_UNUSED
                                                                                                        ^~~~~~~~~~~~~~~
  1 error generated.

```

lains:
And this comes from the glib headers used by the build process:
During compilation for ../../../../../support/wordexp/wordexp.c, the compiler is called with
`-I/usr/include/glib-2.0`, which uses the local installed glib. If I replace with
`-I../../../../../support/glib`, then the compilation works...

lains:
But probably, this is not coming from this PR, but rather from something else that was broken at commit https://github.com/navit-gps/navit/commit/4e5b35a97ea2f2e15b9b5f361ebd3bb9094a4be1 (there was quite some rework on the glib package detection there...)

jandegr:
since there was some work required to fix the building on circleci, I added arm64-v8a again and applied some other fixes and updated just about everything I could, incl. ndk, in an attempt to get it running on  android 9, SDK 28, ABI arm64-v8a too.

building android does not wait anymore until that sanity thing ends, holding each build until that finishes just wastes too much time.

Navit can have that codefactor thing to play with, but from what i've seen nobody cares for it and it just causes frustration. The issues it reports are generally valid, but like in the case of this pr, a small change in one line suddenly reports over 40 issues it should have seen before.

pgrandin:
Codefactor isn't really a merge blocker if it complains about code that wasn't touched by the PR. 
It's more a way to ensure we don't introduce new issues, and try to fix what it finds whenever possible. I've created https://github.com/navit-gps/navit/issues/832 to track the issues it reported in this PR.

Thanks a lot for your help here @jandegr 

The arm64 specific issue is tracked in https://github.com/navit-gps/navit/issues/828

lains:
Great work, thanks!

I successfully tested on an Android device running LineageOS
LineageOS version: 15.1
Android version: 8.1.0

I had to tweak the build process because of glib dependencies that were mistakenly pointing to the build host rather than the Android toolchain.
I will submit the changes in a separate PR if someone is interested (it does not seem to occur on CircleCI builds, only when manually building on a host where glib, gmodule and/or dbusglib headers are present).

lains:
Tested today on WinCE. Build completed successfully. I had an issue with the resulting archive, but I don't think this is at all related to this PR (copying the navit.exe (generated with this patch on cmake) into an older archive works, so the issue comes from another file.

mvglasow:
@lains I am having issues with the Android build as well (the same ones), see #840. Would you mind sharing how you tweaked the build? Something there is not quite right yet…

mvglasow:
Found it, the build works after deleting all the `.gitignore`d stuff in `navit/android`.

lains:
Indeed, I got the same problem and I described the process I followed [here](https://github.com/navit-gps/navit/pull/833#issuecomment-522291145).
Hopefully, this will not occur anymore to future users now that https://github.com/navit-gps/navit/pull/833 has been merged.

PR #831, opened 2019-08-16
Pr 830

viktorgino:
https://wiki.navit-project.org/index.php/Commit_guidelines

### DIFF
--- a/navit/navit.c
+++ b/navit/navit.c
@@ -2460,6 +2460,27 @@ void navit_set_center_cursor(struct navit *this_, int autozoom, int keep_orienta
 }
 
 /**
+ * @brief Drags (moves) the map
+ *
+ * Drags (moves) the map from origin point to the destination point
+ *
+ * @param navit The navit instance
+ * @param origin The point point where the drag starts
+ * @param destination The point where to map should be dragged to
+ * @returns nothing
+ */
+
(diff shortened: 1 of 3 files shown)


pgrandin:
I think that the build_android failure is related to https://github.com/navit-gps/navit/pull/827/files#diff-f940ede166f88341829e50c9c345047b

viktorgino:
@pgrandin should be related to this: https://github.com/navit-gps/navit/issues/834

navit/navit.c
```diff
@@ -2459,6 +2459,27 @@ void navit_set_center_cursor(struct navit *this_, int autozoom, int keep_orienta
...
+ * Drags (moves) the map from origin point to the destination point
+ *
+ * @param navit The navit instance
+ * @param origin The point point where the drag starts
+ * @param destination The point where to map should be dragged to
+ * @returns nothing
+ */
+
+void navit_drag_map(struct navit *this_, struct point *origin, struct point *destination) {
+    update_transformation(this_->trans, origin, destination);
```
metalstrolch: Since this shall get a "public" function, I suggest to do some sanity checking here, like aborting if some of the parameters is NULL, issuing some err message if this is not expected. May help future users of this in calling the function right.

PR #833, opened 2019-08-17
Fix: Fix pkgconfig detection, and multiple subsequent build for Android

lains:
This PR brings 2 changes for the build for Android platforms:

1. It fixes rename of the filenames (have to be lowercase on Android) so builds work even if it is run multiple times (today, only the first build works, further builds fail because rename itself fails, as renamed lowercase files still exist in the previous build working directories)

2. It changes the way the build process uses pkgconfig.
Indeed glib and ezxml dependencies are checked by default against the build host, not against the toolchain Android libs.
It seems that this started from https://github.com/navit-gps/navit/commit/4e5b35a97ea2f2e15b9b5f361ebd3bb9094a4be1.
This issue does not seem to occur on CircleCI builds, but seem to occur when manually building on a host where glib, gmodule, dbusglib  and/or ezxml headers are present.

   Symptom of the wrong libs picked up is an error during compilation:
```
  In file included from ../../../../../support/wordexp/wordexp.c:15:
  In file included from /usr/include/glib-2.0/glib.h:30:
  In file included from /usr/include/glib-2.0/glib/galloca.h:32:
  /usr/include/glib-2.0/glib/gtypes.h:423:3: error: '_GStaticAssertCompileTimeAssertion_0' declared as an array with a negative size
    G_STATIC_ASSERT(sizeof (unsigned long long) == sizeof (guint64));
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /usr/include/glib-2.0/glib/gmacros.h:232:103: note: expanded from macro 'G_STATIC_ASSERT'
  #define G_STATIC_ASSERT(expr) typedef char G_PASTE (_GStaticAssertCompileTimeAssertion_, __COUNTER__)[(expr) ? 1 : -1] G_GNUC_UNUSED
                                                                                                        ^~~~~~~~~~~~~~~
  1 error generated.
```

The source can then be spotted inside the -I options of the compiler's command line.
During compilation for ../../../../../support/wordexp/wordexp.c, the compiler is called with
`-I/usr/include/glib-2.0`, which uses the locally installed glib. The issue is fixed if this argument is replaced by: `-I../../../../../support/glib`


jandegr:
Hi,
I tried to confuse cmake on cirlceci by installing libdbus-glib-1-dev libgtk2.0-dev

and indeed, it goes wrong on circleci too, so it is not just some local issue in your setup


> debug|armeabi-v7a :--->>> speech
debug|armeabi-v7a :Enabled   android ( Android detected )
debug|armeabi-v7a :Disabled  cmdline ( Android detected )
debug|armeabi-v7a :Enabled   dbus ( dbus-glib-1 found )
debug|armeabi-v7a :Disabled  espeak ( Default )
debug|armeabi-v7a :Disabled  iphone ( Default )
debug|armeabi-v7a :Disabled  qt5_espeak ( Qt5 multimedia not found )
debug|armeabi-v7a :Disabled  speech_dispatcher ( speech_dispatcher lib not found )
debug|armeabi-v7a :--->>> support
debug|armeabi-v7a :Disabled  espeak ( Default )
debug|armeabi-v7a :Disabled  ezxml ( native Glib found )
debug|armeabi-v7a :Enabled   gettext_intl ( native libintl missing )
debug|armeabi-v7a :Disabled  glib ( native Glib found )
debug|armeabi-v7a :Disabled  libpng ( Android detected )
debug|armeabi-v7a :Enabled   shapefile ( Default )
debug|armeabi-v7a :Enabled   wordexp ( native wordexp missing )
debug|armeabi-v7a :Disabled  zlib ( native zlib found )
debug|armeabi-v7a :--->>> traffic
debug|armeabi-v7a :Enabled   dummy ( Default )
debug|armeabi-v7a :Enabled   null ( Default )
debug|armeabi-v7a :Enabled   traff_android ( Android detected )
debug|armeabi-v7a :--->>> vehicle
debug|armeabi-v7a :Enabled   android ( Android detected )
debug|armeabi-v7a :Enabled   demo ( Default )
debug|armeabi-v7a :Disabled  file ( Android detected )

However, the solution you propose might work (did not test that), but it removes some flexibility by hardcoding some choices.

context of the remark : for quite a while I build for windows in 2 flavors, classic with glib from support or I build a dll for glib first and then cmake detects it and skips the stuff from support. Just now I wanted to start to port that dual system over to android, but with your solution cmake won't ever detect a prebuilt glib for android and keep building the stuff from support.

jandegr:
Hi,

adding
`export PKG_CONFIG_LIBDIR="/opt" # here to prevent finding regular linux libs`
to the buildscript fixes the confusion I introduced in circleci

maybe we could add
`export PKG_CONFIG_PATH="" # here to prevent finding regular linux libs`
to be on the safe side

If you build smth. for linux again on your box you will have to reinitialize those paths

I don't know enough about cmake to find the actual cause, maybe somone else can.

lains:
Hi @jandegr, I tried your proposal and it does the trick for me:
Adding `export PKG_CONFIG_LIBDIR=""` in scripts/build_android.sh just before the line that runs cmake.
Removing file `CMakeCache.txt`
Cleaning up the folder `navit/android` (to pristine state)
And running scripts/build_android.sh once more.

lains:
@jandegr, I have implemented your proposal, as it works wellfor my setup and it would be a good thing that this is done automatically during build.
In case someone still wants to detect specific libs for the target, providing a non-empty value for PKG_CONFIG_LIBDIR should work.
Also, this is outside of cmake so it only affects building for Android using scripts/build_android.sh
I also included a fix for subsequent builds for Android.
The first build succeeds, but subsequent builds fail (except if a cleanup is done manually on files renamed for Android).
I fixed the call to rename so that the cleanup is done automatically before each build.

lains:
Oops, there was still code on CMakeLists.txt changes in this PR (these diffs were not needed anymore, the implementation has been done another way in the build_android.sh script).
Because this PR has been merged already, I am going to revert on trunk the extra changes that were not supposed to be merge.

PR #837, opened 2019-08-21
Issue/826

gefin:
In bad case the function iso8601_to_secs(time_attr.u.str); is called with a NULL pointer.
The function dont check this...
The bad call of iso8601_to_secs is in navit/track.c line 683.

vehicle_get_attr(tr->vehicle, attr_position_time_iso8601, &time_attr, NULL));
was removed at
https://github.com/navit-gps/navit/commit/03f3aa8637fff3e8e09d3bde21b28654454011fd#diff-3cbc3b1f62b4a7f91ff8ea7691e63e64


pgrandin:
Nice! Adding issue link for easy reference : https://github.com/navit-gps/navit/compare/issue/826

PR #839, opened 2019-08-21
 Fix:toolchain:Remove privileged operation from build script

mvglasow:
As mentioned in #827, this PR moves privileged operations (`sudo apt-get install`) out of the build script, replacing it with a check if the required component is installed (if it is not, the script will abort with an error).

For local builds, you need to install `ant` yourself. If you haven’t done so before starting the build, the script will abort with a meaningful error message.

For CircleCI, `ant` has been added to the list of packages to install.

On F-Droid this will be a separate merge request (Link will follow).


mvglasow:
@jandegr could you take a look at this as well, just to make sure I am not inadvertently breaking anything?

mvglasow:
And the related F-Droid merge request is at https://gitlab.com/fdroid/fdroiddata/merge_requests/5289

pgrandin:
I think that it makes sense :+1:

scripts/build_android.sh
```diff
@@ -23,9 +23,9 @@ export JVM_OPTS="-Xmx3200m"
 export GRADLE_OPTS='-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
 
 # processing xml is messed up a bit after the original introduction of gradle
-# so a useless install of ant here even if using gradle/ninja
+# so require a useless install of ant here even if using gradle/ninja
 
-sudo apt-get install -y ant
+which ant > /dev/null || { echo "FATAL: ant is not installed; install manually and retry."; exit 1; }
```
lains: Maybe this line good be written in a way that is more compatible with generic shell syntax like:
```
if which ant >/dev/null; then
    echo "FATAL: ant is not installed; install manually and retry." >&2
    exit 1
fi
```

mvglasow: This will fail if ant is found and succeed if it is not found (just tried it). I am not a shell guru, though; all I know is that if success/failure/true/false/`if` is highly confusing.

Redirecting to stderr is a good point, though.

lains: Mmm... OK, I missed the fact that there is an issue because of the mix of if then construct and set -e

Issue #828, opened 2019-08-15
navit crashes on android 9, SDK 28, ABI arm64-v8a


pgrandin:
backtrace :

```
F DEBUG   :
F DEBUG   : backtrace:
F DEBUG   :     #00 pc 000000000009336c  /system/lib64/libc.so (pthread_mutex_trylock)
F DEBUG   :     #01 pc 000000000010dd34  /data/app/org.navitproject.navit-0OKjHlPdos3Q0izkG9HlVg==/lib/arm64/libnavit.so
F DEBUG   :     #02 pc 000000000010da58  /data/app/org.navitproject.navit-0OKjHlPdos3Q0izkG9HlVg==/lib/arm64/libnavit.so
F DEBUG   :     #03 pc 000000000010cb04  /data/app/org.navitproject.navit-0OKjHlPdos3Q0izkG9HlVg==/lib/arm64/libnavit.so
F DEBUG   :     #04 pc 000000000010c63c  /data/app/org.navitproject.navit-0OKjHlPdos3Q0izkG9HlVg==/lib/arm64/libnavit.so (g_slice_alloc+144)
F DEBUG   :     #05 pc 0000000000108a88  /data/app/org.navitproject.navit-0OKjHlPdos3Q0izkG9HlVg==/lib/arm64/libnavit.so (g_hash_table_new_full+36)
F DEBUG   :     #06 pc 0000000000108a54  /data/app/org.navitproject.navit-0OKjHlPdos3Q0izkG9HlVg==/lib/arm64/libnavit.so (g_hash_table_new+40)
F DEBUG   :     #07 pc 00000000000339dc  /data/app/org.navitproject.navit-0OKjHlPdos3Q0izkG9HlVg==/lib/arm64/libnavit.so (atom_init+24)
F DEBUG   :     #08 pc 0000000000087018  /data/app/org.navitproject.navit-0OKjHlPdos3Q0izkG9HlVg==/lib/arm64/libnavit.so (main_real+112)
F DEBUG   :     #09 pc 00000000000aa310  /data/app/org.navitproject.navit-0OKjHlPdos3Q0izkG9HlVg==/lib/arm64/libnavit.so (Java_org_navitproject_navit_Navit_NavitMain+904)
```

crashdump:
```
D Navit   : **1**A android.intent.action.MAIN
D Navit   : **1**D null
D Navit   : lang=en
D Navit   : Country1 US
D Navit   : Country2 US
D Navit   : Language en
[...]

PR #851, opened 2019-08-29
fix android 64

jandegr:
The first step in fixing https://github.com/navit-gps/navit/issues/828

Glib from support already worked for 64 bits on windows but just needed a small tweak to get it going on android 64 as well.
Before fixing this I tested it with a glib built from latest glib sources (glib-2.61.2) to get a view of what other work there is to do in navit to fix it for arm64 but it is too much to build libglib-2.0.so just for distribution of navit apk's.

After this it still crashes on arm64 as there are a bunch of fixes to follow in the android specific navit code, but I thought to commit this first seperatly so it can bet tested for regressions on other platforms before committing the rest of the fixes.

The rest of the fixes are in android code, so those can't have sideffects on other platforms.

tested with a version including the most essential fixes in the android code on
android 8.1 arm64-v8a
android 9 x86_64

### DIFF
--- a/navit/support/glib/fake.h
+++ b/navit/support/glib/fake.h
@@ -1,4 +1,4 @@
-#include "../config.h"
+#include "config.h"
 #ifndef HAVE_API_WIN32_BASE
 #define USE_POSIX_THREADS 1
 #endif
--- a/navit/support/glib/fake.h
+++ b/navit/support/glib/fake.h
@@ -1,4 +1,4 @@
-#include "config.h"
+#include "../config.h"
 #ifndef HAVE_API_WIN32_BASE
 #define USE_POSIX_THREADS 1
 #endif
@@ -6,6 +6,7 @@
 #include <pthread.h>
 #endif
 #include "debug.h"
+#include "gtypes.h"
 
 #define g_return_if_fail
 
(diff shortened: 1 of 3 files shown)


jkoan:
Hi @jandegr 
I had finaly time to test this fix, for me its not working right now.
I run a Logcat und did a `grep -i navit` on it. Hers the output:

[navit_log_pr_851.log](https://github.com/navit-gps/navit/files/3612703/navit_log_pr_851.log)

jandegr:
@jkoan 
As stated in the opening post it still should have crashed, thx for confirming that it crashes not in glib anymore but now a lot further in the anroid code exactly where it should so that is a valid testresult for me and it confirms your phone can actually run 64bit apk's, otherwise it would have loaded the armv7 binary just as before and run fine.

Again as stated in the opening post, I posted this small snippet from  https://github.com/navit-gps/navit/pull/863 for in case somebody wants to test for regressions on other posix platforms that use glib from support.

The windows port does not use posix threads and I suppose it is the same for winCE, Tomtom builds its own glib, so that excludes those 3 but maybe sailfish is to be tested ?

thx for testing

jandegr:
apparently sailfish provides glib, so I can close this I suppose

PR #854, opened 2019-08-29
fix crosscompilation without PkgConfig

jandegr:
Since https://github.com/navit-gps/navit/commit/4e5b35a97ea2f2e15b9b5f361ebd3bb9094a4be1 a dependency on PkgConfig creeped in for crosscompiling.

It is typical not to have gtk stuff installed if not compiling for linux, so PkgConfig isn't there either.

I fixed it in the first way that came to mind, anybody feel free to change.


jandegr:
closing this since I needed it to work on https://github.com/navit-gps/navit/pull/877 and as such is included there.

Issue #853, opened 2019-08-29
The certificate for maps.navit-project.org expired on 05 June 2016.


jandegr:
The title says it all.

background : tried to access maps over https because of

` java.io.IOException: Cleartext HTTP traffic to maps.navit-project.org not permitted`

this is since android 8
There are solutions to work around this restriction but I report the issue in the title anyway.

jkoan:
Thx for Reporting, adding @pgrandin since i dont have access to the servers. Hopefully we can add letsencrypt certs to those servers.

PR #856, opened 2019-08-30
android:fix mapdownload

jandegr:
Errormessage on android 9 when downloading a map as mentioned in https://github.com/navit-gps/navit/issues/853
`E/org.navitproject.navit.NavitMapDownloader: Error reading from server: java.io.IOException: Cleartext HTTP traffic to maps.navit-project.org not permitted`

When Google released Android Marshmallow (API level 23), it provided a configuration to disable clear text traffic, which would prevent your app from making clear HTTP requests. With the release of Android Pie (API level 28), clear text traffic is disabled by default.

This PR enables cleartext traffic.


jandegr:
closing because included in https://github.com/navit-gps/navit/pull/877

Issue #847, opened 2019-08-28
Add administrative divisions for Lithuania (and several other countries)


mvglasow:
Lithuania currently does not have any administrative division associated with it. As a result, if I want to go to Trakai (the place with the island castle) and enter LT/Trakai, I get two pages of places named Trakai and have only the coordinates to figure out which is the right one.

If we had administrative area information for Lithuania, then the additional information for the “correct” Trakai would look like this:

Trakų seniūnija, Trakų rajono savivaldybė, Vilniaus apskritis

A brief check shows me that none of the Baltic and Scandinavian states have this information.

mvglasow:
See 600edcad for an example

jandegr:
from IRC
> [21:27:01] <mvglasow> and if I were to change that to, say, "3c5678", will it cause admin level 4 to be interpreted for the country label but leave the rest as is?

Hi,
Many years ago I figured this out for a few countries and from memory I recall c denotes county, not country. 
More detail is found there
https://github.com/navit-gps/navit/blob/trunk/navit/maptool/osm.c#L2030

You would have to test it, but according to https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dadministrative#10_admin_level_values_for_specific_countries "3c5678" would solve your issue by adding county name in the search results in gui internal, that is at least if both the Trakai's are in a defferent county.

Poland is a good sample for you https://github.com/navit-gps/navit/commit/600edcadba89b998cd1053597cec1bd63a41946d#diff-47a43ee0434c1ca3debb4da277badb12

mvglasow:
argh, typo, I meant county, of course, not country :-) In the meantime I tested that and found out the same. See changes in #859 for a documentation of what I understand each character does, and please do correct me if I am wrong.

metalstrolch:
Oha, github is quite smart and auto closed this because of merge of #859.

PR #859, opened 2019-08-30
Admin levels for additional countries

mvglasow:
Addresses #847 for the Baltic states.

This may not yet be perfect but is better than what we have now (which is just the town name with no further information). Improvements are welcome!


mvglasow:
I have added a few more countries based on information from https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dadministrative#10_admin_level_values_for_specific_countries: we now cover most of the EU as well as Iceland, Belarus, Serbia, Georgia, Turkey, India, Japan, both Koreas, New Zealand, Canada, Mexico and Brazil. Of these, I have only given Czechia and Austria a quick test. However, none of these changes should introduce any major issues: at the worst, we still won’t get any useful information underneath the town name. Not all of them may be perfect, but this can only be better than what we had before, and anyone who knows better is welcome to suggest improvements!

metalstrolch:
Merging this. Looks fine and improves usability a lot for those countries. Fixes #847

PR #860, opened 2019-09-02
Fix:Core:Fix crash

jandegr:
fixes a crash in the following scenario

click a point on the map
set it as position with gui_internal
click the button that centers the map around your position


metalstrolch:
Unable to merge this, as branch is outdated, but @jandegr doesn't give push permission on his branch.
@jandegr please update (and merge) your branch.

jandegr:
I gave permissions, but why not use this one as a testcase for https://github.com/navit-gps/navit/issues/868

thx.

jkoan:
> I gave permissions, but why not use this one as a testcase for #868
> 
> thx.

if the commit 43485aee2deaee06958a3465f74923f4aa199405 would follow the commit guidelines i could "rebase and merge" directly, but now i would need "Squash and merge" to give a new commit message

jandegr:
Indeed, I just could fix the title to comply with https://navit.readthedocs.io/en/latest/development/commit_guidelines.html
I can't change the commit message myself anymore either, will close this one and open a new one for it.
But why was this not done correctly for so many of those other commits in trunk if it is so simple to deal with  https://github.com/navit-gps/navit/issues/868

PR #861, opened 2019-09-03
Add:graphics/sdl: Polygon with holes support.

metalstrolch:
Add polygon with holes support to SDL graphics.

### DIFF
--- a/navit/graphics/sdl/graphics_sdl.c
+++ b/navit/graphics/sdl/graphics_sdl.c
@@ -319,42 +319,10 @@ static void draw_polygon_with_holes (struct graphics_priv *gr, struct graphics_g
 }
 
 static void draw_polygon(struct graphics_priv *gr, struct graphics_gc_priv *gc, struct point *p, int count) {
-    if ((gr->overlay_parent && !gr->overlay_parent->overlay_enable) || (gr->overlay_parent
-            && gr->overlay_parent->overlay_enable && !gr->overlay_enable) ) {
-        return;
-    }
-
-    Sint16 *vx, *vy;
-    Sint16 x, y;
-    int i;
-
-    vx = alloca(count * sizeof(Sint16));
-    vy = alloca(count * sizeof(Sint16));
-
-    for(i = 0; i < count; i++) {
-        x = (Sint16)p[i].x;
-        y = (Sint16)p[i].y;
-        vx[i] = x;
-        vy[i] = y;
-
-        dbg(lvl_debug, "draw_polygon: %p %i %d,%d", gc, i, p[i].x, p[i].y);
-    }
-
-    if(gr->aa) {
-        raster_aapolygon(gr->screen, count, vx, vy,
-                         SDL_MapRGBA(gr->screen->format,
-                                     gc->fore_r,
-                                     gc->fore_g,
-                                     gc->fore_b,
-                                     gc->fore_a));
-    } else {
-        raster_polygon(gr->screen, count, vx, vy,
-                       SDL_MapRGBA(gr->screen->format,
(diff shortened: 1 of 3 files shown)


metalstrolch:
@jkoan: you can retest multi polygon support with SDL. I added the support for it.

metalstrolch:
Can somebody check this on tomtom as well? Seems tomtom uses sdl.

gefin:
Yes, i can test tomtom.. 
Is there a source to download a matching map ( ~300-800 MB )

metalstrolch:
I've created one: http://metalstrolche.de/bayern.bin
Covers Bavaria / Germany. Sample map Munich is contained within it.
I will edit this post when I take it down again.

metalstrolch:
You may want to try the map contained there as well. Center Navit to Munich before loading this map. You should see some shapes. Some of them should have holes.
[fake-munich.zip](https://github.com/navit-gps/navit/files/3580005/fake-munich.zip)

gefin:
Real map look good. Multi polygons are drawn.
The Fake-Munich stay blank.
Is there a debug line to grep for?

metalstrolch:
Yes the size is OK, there are only few polygon shapes inside. That's no "real" map.
Of course you can check if the map got loaded at all. But I think you just miss the not that big elements.
They are roughly at 48.1333N 11.5666E. You can enter these coordinates in Actions->Coordinates in internal GUI.

gefin:
All blank. Map are loaded. May be my Layout don't show this type of object.
Are also normal objects on the map?

metalstrolch:
The objects are all of type poly_landfill. You can enable them by adding
```
      <itemgra item_types="poly_landfill" order="10-">
         <polygon color="#999893"/>
         <text text_size="5"/>
      </itemgra>
```
To your layout. There are no other objects inside the fake map. Maybe I do one with some POIs and different landuse tags tomorrow.

metalstrolch:
If you have an old version without this patch, you still will see the polygons, but the holes are filled as well.

gefin:
Yes with the "landfill" i see a polygon (6 corner) with a hole (3 corner) and a square with a hole  and a polygon in this hole.
look like it work ;-)

metalstrolch:
Seems we're ready to go.

gefin:
After driving ~100km with this version on tomtom 730 i found a location where the performance drop strong. Time for one screen update is then up to 4 sec. Things like unexpected "turn if possible" happens.
The problem exist only with car layout. Car-simple layout 0.3 sec is good. 

Location is 48°8'24'' N    12°1`50`` E  along the B12
zoom=32
I dont see a problematic object at this location.

metalstrolch:
No, it's not because of replacing draw_polygon() by draw_polygon_with_holes() as this works exactly the same as before if something doesn't have holes. The new version should be even faster than before as the coordinates are memcopied one time less.

It's missing clipping for polygons with holes I think, together with the sheer size of some polygons.

Seems this happens because of two really large polygons with many holes having a quite high node count each:
https://www.openstreetmap.org/relation/2892951
and
https://www.openstreetmap.org/relation/68363
At the location 48°8'24'' N 12°1`50`` E along the B12.

Currently core doesn't implement clipping for polygons with holes. It does for polygons without holes.
And the WindowsCE (and SDL) graphics do not implement clipping as well.
So the polygons are completely raycasted and drawn even if only small portions of them are really on screen.

I'll try two things:
1. Port the "clipping" version of polygon handling in graphics.c to support holes as well. I've left a TODO there anyways.

2. Check  layer geometry on Windows / SDL graphics to only raycast lines which are actually on layer.

Opened #869 as this pull request is already closed.

PR #862, opened 2019-09-03
Add:graphics/windows: polygons with holes for windows CE

metalstrolch:
Add polygon with holes support for windows CE

### DIFF
--- a/navit/graphics/win32/graphics_win32.c
+++ b/navit/graphics/win32/graphics_win32.c
@@ -829,8 +829,177 @@ static void draw_polygon(struct graphics_priv *gr, struct graphics_gc_priv *gc, 
 
 #if HAVE_API_WIN32_CE
 /*
- * Windows CE doesn't support PaintPath used for other versions. No polygon with holes support for CE yet.
+ * Windows CE doesn't feature GraphicsPath, so in order to draw filled polygons
+ * with holes, we need to resort on manual raycasting. The following functions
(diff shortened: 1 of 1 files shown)


metalstrolch:
Untested on real windows CE device!
Needs to be confirmed working before merge!

metalstrolch:
See  #861 for preprocessed map of Bavaria, as well as a test map showing only some test polygons.

gefin:
search for Prien to find the location. 
It is a bit slow, but work.

metalstrolch:
OK. Great to see it's working at least, and the drawing looks OK as far as I can see. It only places the zoom buttons wrong, but this is unlikely introduced by this change.

About the speed: is it actually slower than before introducing the multipolygons? The device is not really a rocket.

About memory usage: I know about this being an issue on devices with only 64MB memory. But not specially introduced by this pull request. Feel free to open another issue about this topic.

metalstrolch:
OK. Going to merge this.
I think for devices like this it would be best idea to process a map throwing out everything one doesn't want to render anyway. While being done quite easy with maptool, we don't have an infrastructure to provide such maps yet. Additionally we could provide a map with all the rest. So one could have two binfiles. One for the roads and routing things. And another for the eyecandy like landuses and so forth. Then we could enable or disable that via the maps menu. Or opt not to have the second one at all. Maybe I'll give that aproach a try. Stay tuned.

PR #864, opened 2019-09-05
update:doc: Add note about the partnershipt with short-circuit.com

aerostitch:
While going through the pages of our old wiki to migrate it I found in https://wiki.navit-project.org/index.php/Changelog that we have or had a partnership with short-circuit.
If it's still current we might want to advertise that in our readme.
Not sure who knows about the status of that partnership.


jandegr:
Navit is stil mentioned there https://www.short-circuit.com/support/links.php6
so somebody should volunteer to buy smth. there and try to get the 5% discount or mail them to inquire.

mvglasow:
Besides, the link on short-circuit.com still points to Sourceforge.

jkoan:
For me the question remains which type "partnership" this should be. 
Yes, a little discount for navit users is nice, but looking at the shop from my needs/perspective I don't see anything that is worth buying. Not even to mention the shop seems to be really old and not totally trustworthy.
So is there any benefit for the Navit project itself?

pgrandin:
For the Navit project itself, no real benefits. We won't be getting back any cut from the benefits. For our users, a deal could be nice, although I wonder how many times this has been used in the past.
Regarding trustworthiness, I have purchased from that shop myself in the past (this is where the deal comes from). The deal is probably more interesting for users building carpc.

I could reach out to them to get an update and get a sense of how many times our users purchased from that shop.

jkoan:
Any update on this?

jkoan:
Closing this for now

PR #865, opened 2019-09-05
update:doc: Add a tip on about a commit-msg hook [skip ci]

aerostitch:
As I'm adding some doc I realize that we're spending a lot of time on CI builds for stuffs that don't affect the builds so I made a `commit-msg` git hook for myself and thought that it might be useful to have it in the documentation in case someone wants to use it too without having to reinvent the wheel.

I made it in python so that it's portable on several platforms.
It's also compatible python 2 and 3.


jkoan:
Normaly perfect to save CI time, but there are two concerns:
1. spellcheck won't run
2. Github needs those statuses to pass the checks, earlier on GitHub this wasn't a Problem, but since Github somehow Cashes the "Expected" results, it is.

aerostitch:
Alright. Closing then.

Issue #869, opened 2019-09-07
Missing clipping for polygons with holes leads to slow rendering performance on Windows CE and SDL


metalstrolch:
Currently the core is lacking clipping support before handing over polygons that contain holes (attr_hole) to graphics back end. As Windows CE and SDL are rendering by manually raycasting polygons (with or without holes), and do not support clipping as well, rendering performance is poor when the screen shows many small portions of large polygons.

This was reported in #861 by @gefin first.

metalstrolch:
Seems this happens because of two really large polygons with many holes having a quite high node count each:
https://www.openstreetmap.org/relation/2892951
and
https://www.openstreetmap.org/relation/68363
At the location 48°8'24'' N 12°1`50`` E along the B12.

Currently core doesn't implement clipping for polygons with holes. It does for polygons without holes.
And the WindowsCE (and SDL) graphics do not implement clipping as well.
So the polygons are completely raycasted and drawn even if only small portions of them are really on screen.

I'll try two things:
1. Port the "clipping" version of polygon handling in graphics.c to support holes as well. I've left a TODO there anyways.

2. Check  layer geometry on Windows / SDL graphics to only raycast lines which are actually on layer.

gefin:
you are right. The location B12 is fast and smooth drawn with poly_wood removed from layout.
With a active polygon with hole the slowdown begins a screen width away from the polygon. 
Thanks for explain.

metalstrolch:
#870 should fix this.
@gefin: can you give it a try?

gefin:
Tested Build16731 Tomtom730 . Polygons with Holes work very well. The delay is gone. Thanks 
Thursday real drive...

metalstrolch:
Will wait for Thursday before merging #870

gefin:
Build16731 work well while real driving. No lags or other problems.

metalstrolch:
Fixed in #870

PR #870, opened 2019-09-10
fix:core:add clipping to polygons with holes

metalstrolch:
As some simple graphics like SDL or windows CE do not clip on drawing, the rendering performance of huge (not necessary in node count but on screen) polygons with holes cause poor drawing performance even if only small portions of those polygons are on screen.

To fix this we clip polygons with holes before calling back end to draw like done with traditional polygons as well.

### DIFF
--- a/navit/graphics.c
+++ b/navit/graphics.c
@@ -2137,6 +2137,105 @@ static void poly_intersection(struct point *p1, struct point *p2, struct point_r
 }
 
 /**
+ * @brief clip a polygon inside a rectangle
+ *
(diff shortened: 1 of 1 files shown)


metalstrolch:
I will ignore CodeFactor once more, as it is totally stupid to implement, and test code, and then on creating the pull request being told by CodeFactor it was too complex. **I will not rewrite tested code because of this.** It's hard to guess the metric of CodeFactor applies.

metalstrolch:
Confirmed working in #869 by @gefin

metalstrolch:
Merging this as confirmed to be fine in #869

PR #872, opened 2019-09-10
Add watch after reopen the gps file

gefin:
vehicle_file_close(priv) disables watch of the file.
vehicle_file_open(priv) dont reenable it. So the gps position was frozen

In my case it is suspend/resume which lead to GPS the reopen.


pgrandin:
Thanks @gefin !

Issue #823, opened 2019-07-30
tomtom_plugin and tomtom_minimal dont contain a navit.xml


gefin:
the circleci build Job 15587  tomtom_plugin and tomtom_minimal dont contain a navit.xml.

also the locales Path is wrong.
The locales are stored in
navit/share/locales/locales
instead of
navit/share/locales

gefin:
I think navit.xml is missed because the build environment dont contain a XSLT processor: 
-------------------------------------------------------
CMake Warning at CMakeLists.txt:591 (message):
  No XSLT processor available.  You have to configure navit.xml yourself, or
  install an XSLT processor (supported:
  saxonb-xslt;saxon;saxon8;saxon-xslt;xsltproc;transform.exe).
-------------------------------------------

gefin:
After testing Pull request #875 i found that the added navit.xml is the navit_shipped.xml

I think we need to install a xslt processor to the build environment.
We should also copy the separate layout xml files. The pull copy only navit.xml to the target.

PR #875, opened 2019-09-18
fix:tomtom: Make sure the tomtom_plugin and tomtom_minimal contain navit.xml

gefin:
the circleci build Job 15587 tomtom_plugin and tomtom_minimal dont contain a navit.xml.
also the locales Path is wrong.
Closes: #823


gefin:
With this patches i get a tomtom480.xml, but the XML is not for Tomtom.
Graphics are gtk, but should be SDL.
The Resolution is also wrong.
 
How to change?

lains:
It seems for tomtom, there is a [dedicated XLST](https://github.com/navit-gps/navit/blob/trunk/navit/xslt/tomtom.xslt) transformation that is applied [during the build](https://github.com/navit-gps/navit/blob/6a525367ae7e5adc466e7935be1b51f5e8bb4e68/scripts/build_tomtom_minimal.sh#L33).

I had a similar issue before for platform WinCE, and it was because no proper xslt processor was found during the build from CircleCI...
From the circleCI output for tomtom_minimal (below), it seems its is the same issue:
```
CMake Warning at CMakeLists.txt:589 (message):
  No XSLT processor available.  You have to configure navit.xml yourself, or
  install an XSLT processor (supported:
  saxonb-xslt;saxon;saxon8;saxon-xslt;xsltproc;transform.exe).
```

lains:
OK, I just read in the separate issue (https://github.com/navit-gps/navit/issues/823#issuecomment-533221586) that you had already found out.

gefin:
Yes, but i dont know how to add xslt support to ci.
A second problem is that the layouts now in extra files. In my home environment this layout files are empty.

lains:
The docker image (navit/tomtom-ng) used for build by circleci is specified here:
https://github.com/navit-gps/navit/blob/trunk/.circleci/config.yml#L182

So, I guess either the docker image should be updated to include an xslt processor, or maybe an easier way is to add a step that installs the right tool during the setup, similarly to what I did here:
https://github.com/navit-gps/navit/blob/trunk/scripts/setup_wince.sh#L7

gefin:
I need help with this patch request.
How to remove the last two commits.
I was not able to add xslt support. sorry.

lains:
Hi @gefin, the easiest way to undo your last two commits is to do a git revert on these commits.

lains:
CircleCI seems to only call [scripts/setup_common_requirements.sh](https://github.com/navit-gps/navit/blob/trunk/.circleci/config.yml#L208).

Changing setup_common_requirements.sh will probably have side effects, so you're right, your changes probably should go into [setup_tomtom_requirements.sh](https://github.com/navit-gps/navit/blob/trunk/scripts/setup_tomtom_requirements.sh).

And then, I would assume that this setup_tomtom_requirements.sh should be executed by CircleCI right after running setup_common_requirements.sh (this requires adding a new line inside https://github.com/navit-gps/navit/blob/trunk/.circleci/config.yml)

lains:
Tyring to implement what I suggested [here](https://github.com/navit-gps/navit/pull/875#issuecomment-533979936)...

lains:
Now xslt seems to be applied... but running the whole setup_tomtom_requirements.sh seems overkill as it downloads and builds plenty of tools that are probably already in the docker image used for build.
So the really only missing step here was:
`apt-get install -y xsltproc`

lains:
Moving the xslt install command inside circleci (this avoids running the full https://github.com/navit-gps/navit/blob/trunk/scripts/setup_tomtom_requirements.sh).
This lines-up the way circleci builds tomtom with what is done for Linux or Windows for example (where a bunch of apt-get install commands are run inside the environment preparation step)

gefin:
Now the zip contain a good navit.xml for tomtom.
Also the layout files like navit_layout_car.xml are in the zip.

The problem now is that navit_layout_car.xml are empty and navit.xml dont include it.

lains:
I give it a try...
XLST format has always been a challenge for me, but I think this should work on both included and stand-alone layout definitions.

I also updated the part of the script that copies over of XML config giles, it is now more specific which will hopefully avoid copying over non-related xml files.

@gefin, does this work better?
I guess you have a tomtom target to test this on. I have an interest of knowing if the split of layouts into stand alone XML ([that I developped one year ago](https://github.com/navit-gps/navit/commit/e70a28963ff9754d3257361fad93aebf59edee08)) works on that platform. At least, it seems I broke it until now... sorry!

gefin:
@lains Thank you. 
The Layouts work now well on the real device.
Some gui internal icons are a little to big, but i think i am able to change this.

Need some tests.

gefin:
Today real drive worked good.
I changed some lines in navit.xml because i fond no setting in the XSLT.
So the GUI internal fit better to screen:
from
<gui type="internal" enabled="yes" >
to
<gui type="internal" enabled="yes" font_size="250" icon_l="48" >

Also added 2 OSD zoom buttons:
 <osd enabled="yes"  type="button" w="60" h="60" x="-60" y="50" command="zoom_in()" src="gui_zoom_in.png"/>                                    
 <osd enabled="yes"  type="button" w="60" h="60" x="-60" y="100" command="zoom_out()" src="gui_zoom_out.png"/> 

is this possible in the XSLT?

mvglasow:
@gefin re zoom buttons, have a look at #737, where we implemented a common OSD layout across all platforms. Tomtom was the only one left out back then because the build was broken and we had no way to verify things worked as intended. The (IMHO) perfect solution would be to apply the changes from #737 to Tomtom as well, with the necessary modifications—preferably by someone who knows their way around that particular build and can test on a real device :-)

mvglasow:
correction to the above: as I infer from the PR, we did implement this on Tomtom in the hope that it would work, as we couldn’t test it back then…

gefin:
I can test on TT 730 (480x272).

gefin:
@mvglasow just checked against #737 and it look like osd_minimum is included in tomtom.xslt
After build navit.xml contain no osd lines .
The other Question is how to add icon and font size.

metalstrolch:
You can have a look at the "sailfish" xslt's. I think they basically do what you need. You can copy from them.

mvglasow:
If navit.xml contains no OSD lines, that looks like something has gone wrong. Do the other settings from tomtom.xslt take effect? If not, that would be a sign the whole XSLT is not getting applied. One way or the other, you might want to look out for warning/error messages related to XML processing.

gefin:
I dont see errors or warnings.
The lines for gps source, map storage,  graphic, geometry some other things are applied.

If i change something bad i see messages.
The sailfish configuration is to different for me to see where the problem is.

gefin:
My changes from tomtom.xslt dont work.
With " copy-of " i get osd buttons into the xml, but i get them not enabled. 

I need help with tomtom.xslt

jkoan:
My suggestion would be that we only need one configuration regarding the display size. This can be achieved by using relative sizes within osd for example. This will simplify things. Also I would like to provide another pull request wich also tryes to fix #823 but with a cpack approach. If this is okay for you. Then we can decide which solution is better.

gefin:
Yes it is ok for me. We have only two relevant resolutions. 480x272 and 320x240

jkoan:
Will merge this now, and correct everything else in another PR since there are more than only on thing to work on. THX for your work

PR #863, opened 2019-09-03
[WIP] Android Navit64

jandegr:
fixes https://github.com/navit-gps/navit/issues/828

acts upon https://github.com/navit-gps/navit/issues/832

PR #876, opened 2019-09-19
Fix:Android:64bit and cleanup

jandegr:
A cleaned version as a followup on
https://github.com/navit-gps/navit/pull/863

Many more fixes to do but I have to draw a line from time to time


jandegr:
ouch, will open a new one

PR #878, opened 2019-09-20
fix:graphics/qt5; Allow to draw transparent polygons

metalstrolch:
There is a "misconception" in qt5 graphics about how to draw transparent stuff causing transparent items on the map to not correctly work. This PR changes qt5 graphics to force-clear overlays on starting to draw them instead of force clearing the shape of a transparent item. While this imposes the limitation that the content of an overlay cannot be drawn in multiple runs, it allows to draw transparent stuff on the map.
Luckily I don't know of any overlay item yet that is not drawn in one run. So this seems to work OK.


metalstrolch:
Immediately merging this as only relevant for qt5 graphics. Blame me on new bugs imposed by it.

Btw. I strongly need to find out how Navit initially intended to clear the overlays, or if and when it expects the overlay to be clear. No doc about it, so still guessing. Assume the background_gc callback of the graphics has something to do with that.

Issue #852, opened 2019-08-29
Missing "polygon with holes" support for android


metalstrolch:
It is planned to include multipolygon support #824 into navit before 0.5.4 release, which is a good thing indeed.
However for full experience of the then enhanced maps, it would be good to add graphics support for drawing polygons with holes on Android, SDL and Windows CE targets as well.

I added Windows support (except CE) to #824 as it was no big issue to do even with only CI as build environment.

I added Windows CE support in #862

I added SDL support in #861

Despite willing to do so I cannot help much for android, as I don't have access to any Android device to test the result. But I am willing to help pointing out what needs to be done.

metalstrolch:
Added Windows support except CE to #824.

jkoan:
Is there a way to add wince as well? i can do the test on wince

metalstrolch:
Not on the fly and not for me.
Windows CE seems to have a really limited and crippled GDI which doesn't support DrawPath used by me for win32. I fear that WindowsCE will be one of the platforms that doesn't have a native graphics library capable of drawing polygons with holes. So here we may be forced to manually triangulate.

But as not being a Windows dev, and the documentation about WindowsCE (C language specially) is really limited I may be wrong, but it seems I'm done here. I don't have any Device to test either, and the development cycles with only CI as build environment is rather slow.

Hopefully someone with Windows CE knowledge may help out here. Btw. I wonder that this old system is still in productive use.

metalstrolch:
Add SDL with #861
Same attempt suitable for Windows CE.

metalstrolch:
Add Windows CE with #862

pgrandin:
Really nice work @metalstrolch 

We have an experimental openSTF farm available. This could allow you to test android related features. Would you be interested?

metalstrolch:
Thanks for the offer @pgrandin,
but I think this may not make much sense. I'd need to refresh my knowledge on Java, learn about Android graphics API,  and learn some JNI. Additionally I'd need to learn how to build and deploy to Android as I virtually never even used any Android device. And It would require me to set up a development environment as only CI is too slow a development cycle. It worked for Windows because it's just "C" and I were lucky i got it right with few iterations.

Last but not least it would require me to learn coding habits for Android, as #832 convinces me is something I better not make my fingers dirty. I don't like to be blamed for code I write for alien architectures barely knowing the system.

[...]

PR #879, opened 2019-09-20
enhancement:layout_car:use transparency to draw some map features

metalstrolch:
With the introduction of the multipolygon code we now have all the nice landuses in the map. However, in OSM there are some map features that are not mapped as multipolygon but just "above" the other landuses. Like poly_zoo, poly_theme_park or poly_airfield.
Current configuration causes them to be hidden by the landuses, as these are drawn after the above mentioned. OSM's Mapnik style solves the problem by drawing those feature transparent above the landuses.
This pull request does so for some map features as well causing the map to look way better.

If transparent drawing is not supported by the platform, this causes the mentioned polys to be drawn above the landuses, showing them, eventually hiding some landuse details underneath. But this looks even better than before too.

NOTE: Transparent drawing is known to work on: Qt5 (sailfish) and gtk.
NOTE: Transparent drawing is known NOT to work on SDL
Dont't know for all the others. Sombody might want to check Android? Remember #852 on comparing screenshots though.

### DIFF
--- a/navit/navit_layout_car_shipped.xml
+++ b/navit/navit_layout_car_shipped.xml
@@ -239,8 +239,8 @@
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_farm" order="0-">
-			<polygon color="#c7f1a3"/>
-			<polyline color="#79c691"/>
+			<polygon color="#eef0d5"/>
+			<polyline color="#c7c9ae"/>
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_meadow" order="0-">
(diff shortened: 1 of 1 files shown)


metalstrolch:
See the difference on the small snippet near Ammersee, Munich:
zoo_solid
zoo_transparent
Note the airport on the upper left corner, and the "zoo" in the center. The zoo is drawn as a transparent outline only, the Airport is tinted by overlaying the shape transparent.

jkoan:
I did realignment and diffed the images. The Purple paths are where the PR makes changes:
diff

metalstrolch:
Wow thats cool. I was too lazy to do show it like that.
Most changes are due to giving meadows a lighter color. Nothing unexpected.

Android anyone?

Do we like that?

metalstrolch:
If no objections going to merge this at Fr, 27th of September.
I like it.

metalstrolch:
OK, merging this even if it's not friday yet

PR #881, opened 2019-09-22
Add xslt support

gefin:


gefin:
looks like 
scripts/setup_tomtom_requirements.sh
is not used by circleci.

gefin:
I close this request because it should be a addition to Pull #875.
To make things not so complicated i asked there for help.

PR #882, opened 2019-09-22
update:graphics:Avoid declaring variables if count < 2 in graphics_draw_polyline_as_polygon - minor

aerostitch:
When working on something else I realized that we do initialize some variables for nothing when count < 2 in that function.
So this is just to move the check of count before we start to allocate some memory.

### DIFF
--- a/navit/graphics.c
+++ b/navit/graphics.c
@@ -1911,12 +1911,13 @@ static void draw_init_ctx(struct draw_polyline_context *ctx, int maxpoints) {
 
 static void graphics_draw_polyline_as_polygon(struct graphics *gra, struct graphics_gc *gc,
         struct point *pnt, int count, int *width) {
+    if (count < 2)
+        return;
+
     int maxpoints=200;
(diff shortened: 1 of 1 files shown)


navit/graphics.c
```diff
@@ -1911,12 +1911,13 @@ static void draw_init_ctx(struct draw_polyline_context *ctx, int maxpoints) {
 
 static void graphics_draw_polyline_as_polygon(struct graphics *gra, struct graphics_gc *gc,
         struct point *pnt, int count, int *width) {
+    if (count < 2)
```
metalstrolch: I wouldn't do that change.
A: this is forbidden in pre C99 C.
B: Variables in question are initialized static and will be copied on stack as a block by compiler most probably.
So I'd propose either leave this as is, or else put whole function in "else" block instead. Then the variables are in another context. But changes are high that even then they are copied as block on stack.

aerostitch: Alright, closing then.

PR #884, opened 2019-09-23
fix/enhancement:graphics/layout:get default icon size from layout + draw tunnels transparent + mark oneway streets

### LINKED ISSUE #819: New icons are shown very big
Some of the POI icons are shown really big in the map. They happen to be the ones that look the newest icons (flat style in blue color).

Opening the svg files with a text editor, I see that they have properties width and height = 200. The ones that are shown at a proper size are 22x22.

I think that they should use a size defined in the code, or in navit.xml or something like this, not the size defined in the .svg itself since this is too error prone.

However, just in the case that you want to fix the .svg files, this is the list of affected files (I hope it's complete, but there may be not)

```
grep -rn ' height=' *.svg | awk '{FS=":"; if($2<25) print}' | awk '{FS="\""; if ($2 == 200) print}'
bowling.svg:12:   height="200"
bridge.svg:12:   height="200"
car_wash.svg:12:   height="200"
communication.svg:12:   height="200"
concert.svg:12:   height="200"
crossing.svg:12:   height="200"
dam.svg:12:   height="200"
danger_area.svg:12:   height="200"
diving.svg:12:   height="200"
drinking_water.svg:12:   height="200"
emergency.svg:12:   height="200"
fish.svg:12:   height="200"
forbidden_area.svg:12:   height="200"
ghost_town.svg:12:   height="200"
government_building.svg:12:   height="200"
gui_heightprofile.svg:1:<svg width="200" height="200" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
hotspring.svg:12:   height="200"
```

navit-big-icons

metalstrolch:
This pull request adds the possibility to globally set a default for icon size of a layout.
You can now give "icon_h" and "icon_w" properties in "layout" tag. This causes navit to not use the real size of an icon but to scale it to have the requested size. Guessing prescaled icons (the name_w_h.png's of course works. Default size of 22x22px which is the default size hint on most of the svg's is used. This fixes #819.

This pull request adds the property "underground_alpha" to the "graphics" tag giving the alpha value to use as transparency if ways are flagged with AF_UNDERGROUND. This effectively renders tunnels in transparent.

This pull request adds a "oneway" tag to itemgras in layouts. Together with the enhancements of the "arrows" itemgra used for drawing the route graph one can print nice arrows on oneway roads.

### DIFF
--- a/navit/attr_def.h
+++ b/navit/attr_def.h
@@ -263,6 +263,7 @@ ATTR(waypoints_flag) /* toggle for "set as destination" to switch between start 
 ATTR(no_warning_if_map_file_missing)
 ATTR(duplicate)
 ATTR(has_menu_button)
+ATTR(oneway)
 ATTR2(0x0002ffff,type_int_end)
 ATTR2(0x00030000,type_string_begin)
 ATTR(type)
--- a/navit/attr_def.h
+++ b/navit/attr_def.h
@@ -202,6 +202,7 @@ ATTR(autozoom_max)
 ATTR(nav_status)
 ATTR(virtual_dpi)
 ATTR(real_dpi)
+ATTR(underground_alpha)
 ATTR2(0x00027500,type_rel_abs_begin)
 /* These attributes are int that can either hold relative or absolute values. See the
  * documentation of ATTR_REL_RELSHIFT for details.
--- a/navit/color.h
+++ b/navit/color.h
@@ -21,7 +21,7 @@
 #define NAVIT_COLOR_H
 
 struct color {
-	int r,g,b,a;
+    int r,g,b,a;
 };
 
 #define COLOR_BITDEPTH 16
@@ -34,6 +34,8 @@ struct color {
 #define COLOR_TRANSPARENT ((struct color) {COLOR_TRANSPARENT_})
 #define COLOR_FMT "0x%x,0x%x,0x%x,0x%x"
 #define COLOR_ARGS(c) (c).r,(c).g,(c).b,(c).a
+/*default alpha value to apply for all things flagged AF_UNDERGROUND*/
+#define UNDERGROUND_ALPHA_ 0x3333
 
 #define COLOR_IS_SAME(c1,c2) ((c1).r==(c2).r && (c1).g==(c2).g && (c1).b==(c2).b && (c1).a==(c2).a)
 #define COLOR_IS_WHITE(c) COLOR_IS_SAME(c, COLOR_WHITE)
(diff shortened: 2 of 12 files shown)


metalstrolch:
While the icon size is too stupid too screenshot, the tunnel transparency is quite nice. See the following screenshots (done with gtk_drawing_area) as example:
tunnel_solid
tunnel_alpha
Note the shadows at the road joints in the transparent tunnel show how Navit is actually drawing internally and cannot easy be removed. But that's better than nothing.
 The transparency can be controlled by "underground_alpha" value in layout xml.

mvglasow:
Great work! However, I would go for slightly less transparency for the underground ways to make them more visible. I would suggest something close to the OSD elements, which IIRC are around 70% (0% being fully transparent, 100% being completely opaque).

metalstrolch:
OK. Enabled arrows on oneway streets for higher zoom levels on car layout.
arrows_2d
arrows_3d
It's working on 2d and 3d view. And I increased the tunnel density.

lains:
This is a great feature that I have been missing a few times while driving (both tunnels and one way roads). Thanks for making it alive!

metalstrolch:
WIP again, as there still is a BUG in calculation of the arrow placement for one way. Seems I've missed some corner case. Try to fix this this evening...

metalstrolch:
Should be fixed now. Keep WIP until I've gotten more time to test.

metalstrolch:
Yeslooks good. Ready for review and merge.

lains:
Great! Great!

Any chance we can reduce the number of arrows displayed (for example first and last ones on a road where there is no other way in/out). This is probably tricky to implement though...

Also, if I understood the screenshots correctly, only tips of arrows are currently drawn, maybe a body (even short) would be good to disambiguate the direction (especially in 3D) when arrows end up being drawn almost as a triangle.

metalstrolch:
Actually, the fix for drawing reduces the number of arrows somehow. But we cannot just draw the first and last before a junction, because on drawing we don't know if there are junctions any more. Drawing is done item by item. Looking at the transparent parts of the screenshots you can see the granularity by spotting the transparent circles visible. Improving this would require a complete rework how Navit draws. This isn't worth it, I think. Specially given that Navit couldn't mark one way streets ever since.

metalstrolch:
And off we go ...

metalstrolch:
Does somebody know if/where we document attributes for xml?

mvglasow:
We have something on the wiki, I believe the entry point is here: https://wiki.navit-project.org/index.php/Configuration —updates welcome!

PR #886, opened 2019-09-26
cleanup:search:Minor - Use fallthrough like the rest of the cases

aerostitch:
This is really a pedantic commit but as I was reading that part of the code I wasn't sure of why we didn't use the fallthrought like the rest of the cases here, so just a tiny cleanup.
Let me know if there's a reason I missed.

Thanks
Joseph


jkoan:
In my oppinion we should do exectly the opposite and get rid of all fallthrought code over time. As in my oppinion fallthrought code does not increases the readability

aerostitch:
I do agree that fallthrought decreases readability and I just wanted consistency in there.
So instead I moved to no fallthrought. Are you still ok with that @metalstrolch ?

pgrandin:
I agree with @metalstrolch on this one. The latest commit in this PR decreases readability IMO

aerostitch:
Lol! Works for me!
I reverted the last change so that it's only the 1 line removal.
I'll merge as soon as circleci is done.
Thanks for the review guys! :)

mvglasow:
IIRC fallthrough is needed to silence compiler warnings about case statements without `break`. I have added some of those in the past to get rid of these excessive warnings. However, I am not sure if case statements with no body at all would trigger that warning as well.

I don’t see readability as so much of an issue, either with or without. Where the lack of either fallthrough or `break` does not generate compiler warnings, I’m OK with removing the fallthrough. Elsewhere we’ll need to see if we can disable compiler warnings about case statements without `break` altogether—at the cost of no longer getting informed about truly forgotten `break` statements.

metalstrolch:
As far as I noticed, at least gcc does not warn if cases fall through without any statements between the cases. This is the most common and widely accepted use of falling through case usage. Otherwise gcc requires a comment statement like /*falls through*/ to not generate a warning on falling through case.

PR #877, opened 2019-09-19
Fix:android:64 bit and cleanup2

jandegr:
A cleaned version as a followup on
#863

Many more fixes to do but I have to draw a line from time to time

after this you may have to set autobrief on doxygen if not already so
`If the JAVADOC_AUTOBRIEF is set to YES then doxygen will interpret the first line (until the first dot) of a JavaDoc-style comment as the brief description.`

fixes https://github.com/navit-gps/navit/issues/828
fixes https://github.com/navit-gps/navit/issues/832
fixes https://github.com/navit-gps/navit/issues/550

PR #890, opened 2019-09-29
Fix:Android:system bars 

jandegr:
Fixes system bars losing transparency while dragging the map, a todo leftover from https://github.com/navit-gps/navit/pull/877, and a bit of cleanup for the pedestrian plugin.

### DIFF
--- a/navit/android/build.gradle
+++ b/navit/android/build.gradle
@@ -7,7 +7,7 @@ android {
     buildToolsVersion "29.0.2"
     defaultConfig {
         applicationId "org.navitproject.navit"
-        minSdkVersion 9
+        minSdkVersion 10
         targetSdkVersion 28
         versionCode gitVersionCode
         versionName gitVersionName
--- a/navit/android/res/values-v19/styles.xml
+++ b/navit/android/res/values-v19/styles.xml
@@ -1,8 +1,5 @@
 <resources>
     <style name="NavitTheme" parent="android:Theme.DeviceDefault.NoActionBar">
-
         <!-- API 19 theme customizations can go here. -->
-        <item name="android:windowTranslucentNavigation">true</item>
-        <item name="android:windowTranslucentStatus">true</item>
     </style>
 </resources>
--- a/navit/android/res/values-v21/styles.xml
+++ b/navit/android/res/values-v21/styles.xml
@@ -1,21 +1,5 @@
 <resources>
     <style name="NavitTheme" parent="android:Theme.DeviceDefault.NoActionBar">
-
         <!-- API 21 theme customizations can go here. -->
-        <!--
-            Don't use translucent system bars on API 21 as they are drawn with a semitransparent
-            black background which can't be changed.
-        -->
-        <item name="android:windowTranslucentNavigation">false</item>
-        <item name="android:windowTranslucentStatus">false</item>
-
-        <!--
-            We could set any semi-transparent color here (or change it in code), but this would not
-            be available on API 19/20. Simply specifying full transparency allows us to implement a
-            separate mechanism that will work on all versions.
-        -->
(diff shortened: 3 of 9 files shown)


jandegr:
bumps minsdk from 9 to 10, as discussed on IRC and is open for discussion here as well.

motivation :
- don't promise it will work on devices it is not tesed on
- the oldest device known to me amongst the contributers is an sdk 10 tablet
- Google dropped support for armeabi a few ndk's ago and to my best knowledge, the oldest armv7 devices have sdk 10.
- STF works from sdk 10 upwards

jandegr:
tested on

android 4.2.2 (sdk 17) Asus
android 4.4 (sdk 19) AVD
android 8.0 (sdk 26) SamSung STF @pgrandin 
android 9.0 (sdk 28) AVD

gefin:
tested with android 7.1.2 LG

jandegr:
On the MI A2 STF @pgrandin , just because I was asked to post STF screenshots :)



MIA2_1

jandegr:
Changed the code for fullscreen mode to the few lines we had for this years ago for devices below kitkat. Fullscreenmode does nothing from kitkat upwards since those have the transparent bars.

mvglasow:
Tested on a Nexus S, Android 4.4.4 (Cyanogenmod) and a OnePlus One, Android 8.1 (LineageOS 15.1); both work.

mvglasow:
Is there a need to remove fullscreen for Lollipop and up? If it still works, I’d suggest leaving it in. It still offers slightly more screen real estate than the transparent system bars do and some people might prefer it that way.

Otherwise, is there any obstacle to merging this?

jandegr:
Hi

>  is there any obstacle to merging this?

I am not into merging my own things. I only post solutions with some kind of a test report.
If navit-gps wants it, anybody can (squash)merge it.

> Is there a need to remove fullscreen for Lollipop and up?

I tried to make it clear that I only back compliant, properly tested and working solutions. During testing it became clear there was trouble with the fullscreenmode. Reverting it to the few lines we had long ago and have been used for many years on pre kitkat devices seemed a safe option until an extensively tested, working for everybody solution becomes available.

jandegr:
Testing on a nokia android 9 (sdk 28)
It is one of those where the camera is in a notch in the screenand the screen has rounded corners but the screenshots show no sign of it.

Fresh install and download europe on a 64GB SD card

device-2019-10-14-180142


Using the all-of-europe map
device-2019-10-14-192149

jandegr:
These Android-reworked builds are already prepared for the upcoming restrictions on access to storage since they use `getExternalFilesDir()` and there is no need to use the provided temporary workarounds.

About their warning `next year.... independent of target SDK level` I just wonder how they will enforce that.

For fresh installs there will be no problem, but these builds preserve non-compliant paths in an upgrade situation as a convenience. Their announcement is a little bit ambigue to me, but if I interpret it in a strict sense, then the convenience of picking-up existing non-compliant paths may have to be removed some time next year at the latest. I will try to test that as soon as possible using the `opt-in` option.

jandegr:
### final statement

Donwload of large binfiles (over 3,8GB) has been tested on the nokia onto the SD card and on an AVD sdk28 tot the onboard memory (after increasing the memory size in its settings). While both tests revealed no problems, it is not enough to be conclusive, so the option is not included in this PR.

The probable issue with storage may need to be reviewed as descirbed above in the coming months.

For the rest, I tested just about everything I could on any device within reach + AVD's + devices on remote locations + a few occasional test reports from contributers.

There are a few scenario's known to me that lead to a crash, but with reasons deeper down in the C code shared with other platforms, if they were in the android-only code I would have fixed them.

good luck with it folks.

Issue #888, opened 2019-09-27
circleci: sanity_check downloads hundrets of megabytes environment.


metalstrolch:
Currently the (annoying) sanity_check step of circle_ci seems to apt-get a full blown development environment inside it's environment, albeit it should not require that for astyle checking.

This makes sanity_check take way longer than required.

metalstrolch:
fixed by #891

PR #891, opened 2019-09-30
fix:circleci: do not download build environment for sanity check

metalstrolch:
and reenable xml lint for all configurations shipped. To do this the navit.dtd model is updated by the new properties added.


metalstrolch:
fixes #888

PR #897, opened 2019-10-03
fix:core:prevent stack overflow on drawing huge polygons

metalstrolch:
This commit prevents c_alloca from flooding the stack if we have to deal
with really huge polygons. Use slower g_malloc instead.


metalstrolch:
No objections. Off we go

PR #901, opened 2019-10-07
enhancement:maptool:improve processing speed and fix some memory holes

metalstrolch:
This pull request increases processing speed of maptool on large areas (world) by partly parallelizing the turn restriction processing as well as by introducing buffered read of nodes on relation processing.
Time saved on my box: -1day / ~20%

### DIFF
--- a/navit/maptool/itembin_buffer.c
+++ b/navit/maptool/itembin_buffer.c
@@ -21,19 +21,50 @@
 #include "maptool.h"
 #include "debug.h"
 
-
 /** Buffer for temporarily storing an item. */
 static char misc_item_buffer[20000000];
 /** An item_bin for temporary use. */
 struct item_bin *tmp_item_bin=(struct item_bin *)(void *)misc_item_buffer;
-/** A node_item for temporary use. */
-static struct node_item *tmp_node_item=(struct node_item *)(void *)misc_item_buffer;
 
 struct node_item *
 read_node_item(FILE *in) {
-    if (fread(tmp_node_item, sizeof(struct node_item), 1, in) != 1)
-        return NULL;
-    return tmp_node_item;
+#define ITEM_COUNT 1*1024*1024
+    /* we read in bigger chunks as file IO in small chunks is slow as hell */
+    static FILE * last_in = NULL;
+    static long last_pos = 0;
+    static int in_count;
+    static int out_count;
+    static struct node_item item_buffer[sizeof(struct node_item) * ITEM_COUNT];
+
+    struct node_item * retval = NULL;
+
+    if((last_in != in) || (last_pos != ftell(in))) {
+        fprintf(stderr, "change file. Still %d items\n", out_count - in_count);
+        /* got new file. flush buffer. */
+        in_count=0;
+        out_count=0;
+        last_in=in;
+    }
+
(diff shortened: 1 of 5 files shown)


jkoan:
How many threads did you assigned on your tests and what would you suggest as a optimum if running with ryzen 2600 and NVMe ssd?

metalstrolch:
I used the now default of 8 threads. I wouldn't use many more because, with every thread, you split the resulting hash. I used 7GiB of slice size on a 16GiB RAM machine. Remember to do `ulimit -s unlimited`.

My machine is a core I7 NUC with a spinning USB attached disc. So I expect yours to be way faster. Unfortunately currently the disk speed doesn't help that much.

Some explanation: 

On creating the hashes, you basically read all items in a common queue, and every thread hashes the next item it gets into its own private hash. As hashing scales bad with number of elements, this helps a lot because a: more processing power, and b: smaller hashes per cpu.

BUT:
On processing the hash, you have to try all items against all hashes! Sot the more threads you used before the slower that gets. Luckily using the hash scales way better. This can only be done by one thread currently, as most parts of maptool are not thread safe!

Additionally using your SSD, the smaller fix contained will help a lot, I think. instead of reading the coordinates (several billions in world) one by one from disk I added a buffer reading them several MiB at once and then serving them from the buffer. This helps most on turn restriction processing.

metalstrolch:
Somebody planning to test this or is already testing this? Works for me.

If no objections: Planning to merge on Friday 18th October 2019.

gefin:
Sorry i cant test (map size>>device size) 
I can test the planetextractor result after it is active...

jkoan:
> Somebody planning to test this or is already testing this? Works for me.
> 
> If no objections: Planning to merge on Friday 18th October 2019.

Maptool running against pbf of Germany ;) Will show a comparison when finished

metalstrolch:
Does anybody have a ready-made tool to extract regions out of a world binfile? While the algorithm is basically described in https://navit.readthedocs.io/en/latest/maps.html#binfile , it would stupid to reimplement something that seems to exist at least for planet extractor. Would help a lot to check #906

jkoan:
| version | run num | real | user | sys | 
| --- | --- | --- | --- | --- |
| trunk  | 1 | 41m15,070s | 40m22,668s | 1m55,011s | 
| trunk  | 3 | 44m19,450s | 42m38,990s | 2m26,089s | 
| maptool_turn_restriction_threading | 2 | 39m8,376s | 38m22,635s | 1m59,468s | 
| maptool_turn_restriction_threading | 4 | 40m15,357s | 38m50,225s | 2m18,419s | 

Runs number 1&2 did run after each other. Run number 3 and 4 run at the same time (different dirs). 
Command: `maptool --protobuf -T $(nproc) -i germany-latest.osm.pbf germany-trunk2.bin`
pbf-File: `https://download.geofabrik.de/europe/germany-191015.osm.pbf`

metalstrolch:
Merging this albeit it seems to be system dependent or dependent on the size of map file how much speed is gained. At least on my setup this saves around 20% processing time. At least some memory holes get fixed.

metalstrolch:
Will keep the branch for a while for reference.

Issue #548, opened 2018-04-26
Android: Add support for Document Provider


jkoan:
From a Facebook report i got a issue that Navit is not able to write to the SD-Card because of the new Document Provider which must be used for a

https://developer.android.com/guide/topics/providers/document-provider
Also a good comment on this toppic: https://stackoverflow.com/a/26765884

---- The Log
04-26 18:50:59.579  5051  5130 E NavitMapDownloader: Could not write info file for map download. Resuming will not be possible. (/storage/0000-0000/navit/Niedersachsen.tmp.info (Permission denied))
04-26 18:50:59.579  5051  5130 D NavitMapDownloader: size: 861307928, read: 0, timestamp: 0, Connection ref: http://maps9.navit-project.org/api/map/?bbox=6.40,51.24,11.69,54.22&timestamp=180424
04-26 18:50:59.580  5051  5130 E NavitMapDownloader: Could not open output file for writing: java.io.FileNotFoundException: /storage/0000-0000/navit/Niedersachsen.tmp (Permission denied)

jandegr:
Do you know smth more of the used devive, api and such.
Just did a test on android 6.0 downloading to SD card after giving both the permissions and changing the location to SD, all went fine

jkoan:
https://www.facebook.com/navitproject/posts/1812419968781103 look at the comments. The log is from the same person, he''s using a samsung smartphone (Galaxy J3)

jandegr:
I did not find the relevant entry on Facebook, but don't have a FB account either.
Anyway, those 0000-0000 are weird, when I tested it was smth like /storage/C348-1DE1/navit....

killerhippy:
those 0000-0000 come from an original Samsung SD Card, before I had another which announced itself like 348-1DE1 - strange, but works identically. ES Dateimanager behaves on both SD cards the same like navit does.

killerhippy:
![screenshot_20180424-184506](https://user-images.githubusercontent.com/15832555/39326762-507168e8-4996-11e8-87ca-8df0ef690ace.png)
![screenshot_20180424-184517](https://user-images.githubusercontent.com/15832555/39326763-5093c2e4-4996-11e8-856c-d965828675db.png)
![screenshot_20180424-184525](https://user-images.githubusercontent.com/15832555/39326765-50b6e17a-4996-11e8-8fc7-b357147da8fd.png)
How ES Dateimanager reakts on my try to create a new directory at the root dir of the external SD-Card.

[...]

PR #908, opened 2019-10-12
Android:feature settings

jandegr:
Serves as an illustration for https://github.com/navit-gps/navit/issues/548 and since this is for testing only this PR will never leave Draft status.

Adds an `experimental settings` entry in the Android menu.
The only implemented one is a test for devices troubled by https://github.com/navit-gps/navit/issues/548

This is a very minimalistic implementation with the following properties
- in an updrage install where a custom location was used, this location will be pciked up
and further used, but once you change to one of the suggestions in the settings menu, there is no way to go back to the original custom location (yet).
- unfortunatly it still requires a restart after changing the map location
- the suggested location for SD is supposed to fix it for all devices in  https://github.com/navit-gps/navit/issues/548
- for Marshmallow and up only
- no need for permission storage to use the suggested locations
- you can have maps in both the suggested locations and switch between them (with the restart inbetween ofcourse)

Screenshot_1570896439

### DIFF
--- a/navit/android/AndroidManifest.xml
+++ b/navit/android/AndroidManifest.xml
@@ -10,7 +10,9 @@
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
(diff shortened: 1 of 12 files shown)


jandegr:
Since this pr just serves  as an illustration for #548 there is no need for this to be an open PR, even if it has a draft label, so closing it.

Issue #910, opened 2019-10-13
Too much time spend in CI


jkoan:
When looking into [this](https://circleci.com/workflow-run/68e792ae-fb84-47ed-9206-d42eafaecd02) i see that 3 builds take over 10mins. This is a problem because we only have limited time to spend on the ci. Which we BTW already exceed a lot (strange that we can still build)

For run_doxygen my suggestion would be to store no artifacts since they are already stored within github or zip and upload them. The problem there is only about the small file sizes.

For build_linux again the artifact upload is the problem, we dont need to upload the unpacked files, *STGZ,TGZ,TZ should be enough.

For build_sailfish i have no clue, this takes long, but i dont have analyzed it right now.

We should also look into the other builds but they are not so high in build times. 

https://circleci.com/workflow-run/68e792ae-fb84-47ed-9206-d42eafaecd02

lains:
And it seems some builds are on hold (at least, build results for https://github.com/navit-gps/navit/pull/812 have been pending for one hour now, it seems that not having these steps completed prevents merging)

jkoan:
Thats not true, but the results are now shown on the github checks tab

jandegr:
Hi,
How does circleCI measure time used ?

From the start of the workflow to the end of it or as the summ of time consumed by each induvidual container ?

Either way, I don't see the point in making all the other steps wait untill astyle finishes, maybe it makes sense to let 3 others run aside of it since a failure of astyle will block a PR anyway.
Or would this increase time used on ci because of the 3 that were allowed to run next to astyle ?

 Years ago when I ported the tomtom build over from the autotools version on the old buildserver to a cmake version on CI, I got a remark for building both the plugin version and the full version, so I disabled one of them (don't remember which one) but I did not delete that buildscript. However since quite a while I see both get built again. I am not involved in the tomtom builds andymore but I suppose you could poll those that still are into tomtom which version they prefer, they can still build the other one on a need to basis since the script to do so would not be deleted.

The tomtom remark falls under 

> We should also look into the other builds but they are not so high in build times.

jkoan:
> How does circleCI measure time used ?

It's the sum of all invidual containers including setup and artifact upload. So when we already know that the PR will fail due to astyle we can safe time here

jandegr:
Ok, thx
So this leaves the remark about tomtom and android translations I made higher 
+ new remark : why build everything all over that has already been built when merging into master ?
[...]

PR #911, opened 2019-10-13
CI:cleanup

jandegr:
Storing translations was added to the android build on CI a long time ago to debug a specific issue.
It takes up to 30 seconds but is not needed anymore.
relates to https://github.com/navit-gps/navit/issues/910


jkoan:
Merged this PR with admin rights since the pull requests from @jandegr seem to run on his account instead of the Navit-gps account which leeds to different statuses which are reported back to github.

PR #915, opened 2019-10-17
fix:maptool:create valid zip64 central directory

metalstrolch:
This fixes the zip64 central directory created by the maptool.
While the broken one works on navit, it failed to extract with usual zip
tools due to parts of the zip64 end of directory part not being filled
correctly with values.
This happens on all 64 bit zip files generated by maptool including the sample map since the multipolygon patches.

Files downloaded from planet extractor are and always have been corrupt as well. Without the source of the planet extractor, I don't know if this actually fixes the files shared by planet extractor as well.


jkoan:
As far as i rememper from reading the zip spec some time ago this is correct :D

metalstrolch:
Merging this. Hope it doesn't have any negative effect on the planet extractor.
Done this to ease up working on a "planet extractor" tool myself.

Issue #898, opened 2019-10-04
Cleanup:Duplicate function search_fix_spaces() in android.c and search.c


lains:
Code is almost identical:

https://github.com/navit-gps/navit/blob/56f6be45d5ce5f961c1826ef2f06d6678f770eb0/navit/search.c#L138-L164

and 

https://github.com/navit-gps/navit/blob/56f6be45d5ce5f961c1826ef2f06d6678f770eb0/navit/android.c#L683-L710

aerostitch:
Let me know if #917 is the fix you had in mind. I'm open to suggestions about better ways to do that as I'm a novice in C programming.

lains:
Thanks for your work.

PR #917, opened 2019-10-20
cleanup:search:Remove duplicate code for search_fix_spaces

aerostitch:
Closes: #898 

The only differences I can see between the function in search.c and the one in android.c are:
* the signature takes a `const char *` in android.c instead of a `char *` which shouldn't be a problem
* the 1st if of the for loop has a duplicate `ret[i] == ',' ||` condition which is cleaned up with the PR

Note that this would add a compilation warning saying:
```
warning: passing argument 1 of ‘search_fix_spaces’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
```
But as we need the const context for the android part to call `ReleaseStringUTFChars` I'm not sure if there's a better way to do that.

### DIFF
--- a/navit/android.c
+++ b/navit/android.c
@@ -682,7 +682,8 @@ static void android_search_idle(struct android_search_priv *search_priv) {
 
 static void start_search(struct android_search_priv *search_priv, const char *search_string) {
     dbg(lvl_debug,"enter %s", search_string);
-    char *str=search_fix_spaces(search_string);
+    char tmp_search_string = *search_string; // This is allows us to avoid warnings about discarding const qualifier
+    char *str=search_fix_spaces(&tmp_search_string);
     search_priv->phrases = g_strsplit(str, " ", 0);
     //ret=search_address_town(ret, sl, phrases, NULL, partial, jni);
 
--- a/navit/android.c
+++ b/navit/android.c
@@ -680,35 +680,6 @@ static void android_search_idle(struct android_search_priv *search_priv) {
     dbg(lvl_info, "leave");
 }
 
-static char *search_fix_spaces(const char *str) {
-    int i;
-    int len=strlen(str);
-    char c,*s,*d,*ret=g_strdup(str);
-
-    for (i = 0 ; i < len ; i++) {
-        if (ret[i] == ',' || ret[i] == '/')
-            ret[i]=' ';
-    }
-    s=ret;
-    d=ret;
-    len=0;
-    do {
-        c=*s++;
-        if (c != ' ' || len != 0) {
-            *d++=c;
-            len++;
-        }
-        while (c == ' ' && *s == ' ')
-            s++;
-        if (c == ' ' && *s == '\0') {
-            d--;
-            len--;
-        }
-    } while (c);
-
-    return ret;
-}
-
 static void start_search(struct android_search_priv *search_priv, const char *search_string) {
     dbg(lvl_debug,"enter %s", search_string);
     char *str=search_fix_spaces(search_string);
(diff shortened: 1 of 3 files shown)


aerostitch:
I added https://github.com/navit-gps/navit/pull/917/commits/1b5044786e63188a57aad83ed5afe2720168cf9e to avoid that discarded qualifier warning. Let me know if you want to keep it or not. Either way is fine with me.

navit/android.c
```diff
@@ -680,38 +680,10 @@ static void android_search_idle(struct android_search_priv *search_priv) {
...
-        }
-    } while (c);
-
-    return ret;
-}
-
 static void start_search(struct android_search_priv *search_priv, const char *search_string) {
     dbg(lvl_debug,"enter %s", search_string);
-    char *str=search_fix_spaces(search_string);
+    char tmp_search_string = *search_string; // This is allows us to avoid warnings about discarding const qualifier
```
metalstrolch: Sorry guys, but this won't work.

The statement 
`char tmp_search_string = *search_string`
will copy the first char (8 bits) out of the char array search_string into tmp_search_string variable.
Later on
`search_fix_spaces(&tmp_search_string);`
passes the reference to this one byte variable as a parameter. This is neither the complete string nor is it zero terminated.


As the input in both versions is g_strdup'ed enyways, it should be safe to treat the input string always as const avoiding the need to do any bulls*** to remove a const.

But are you sure the both versions really behave the same? For me it doesn't really look so on the first glance. The second "do" loop seems to shift the content of "ret" to remove leading and trailing spaces?

I'd start with correctly commenting the header of both versions and use this work to analyse what both versions do.
Then I'd reconsider if and how to unify them.

Bonus: Who frees the duplicate returned as a result in both versions?

jandegr:
To test I started with testing the old code 

2 sample tests (the `-` is printed to show the end and is not part of the string)

searchstring = de, /t          -
resultstring = de t-

searchstring = ,,   ,///,f th,u/h -
resultstring = f th u h-


based on the output of some tests the behaviour can be described as

replaces `,` and `/` by a `space`
then
removes all trailing spaces and limits consequtive spaces 'within' the string to one single

metalstrolch:
@jandegr  Yeah, but the code in search.c for example doesn't remove multiple spaces from the result, while the android.c version does. So if removing the android version, the android code then gets feed with search strings containing multiple space sequences. Which may or may not be a problem. Same the other way round.

jandegr:
I was writing the comment while tests proceeded so I do not have to wast a piece of paper to take notes :)
so step 1 was test (and describe) the existing code from within Android, step 2 and follow in a while,

aerostitch:
Hi @metalstrolch ,

Good catch for the redirect, thanks!
I changed the function signature to add the const (I didn't actually realize that it would work providing a char to a function having a const char in the signature).

I compared both functions using vimdiff and I don't see the difference outside of the extra empty line and signature. Based on my tests it looks to me that they both handle multiple spaces.

I've added also the comments for the function.

Let me know if you still see any issue with the new version.

Thanks a LOT for both of your time guys,
Joseph

aerostitch:
I've made some unit tests separately in https://gist.github.com/aerostitch/faa106d7534416c4dd9dc7e57dbfd209 and discovered a corner case that was not handled properly which is that the pointer was not incremented when the string was only containing chars that we wanted to discard. I addressed that issue in https://github.com/navit-gps/navit/pull/917/commits/1cecebf85ca4ae8ca152a3e8169a7ebc5ff8a368

But then I started thinking that maybe we'd like to discard not only spaces but all white-space characters so I made a slightly different implementation of the function: https://gist.github.com/aerostitch/faa106d7534416c4dd9dc7e57dbfd209#file-search-c

Let me know which one you like best.

Thanks!
Joseph

jandegr:
I am ok with any version, if somebody gives ok for those C related issue.

aerostitch:
Thx! :)

PR #921, opened 2019-10-25
fix:doc:First set of style-fixes for reStructuredText/Read the Docs

jkoan:
These are a lot of fixes to fix to fiy styling of reStructuredText files, because they are invalid, now sphinx-build . build/html -E -v reports no Warnings any more.
More fixes will follow in other PRs to get updates to Read the Docs as fast as possible.


jkoan:
I really have no clue why github checks dosnt show up... but all oldstyle checks passed (its only a docs change anyway)

aerostitch:
Based on https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html you don't **have** to overline the 1st level titles. Maybe we should keep the titles that are not document title (the one on the top of the page) overlined + underlined and the rest just underline. What do you think?
I'm just asking because it seems that overlining is a bit too much here.

Also is the following convention what you're going for? (Just to make sure we're on the same page)
```
# with overline, for parts
* with overline, for chapters
=, for sections
-, for subsections
^, for subsubsections
“, for paragraphs
```

Issue #924, opened 2019-10-28
Navit crashes if displaying big map items


metalstrolch:
If displaying big map items (as today the Lake Nasser, Egypt (https://www.openstreetmap.org/relation/280282) navit crashes because it floods the stack.

On many parts of graphics.c, g_alloca is used for allocating memory on stack for converting nodes. Unfortunately this fails flooding the stack on very big map items as the mentioned lake Nasser.

Proposed fixes: 
1. Use slower g_malloc/g_free if the node count gets too big. There are already some examples of this in code. Navit shouldn't crash as long it doesn't reach out of memory condition. Never.

2. Slice such map elements into smaller parts using the item cutter introduced to maptool with #906. Since for example lake Nasser doesn't trigger slicing yet because it is not that big geographical wise causing it to be shifted too much up in tile stack. This will limit peak memory usage.

metalstrolch:
So #928 fixes the crash. Done 1.

But lake Nasser is about 5 Mib compressed still let's think about 2. next.

metalstrolch:
While Navit doesn't crash anymore as long as there's enough ram, I'll close this and continue work on big map items in another issue.

PR #928, opened 2019-10-29
fix:core:Don't flood stack on big map elements

metalstrolch:
Use g_malloc instead of g_alloca on huge map elements.

Correctly calculate element intersection with map selection
even if buffer is too small.

First part of #924


metalstrolch:
Albeit doing what it should, I think I'll rewrite item_coord_get_within_selection function. Not because it's cyclomatic complexity, but because it got too ugly. Stay tuned.

metalstrolch:
Hmm, seems CodeFactor checked the previous revision? Can somebody confirm this?

metalstrolch:
To be honest, I don't really care for cyclomatic complexity count of CodeFactor. Seriously. I ranted enough about that. What puzzled me this time is that CodeFactor happily refers to outdated version https://github.com/navit-gps/navit/pull/928/commits/8a9c2cb0bb89b030e4d63119e604697d7246f242 in it's problem report. But hey, CodeFactor does lot's of funny things. Merging anyway, as this bug is quite nasty. Let's see what top do with Lake Nasser in map. Lake Nasser is by factors the biggest single OSM polygon.

PR #929, opened 2019-10-29
cleanup:ci:Simplify the listing of files for sanity check

aerostitch:
The current way of doing with interval sometimes leads to the last commit from trunk being included in the diff, leading to false positives.
This change simplifies the way we pull the files, and should make it more reliable.


pgrandin:
I like it!

PR #935, opened 2019-10-30
fix:all:fix compiler warnings

metalstrolch:
This pull request tries to solve most compiler warnings. Specially the nasty ones about wrong function pointer casts.


pgrandin:
That's pure grunt work. Thank you for taking care of it!

Issue #561, opened 2018-04-30
Unit testing in Navit?


aerostitch:
Hi guys,

I wanted to add some unit testing in Navit, that would allow people like me to better test their changes.
Would you be ok with that?
If so what unit test framework would you be comfortable to go with?

Thanks for your help,
Joseph

jandegr:
Hi,

Anything related to CI and Android should also work on Debian, that's what their dedicated android containers are built on.

thx

metalstrolch:
Personally I don't care about which unit test framework to use, as long as it adds no exotic dependency on non unit test build. Navit is quite often cross compiled, and there such dependencies are hard to fullfill.

Additionally, one should think about suitable granularity to test. What are our "units". Can be hard on legacy code already existing.

I think a first step could be better documenting the plugin interfaces (like "graphics" or "vehicle") and try to craft proper test functions for this. Would at least ease up creating proper plugins. And I would be really curious if all plugin interfaces are still known. I myself don't for example know all callback function's properties for the graphics interface. Even after I ported a graphics plugin.

metalstrolch:
Disagree in unit=function. One of the most common mistakes in doing unit tests. Will create LOTS of unit test code with not much benefit. Will even hinder improvements. as nobody likes writing test code.

Usually the unit to be tested should be tied around architectural defined interfaces. So if a function forms a publicly visible interface (like the plugin functions do) then this is what you want to test. Otherwise people will start writing spaghetti functions just for not having to write another unit test...

aerostitch:
Sounds good. Do you have specific frameworks in mind that would help around that or a set of rules we should try to stick to when writing those kind of tests @metalstrolch ?

metalstrolch:
Not really.
First thing to do would be to document architecture to find public interfaces. That's why I propose to start with the plugins, as those interfaces are at least easy to identify and would allow to check if all plugins of same kind behave equal. And this would allow to introduce the choosen framework first. Otherwise functions present in module public headers are a good starting point too. It's about defining which function somebody using the component should actually use. This makes the components interface.

Not knowing the core functionality well enough, but I think there things like map data handling, routing algorithm and such could be identified to add another testworthy in-core interfaces.

This is unfortunately good old manual work that cannot be automated. This is why it is that hard for legacy code where nobody documented / enforced architecture.

aerostitch:
Question that might sound a bit stupid, but when you are updating/fixing functions, how do you test that the outcome works as expected?
[...]

PR #938, opened 2019-10-30
add:tests:Add a way to run unit tests in Navit

aerostitch:
Starting to implement the idea proposed in #561
Just creating the PR for now as I need to do something else.
If you want to try it out or fix something in it feel free to. Just let me know so that we don't work at the same time on the same stuff.

We should probably be able to exploit some of the CMake integration for gtest too: https://cmake.org/cmake/help/v3.10/module/GoogleTest.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -909,3 +909,54 @@ add_subdirectory (man)
 if (USE_NATIVE_LANGUAGE_SUPPORT)
 	add_subdirectory (po)
 endif(USE_NATIVE_LANGUAGE_SUPPORT)
+
+# we will use the network to fetch Google Test sources
+# make it possible to disable unit tests when not on network
+option(ENABLE_UNIT_TESTS "Enable unit tests" ON)
+message(STATUS "Enable testing: ${ENABLE_UNIT_TESTS}")
+include(googletest.cmake)
+
+if(ENABLE_UNIT_TESTS)
+	include(ExternalProject)
+	ExternalProject_Add(
+		googletest
+		SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/googletest-src"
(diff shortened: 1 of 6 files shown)


aerostitch:
I seem to have triggered some kind of infinite loop.
I'm pretty noob on cmake so it's going to take me some time to figure out what's wrong probably.

aerostitch:
Found something that seems more relevant to rely on: https://github.com/google/googletest/blob/master/googletest/README.md#incorporating-into-an-existing-cmake-project

I'll have a look tomorrow.

aerostitch:
Used the example. googletest installs. Now the error is a link to target between navit and the tests.

```
[ 33%] No patch step for 'googletest'                                                                                 
[ 44%] Performing update step for 'googletest'             
[ 55%] No configure step for 'googletest'                                                                             
[ 66%] No build step for 'googletest'                                                                                 
[ 77%] No install step for 'googletest'                    
[ 88%] No test step for 'googletest'                       
[100%] Completed 'googletest'      
[100%] Built target googletest                             
-- Found PythonInterp: /usr/bin/python3.7 (found version "3.7.5")                                                     
CMake Error at CMakeLists.txt:960 (target_link_libraries):                                                            
  Target "navit" of type EXECUTABLE may not be linked into another target.                                            
  One may link only to INTERFACE, OBJECT, STATIC or SHARED libraries, or to                                           
  executables with the ENABLE_EXPORTS property set.        
                                                           
                                                                                                                      
-- Configuring incomplete, errors occurred!
```

aerostitch:
Closing the PR for now. This might take a while before I get things running and we'll need to upgrade the WinCE image which is going to be a nightmare.
Plus I'm kind of afraid that will bring a bit too much annoyance, so closing until I have something solid and easy to manage.

PR #940, opened 2019-10-31
fix:various:fix compiler warnings

metalstrolch:
This PR fixes a lot of compiler warnings. The core and the well maintained plug ins build at least under Linux without severe warnings.

### DIFF
--- a/navit/binding/dbus/binding_dbus.c
+++ b/navit/binding/dbus/binding_dbus.c
@@ -626,13 +626,13 @@ static char *get_iter_name(char *type) {
 }
 
 static DBusHandlerResult request_attr_iter(DBusConnection *connection, DBusMessage *message, char *type,
-        struct attr_iter *(*func)(void)) {
+        struct attr_iter *(*func)(void*)) {
     DBusMessage *reply;
     char *iter_name;
     char *opath;
     struct attr_iter *attr_iter;
 
-    attr_iter=(*func)();
+    attr_iter=(*func)(NULL);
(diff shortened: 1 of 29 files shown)


metalstrolch:
A pity CodeFactor doesn't help here. This would be really helpful.

metalstrolch:
The current CodeFactor "findings" are a good example why CodeFactor should not be blocking. The duplicated code is the plug ins. As those are built and linked separately, one cannot just "unify" the code. I think here the code was duplicated for good reasons. If anybody starts up a new plugin he will likely start by copying one of the existing ones. And most likely obody wants to mess up the old one by trying to unify code.

It's good CodeFactor tells this, but it's also good the developer can decide to ignore it or not. Last but not least as the PR has nothing and really nothing to do with fixing such things.

- just to have mentioned this for the die hard's to recognize.

aerostitch:
Codefactor is not blocking. The only one that is blocking is circleci ("build_all" which is the target for all the circleci builds) and you can see it as it says "Required" next to details.

metalstrolch:
Yes luckily CodeFactor is not blocking anymore. But as there is another round of discussions going on in some PRs I do not refer from here, I left the comment to point out why CodeFactor shouldn't be blocking.

Its a good tool but far away from being fool prove.
And still I wonder that none of the tools complains about the very creative casting done in the C parts. This was common back in the days, but wasn't really a good Idea ever since as the compiler could trash the Stack by function pointer casting going wrong. Im planning to address the alignment issues on arm targets caused by this next.

PR #936, opened 2019-10-30
Move checkstyle into the sanity_check phase

aerostitch:
Trying to work on #725

PR #941, opened 2019-11-02
update:CI: skip the build steps if the change is only for documentation

### LINKED ISSUE #925: CI: skip the build steps if the diff from trunk is only for documentation
I think it would be good if we had a script that would be able to detect if the current PR is only for documentation files (`*.md`, `*.rst` and `docs/*` but also give the ability to accept other patterns in the future without having to rewrite everything).
And if it is detected that the PR is only for those documents, exit the build jobs with success (ending up with only the sanity checks that really run, the rest just exiting with a positive return code.

This would save us time on the CircleCI side but also we would need to wait less time for stuffs that don't affect the build.

Any objections?

aerostitch:
Closes #925 

The idea is to skip the build steps if the change is only done on `*.md` or `*.rst` files or is only in the `docs` folder.
Saves CI time and makes the time to have the relevant checks passing shorter. Should help on #910 

Note that the branch should be be up-to-date with trunk to make it work.

Side note: if there is no diff with trunk (aka that's pushed on master or trunk) then we only check for the files in the last commit).
And if we are on a tag, we just exit 1 as we want to build even if the last change is a doc change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,7 @@ jobs:
     <<: *defaults
     steps:
       - checkout
+      - run: if scripts/check_need_build.sh; then circleci step halt; fi
       - run:
           name: Id
           command: cat /etc/*release
@@ -39,6 +40,7 @@ jobs:
     <<: *defaults
     steps:
       - checkout
+      - run: if scripts/check_need_build.sh; then circleci step halt; fi
       - run:
           name: Install doxygen and other essentials
           command: apt-get update && apt-get -y install doxygen ca-certificates git rsync
@@ -54,6 +56,7 @@ jobs:
     machine: true
     steps:
       - checkout
+      - run: if scripts/check_need_build.sh; then circleci step halt; fi
       - run:
           name: install docker
           command: circleci-install docker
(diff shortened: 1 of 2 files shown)


aerostitch:
Testing right now using https://github.com/navit-gps/navit/pull/941/commits/270e1c7c820a57c9c4cd84a3a2c40fe2d2baef86 it almost works.
In build_linux, build-wince and build-windows I run this before the setup_requirements right now so git is not found. Should be an easy fix

aerostitch:
Alright so frigging build_win* are on old ubuntu with old bash that don't accept the declare options to `local` so I need to rewrite a bit the script.

aerostitch:
Also included in this PR the change for storing the checkstyle report that was broken in #936.
https://18961-30791823-gh.circle-artifacts.com/0/reports/checkstyleMain.html

aerostitch:
Alright. I think it's ready to review and merge.
Both filtered and non-filtered cases have been tested in this PR so we should be good to go I think.

.circleci/config.yml
```diff
@@ -14,10 +14,17 @@ jobs:
...
+          name: Store checkstyle report
+          path: navit/android/checkstyle
+          destination: reports
   build_linux:
     <<: *defaults
     steps:
       - checkout
+      - run: |
+          apt-get update && apt-get install -y git
+          if scripts/check_need_build.sh; then circleci step halt; fi
```
pgrandin: Nice trick, didn't know this one

Issue #725, opened 2019-01-07
Revisit code style checks


mvglasow:
Currently my CircleCI builds keep failing because a code file is getting rejected for non-compliant style. Example: https://circleci.com/workflow-run/7ffa71d2-467e-4b51-a7a9-aa81d8053480

That would be the expected behavior if I had introduced these errors, but the file in question is a file I have not even touched in my branch.

Is there a way to limit style checks only to changes which are not yet in master (for branches), or since the last successful master build (for master)? Even if we could just do that at the file level (i.e. excluding files that have not been touched), that would already be an improvement.

mvglasow:
I have just fixed a few files in a branch (every single one flagged by the sanity check), and after re-running CI, I am still getting new style errors that did not get flagged before. Something’s wring here…

aerostitch:
I think #929 should fix your issue.

But I agree that there's something to do here and I was actually creating a GH issue about it so I guess I'll take this one and do the proposal here:

Hi guys,

Right now we have several tools to help make the code better (sanity check for non-java files and checkstyle for java files, codefactor, etc).
The way we set them up has brought some frustrations in the contributions and I feel like we should focus more on getting working code in navit.
What I'm proposing is not to get rid of what we did around the sanity check or codefactor etc but more to make it less mandatory.

I'm fine doing a bunch of PR to clean up some stuffs when I feel like it. I'm not fine with core contributors starting to be frustrated about the tools we put in place.

What I propose is:
* move the sanity check outside of circleci (especially not blocking the builds). It can be a github action for example, and make it not blocking.
* move the java checkstyle outside of the android build. Same it can be in a github action for example but making sure it's not blocking contributions
* revive coverity but make it not blocking

Once again, I'm fine having a PR merged with non-confirming style or pushing a code style fix in a PR of someone else before merging but I think the current way of doing things is bringing more frustration than benefits which was really not the original idea.

Any comments / issues / enhancements proposal around that change in the implementation?

Thanks
Joseph

metalstrolch:
From my point of view, the first thing to do would be to document the style requirements. Or better point to the documentation if it is already there.
Personally I don't have any problems in e.g. running astyle, but having to inspect CI output to find the embedded Java style error or to find out what sanity_check does (for the non "C" coders) is not simple. Nobody likes that kind of surprises. Specially if you want (or have to) write code for architectures / languages you are not familiar with.

So we have:
for C:
[...]

PR #944, opened 2019-11-08
add:doc:document the sanity check steps and add a note about the Java code

aerostitch:
I believe this is the last step for #725 
@metalstrolch does that answer your questions about the style checks?


metalstrolch:
What I'm still missing is the version of checkstyle required. At least Gentoo still ships one too old. And do we really require gradle to run it? On the boxes I usually code, Ill never get gradle. And no I won't run code installed by an obscure script. But hey, I doubt I ever will do Java for Android again here. So at least it's mentioned.

aerostitch:
Trying to address your concern, I'll need more info in #945 for that if you don't mind. I didn't realize that Checkstyle came also separately from Gradle. ;)

PR #947, opened 2019-11-17
wip: remove custom cmake/FindProtobuf-c.cmake

bignaux:
btw , seems internal cmake https://cmake.org/cmake/help/v3.2/module/FindProtobuf.html doesn't work for  libprotobuf-c-dev


jkoan:
Hi @bignaux,
we have FindProtobuf-c.cmake (note the -c) because we need to find the version for c not the standart variant, so i think that this PR is invalid, right? @pgrandin

jkoan:
Closing this because of the reason from the last comment.

PR #963, opened 2020-01-19
Fix:port/android:Use new icon for notification

mvglasow:
Also updates the Android notification icon, which was still missing after #943.


mvglasow:
> **build_all** _Expected — Waiting for status to be reported_ 

Something’s wrong here, everything shows up as successful on CircleCI but Github isn’t picking it up, even after rerunning the tests. Seems I cannot merge, as even a manual merge gets rejected.

jkoan:
> > **build_all** _Expected — Waiting for status to be reported_
> 
> Something’s wrong here, everything shows up as successful on CircleCI but Github isn’t picking it up, even after rerunning the tests. Seems I cannot merge, as even a manual merge gets rejected.

You need to enable Github Checks on CircleCi, this is a gap i really hate, between old style checks(status) and new style checks. 
The Problem only apperes on fors which have there own circleci build (which shuldent happen according to the circleci org settings, but it does)

mvglasow:
Thanks, I just enabled CircleCI checks and restarted the workflow (as GH still did not pick up the result). We’ll see…

jkoan:
> > **build_all** _Expected — Waiting for status to be reported_
> 
> Something’s wrong here, everything shows up as successful on CircleCI but Github isn’t picking it up, even after rerunning the tests. Seems I cannot merge, as even a manual merge gets rejected.

Yes, the check needs to be triggered from github. Will merge the PR without the check right now, but on your next PR it should work as normal

mvglasow:
Thanks!

PR #966, opened 2020-01-19
fix:plugins: too few arguments to functions navit_attr_iter_new and config_attr_iter_new

aerostitch:
In Debian, the build fails with the 2 following errors:
```
/<<PKGBUILDDIR>>/navit/plugin/pedestrian/pedestrian.c: In function ‘pedestrian_navit_init’:
/<<PKGBUILDDIR>>/navit/plugin/pedestrian/pedestrian.c:1341:12: error: too few arguments to function ‘navit_attr_iter_new’
 1341 |     iter = navit_attr_iter_new();
      |            ^~~~~~~~~~~~~~~~~~~
/<<PKGBUILDDIR>>/navit/plugin/pedestrian/pedestrian.c:1400:12: error: too few arguments to function ‘config_attr_iter_new’
 1400 |     iter = config_attr_iter_new();
      |            ^~~~~~~~~~~~~~~~~~~~
```

This PR should fix that.

Note that the following flags are set during the build. Maybe we should consider having those inside our standard build system to avoid surprises (broken in several lines for readability):
```
-fstack-protector-strong -D_FORTIFY_SOURCE=2 -Wall -Wundef -Wcast-align
  -Wpointer-arith -Wno-unused-parameter -Wno-sign-compare
  -Wno-missing-field-initializers
  -Wextra  -Wmissing-prototypes  -Wstrict-prototypes  -Wformat-security
  -g -O2 -Wformat -Werror=format-security -Wdate-time  -fPIC
```
What do you think?

### DIFF
--- a/navit/plugin/pedestrian/pedestrian.c
+++ b/navit/plugin/pedestrian/pedestrian.c
@@ -1338,7 +1338,7 @@ static void pedestrian_navit_init(struct navit *nav) {
 #endif
     transform_set_scale(trans, 16);
     navit_get_attr(nav, attr_layout, &initial_layout, NULL);
-    iter = navit_attr_iter_new();
+    iter = navit_attr_iter_new(NULL);
     while (navit_get_attr(nav, attr_layout, &attr, iter)) {
         if (!strcmp(attr.u.layout->name, "Route")) {
             dbg(lvl_debug, "found %s", attr_to_name(attr.type));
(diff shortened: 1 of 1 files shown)


metalstrolch:
BTW, the compiler settings do not affect if this gets found or not. If the pedestrian plug in doesn't get built, it won't fail. It think it's that simple...

aerostitch:
Thanks. I'll create a PR to enable it on one of the builds then :)

PR #972, opened 2020-01-20
add:ci: enable the pedestrian plugin by default on linux and android

aerostitch:
Hi guys,

I think it would make sense to have the pedestrian module enabled by default.
At least that's how it's configured on Debian.
Wondering if we should have it built by default so that when we distribute the new release it's there.

What do you think?
Are there other plugins you think would be worth enabling by default?

Thanks,
Joseph

### DIFF
--- a/scripts/build_android.sh
+++ b/scripts/build_android.sh
@@ -28,7 +28,13 @@ export GRADLE_OPTS='-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryEr
 echo Run CMake
 test -z "$PKG_CONFIG_LIBDIR" && export PKG_CONFIG_LIBDIR=""     # Force cmake below to run ignore build host libraries when using pkgconfig.
 # Note: If you want to compile against specific target libraries that are searched using pkgconfig, please run this script with variable PKG_CONFIG_LIBDIR set to the appropriate path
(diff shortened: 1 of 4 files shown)


aerostitch:
Seems it fails to compile on win*. Was it supposed to compile on those platforms?

metalstrolch:
What does it do? I Didn't get it up to now. It's not pedestrian routing, isn't it? But obviously it:
a) doesn't build under win32
b) is missed by nobody.

aerostitch:
I thought it was the pedestrian routing!
And no, aside from people using the debian package (raspbian & co) or compilation from source with explicit flag set.
And it was broken since #966 (but was compiling on last release so not sure exactly what happened)

metalstrolch:
I broke it by forgetting it during my refactoring to remove compiler warnings. As it was not built by CI this didn't show up. This was fixed in #966. But I don't know what it is good for. And I don't think it built for windows last release.

mvglasow:
As far as I remember, the pedestrian plugin was/is this one: https://wiki.navit-project.org/index.php/Augmented_Reality

As far as I know it has not seen any maintenance (I have never used it and am not sure it still works at all) and got removed from the standard Android build sometime during the last 1–2 years. I am just surprised to learn it is part of the Debian build. Grepping the commit history for `pedestrian` should provide some hints at the history of this plugin.

aerostitch:
Alright, if that's the one then it's probably not advisable. I disabled it in Debian.

PR #988, opened 2020-04-12
fix:ci:fix sailfish artifacts uploading

metalstrolch:
For whatever reasons the sailfish rpms are not visible in circleci
any more when they used to be. Maybe because of indirect path usage on
uploading them. This fixes that.


jkoan:
Hi, the reason was that circleci cleaned up old artifacts as far I remember

jkoan:
From Circleci Build-Job Page:

"After 30 days, all artifacts associated with a build automatically expire. Since this run is over 30 days old, the artifacts may have been deleted."

PR #995, opened 2020-04-29
fix:scripts:Correct conditionals in android keystore setup

### LINKED ISSUE #994: Android CI build fails on foreign repositories
Currently the circleci build for android fails on non navit-gps repositories because of issues during key setup for playstore. #993 is a good example. The branch pushed to navit-gps instead builds fine.

metalstrolch:
This commit fixes the conditionals used in setup script for android keystore causing the android build to fail if there are no android keys there. This is usually the case on all repos not being on navit-gps

Fixes #994


fastlane/Fastfile
```diff
@@ -31,15 +31,21 @@ platform :android do
...
-      upload_to_play_store( track: 'beta', 
-      						json_key: 'key.json', 
+      gradle(
+        task: 'assemble',
+        build_type: 'Release'
+      )
+    else
+      gradle(
+        task: 'assemble',
+        build_type: 'nonmaster'
```
jkoan, line 43: Why not just usw Debug?

metalstrolch: As there is no "Debug" target there, and it's not a "Debug" build we're doing here. It's just not signing the result as we don't have a proper key.

But I'm not at all the Android / Fastlane / Gradle... expert to really know what's going on here. All I want to do is to fix the annoying CI build fail.

metalstrolch:
OK, changed it a bit. Does the usual build if scripts/setup_publish_keys.sh previously was able to open a key ring. So on all branches on "navit-gps" it does a "release" build. On all other repos it does a "nosigned" build.
Verified with this branch on "metalstrolch" and on "navit-gps".

Ready to be merged from my side.

metalstrolch:
Why isn't this report one status? and which? Bummer

metalstrolch:
Well, this builds now and CI shows it's green. But the build_all step waits for something.

As I cannot inspect CI config I don't know what it is waiting for. Can anyone help?

jkoan:
> Well, this builds now and CI shows it's green. But the build_all step waits for something.
> 
> As I cannot inspect CI config I don't know what it is waiting for. Can anyone help?

Please enable the Github Checks functionality on CircleCI

metalstrolch:
Nope. Enabling github checks doesn't fix that. It's some step of the build that's not running most probably. The change disables apk signing for android if no keys are available. Eventually there's a check configured that awaits the signed apk somewhere. Would be interesting to know navit-gps merge check settings expect to be there to consider build_all to finish.

The same branch pushed to navit-gps:navit would build properly (as keys are available).

jkoan:
> Would be interesting to know navit-gps merge check settings expect to be there to consider build_all to finish.
> 
Bildschirmfoto vom 2020-05-01 19-12-49
Thats all, its only that we chack build_all passed, which is activated by circleci when github checks is enabled. Could you please try to close this PR and push a new branch and open a new PR ? Of cause after github checks are enabled.

jkoan:
> Would be interesting to know navit-gps merge check settings expect to be there to consider build_all to finish.
> 
Bildschirmfoto vom 2020-05-01 19-12-49
Thats all, its only that we chack build_all passed, which is activated by circleci when github checks is enabled. Could you please try to close this PR and push a new branch and open a new PR ? Of cause after github checks are enabled.

metalstrolch:
See #999, #1000 and #1001 for different behaviour.

PR #998, opened 2020-05-01
add:graphics:gtk_drawing_area:textured polygons

metalstrolch:
This adds the support for textured polygons on gtk_drawing_area.


metalstrolch:
Then it may look like that.
ammersee

PR #1002, opened 2020-05-02
Improve layout

jfitie:
Visual improvements to `navit_layout_car_shipped.xml` layout, mainly tweaked colors for some polygons. Also applied the same (tinted) changes to `navit_layout_car_dark_shipped.xml` for consistency.

### DIFF
--- a/navit/navit_layout_car_dark_shipped.xml
+++ b/navit/navit_layout_car_dark_shipped.xml
@@ -198,7 +198,7 @@
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_construction" order="10-">
-			<polyline color="#060e0b99"/>
+			<polygon color="#060e0b99"/>
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_range" order="10-">
--- a/navit/navit_layout_car_dark_shipped.xml
+++ b/navit/navit_layout_car_dark_shipped.xml
@@ -447,7 +447,7 @@
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_sports_track" order="10-">
-			<polyline color="#051109" width="1"/>
+			<polygon color="#05110999"/>
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_barracks" order="10-">
--- a/navit/navit_layout_car_dark_shipped.xml
+++ b/navit/navit_layout_car_dark_shipped.xml
@@ -447,7 +447,7 @@
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_sports_track" order="10-">
-			<polygon color="#05110999"/>
+			<polygon color="#050b08"/>
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_barracks" order="10-">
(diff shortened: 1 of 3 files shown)


navit/navit_layout_car_shipped.xml
```diff
@@ -175,11 +175,11 @@
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_construction" order="10-">
-			<polygon color="#f47216"/>
+			<polyline color="#ff3333"/>
```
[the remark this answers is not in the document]
metalstrolch: If you think, poly_construction is used to "overlay" other elements to "tint" the rendering, the suggest adding transparency.

jfitie:
Thanks for the review and info, @metalstrolch. Please review these changes.

navit/navit_layout_car_dark_shipped.xml
```diff
@@ -80,7 +80,7 @@
 			<text text_size="5"/>
```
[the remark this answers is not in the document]
jfitie: Dark colors according to https://github.com/jfitie/navit/blob/b57379a5349788ee627932dee86919937e252827/navit/navit_layout_car_dark_shipped.xml#L70

jfitie:
Screenshots: https://www.imgur.com/a/zhNNxZX
fyi @metalstrolch

navit/navit_layout_car_shipped.xml
```diff
@@ -175,11 +175,11 @@
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_construction" order="10-">
-			<polygon color="#f47216"/>
+			<polygon color="#cc776699"/>
```
metalstrolch, line 178: OK, it doesn't hurt to have transparency here as poly_construction is used quite seldom. And I can imagine a lot of mapping errors where poly_construction overlaps something it shouldn't according to OSM.

navit/navit_layout_car_shipped.xml
```diff
@@ -444,15 +445,16 @@
 			<text text_size="5"/>
 		</itemgra>
 		<itemgra item_types="poly_sports_track" order="10-">
-			<polygon color="#b6554e"/>
+			<polygon color="#bb885599"/>
```
metalstrolch, line 448: This again is something where transparency "should not" be required as this "should not" overlap anything. However do you have any good example on where you need transparency on poly_sports_track? Because I think most probably this are only mapping mistakes in OSM.

jfitie: There are many good examples.
Here are a few to make my point:
132635227
420091556
179372237

metalstrolch: According to https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dtrack it's unclear how those things should be tagged and rendered. They are allowed either as line or as area. Since multipolygons are always areas, Id support both. I think adding a polygon and a polyline could be a good compromise. However loop tracks are then as up to now rendered as areas albeit they shouldn't as long as area=yes is missing.
think of:
<polygon color="#bb8855"/>
<polyline color="#bb8855"/>

We need to address the "closed loop but no area" thing in maptool by giving the non area tracks their own map id. E.g. sports_track without "poly_" and decide on the presence of area= tag which one to use.

But still I think the use in https://www.openstreetmap.org/way/179372237 is a mapping error. As it is somehow used like a landuse not giving the outline of the actual track nor the track itselve.
So removing the transparency should be fine.

metalstrolch:
leisure=track ambiguity can be resolved after merge of #1005

PR #1004, opened 2020-05-03
add:graphics:gtk_drawing_area:textured polygons

metalstrolch:
This adds the support for testured polygons on gtk_drawing_area.


metalstrolch:
And off we go. Gtk can have textures as well.

PR #1020, opened 2020-06-14
Enhancement:layout_car:Add stripes to danger_area

metalstrolch:
This pull request adds a red stripe texture that is used for poly_danger_area on car layout.
Additionally to that, a lightly transparent red border is added to it.

To see the diagonal stripes one needs to use either gtk_drawing_area or qt5 graphics as only those currently support textured polygons.


metalstrolch:
poly_danger_area can now look the following under supporting graphics
danger_area_texture
On non supporting graphics it will look like
danger_area_notexture

PR #1024, opened 2020-06-17
Small color tweaks and added polylines

jfitie:
Small color tweaks and added polylines


jfitie:
@metalstrolch Then again, there currently already are polylines around `building`, `meadow`, `farm`, and more. Except for probably `wood`, most polylines added in this PR are on polygons not occuring very frequently in comparison. Please let me know how to proceed.

metalstrolch:
My notice is more or less about two things:
Please keep in mind and check your self if (and how much) performance impact a layout change has.
And of course to:
Wake up the users of low end devices to check and complain early.

So don't stop making the default layouts more beautiful. Just be aware that there could come a point making some of the users complain about decreased performance. But after all everybody can use his private layout.

Despite this, I'm thinking of adding a "outline" property to the polygon itemgra, to allow drawing a polygon with outline in one step by the graphics. All modern graphic systems could do so. even windows (except CE) This would be really simple for flat view, but fail badly for 3d view because of the scaling of the outline. Let's see.

PR #1027, opened 2020-06-20
Android: Activate downloaded maps automatically

jfitie:
Activate downloaded maps automatically on Android by including the `/sdcard/Download` directory.
This is done so that maps downloaded in the browser, f.ex. using the Planet Extractor, which go to `/sdcard/Download` by default, are activated automatically without having to change the map location in de UI. This way, maps can be downloaded in the app using the UI, and maps can also be downloaded through the browser. Both will _just work™_.


lains:
Hello. Thanks for this Pull Request.
Just a thought...
Have you tested what happens if a bin file is downloaded (but is not at all a map)?
I hope navit does ignore the file properly without crashing or ignoring all further maps.

Thanks

jfitie:
@lains Thanks for your comment. Please see the last commit. I think this should solve that possible issue.

PR #1029, opened 2020-06-24
Android: Remove map download size limit on modern Android versions

jfitie:
Android: Remove map download size limit on modern Android versions


jkoan:
We cant do this as it a limitation of the underlying file system on old android devices. The right solution would be to implement something that splits up a map on the fly into chunks that are smaller than 1 Gig so so to be save on all devices and store them info separate files.

jfitie:
@jkoan Please re-review. What do you think of this solution?

mvglasow:
On a side note, we used to have a test instance of the map server which could split maps into smaller files. Maybe that would help here.

The server’s been in test mode for a couple years (I think it went live in 2012/13), now it is just telling me Access forbidden. Maps downloaded from that server worked with no issues, though (I have had Europe maps from that server on my phone since it was set up). Maybe we could finally move that feature on the production server, then have Navit specify the size limit of the file system in its request. “Please select a smaller map” isn’t really helpful when I simply need that big one…

jfitie:
This PR was meant as a quick fix to allow larger map downloads on modern Android versions supporting files larger than 3.8GiB. This would solve this issue for me personally since I currently cannot download some of the maps I require. If this is an acceptabele (temporary) solution, go ahead and merge this. If instead infrastructural changes are required, please close this PR, as I don't have enough knowledge of Navit to work on that.

jfitie:
OK, let's make this PR even simpler and just remove the map download size limit on modern Android versions. Even downloading the whole world (at ~36GiB) is not an issue on these devices anymore.

I would like to keep in the improved expression for checking map size (```mEstSizeBytes >= Math.pow(2,32) * 0.95```), as the old method (```mEstSizeBytes / 1024 / 1024 / 950 >= 4```) seems a bit hacky and uses an arbitrary number (```950```).

So, in short, let's consider doing
```
int osver = Integer.parseInt("0" + android.os.Build.VERSION.RELEASE.replaceAll("[^0-9]*([0-9]*).*", "$1"));
int mi = Integer.parseInt(mapIndex);
if (osver < 7 && NavitMapDownloader.osm_maps[mi].mEstSizeBytes >= Math.pow(2,32) * 0.95) {
 ...
```

@jkoan @mvglasow Please advise.

jandegr:
Hi,

I think you can verify that the map location still corresponds to the default path, it is the only way I can think of to make sure maps will not be stored on a fat32 sd card.

jkoan:
Mergein now, as the solution seems good enough with what we have so far. But we should look into solutions where we can store one map into chunks smaller than 2GB on the fly.

PR #1051, opened 2020-08-22
Inkscape svg2png

lufv:
Inkscape doesn't use --without-gui argument anymore


jkoan:
Hi @lufv 
Thank you for your work.
I saw the patch of Debian here: https://salsa.debian.org/debian/navit/-/blob/master/debian/patches/inkscape-1.0.patch
I think the Debian one would be more compleat, right?
Also, for me its more readable then your Solution.

lufv:
> Hi @lufv
> Thank you for your work.
> I saw the patch of Debian here: https://salsa.debian.org/debian/navit/-/blob/master/debian/patches/inkscape-1.0.patch
> I think the Debian one would be more compleat, right?
> Also, for me its more readable then your Solution.

Yes, I think it looks okay!

lufv:
I guess it also explains why I didn't get any PNG files after my patch. Will check if after this one it'll work

jkoan:
Obsolete as of #1056

PR #1061, opened 2020-10-14
add:graphics:svg_debug:Add svg_debugging plugin

jkoan:
This Pull request adds a graphic plugin to debug other graphics plugins. It acts as a proxy between the graphic plugin under test and the navit core.

It can be configured via the attibutes `name` and `outputdir`. The `name` is the name of the Plugin which should be proxied, and `outputdir` is the part where the svg images should be saved to.

TODOs for later are mainly:

- overlay support
- improved font support (currently only none rotated text is supported)

here's a svg from the Plugin (Github don't like raw svg...):
[svg_debug_frame_5.zip](https://github.com/navit-gps/navit/files/5378043/svg_debug_frame_5.zip)


hoehnp (approved):
looks good to me

Issue #1032, opened 2020-08-05
[Enhancement] Please add integration with GeoClue location service.


Baggypants:
GeoClue allows applications to integrate with multiple location sources, such as Mozilla Location Service, 3G, GeoIP and various GPS systems. Crucially for me it is being provided as the main location source in the Pinephone Mobian distribution and I'd like to use Navit for it. 

GeoClue https://gitlab.freedesktop.org/geoclue/geoclue/-/wikis/home

Mobian Location docmentation for Pinephone https://wiki.mobian-project.org/doku.php?id=location

Pinephone Navit discussion https://forum.pine64.org/showthread.php?tid=10697&pid=74115#pid74115

mvglasow:
Just for the background, we are talking primarily about the Linux build, correct? Then GeoClue would essentially replace gpsd or whatever other location source we support on Linux.

Sounds like we are talking about another `vehicle` implementation (Navit’s way of encapsulating the various location sources found in the wild). That should be sorted out on the architecture side already.

What platforms would we need to support with this?

Baggypants:
Yes, it would be targeting Linux and would be an alternative source to gpsd or reading nmea data out of a serial device on that platform. gpsd and GeoClue have different use cases though. gpsd allows location data to be collected and transmitted across a network. GeoClue aggregates different sources of location data on a local machine to one interface.

jkoan:
I have already looked into the GeoClue API, and from what i see i think it should be quite strait forword. As @mvglasow said this would be another vehicle plugin for Navit. 
My only concern so far is that we don't have the real hardware (Pinephone or some other device which has a image with GeoClue). I ordered a PineTab from the first batch, so i could probably sort some things out, but the PineTab doesn't have GNSS as the PinePhone has.
Do you know if there is any OS to test/prototype with, or some sort of fake Data feedin? Probably we could feed a pipe into GeoClue and feed the pipes input from previous collected nema data. But this approche would take at least some time setting up all this, and is probably a bit sluggish.

metalstrolch:
GeoClue is used for Qt location services on SailfishOs. Don't know if it is v1 or v2 unfortunately. Since the development headers for GeoClue should be available, one might test a GeoClue vehicle plug in on Sailfish OS devices. It's no "win" for SailfishOS, as the Qt5 vehicle is already there and working, but it's OK to give it a try. Sailfish OS is "just Linux" in all cases.
[...]

PR #1063, opened 2020-10-24
add:vehicle:vehicle_geoclue:Add Vehicle geoclue

jkoan:
This PR adds the new Geoclue vehicle.
Geoclue is a Position Libary which is available for at least Gnome based Desktops and also more importantly upcoming Linux Phones like the [PinePhone](https://wiki.pine64.org/index.php/PinePhone) or [Librem 5](https://en.wikipedia.org/wiki/Librem_5) include it as well (with `phosh` distro).

[Geoclue](https://gitlab.freedesktop.org/geoclue/geoclue/-/blob/master/README.md) also has an fallback to GeoIP or CellID (when supported) so the user should be presented with a location roughly in the neighborhood any time, even on PCs.
This is also important since i don't happen to have such a device around to test, but i have a PineTab which can be used for WiFi based location. In my around area we dont have much geotaged WiFis, but some, so i was able to see Navit jumping around. When the coverage would be better i really think this vehicle can be fluid.

Some things are not perfect however:
- geoclue needs an Internet connection for WiFi based location
- geoclue falls back to GeoIP when no WiFi based location could be found. In my case (because of my mobilephone hotspot) this was about 150km off. But with DSL (House internet) the GeoIP is much more accurate.
- In #1032 @Baggypants reported a crash, but i wasn't able to reproduce it at all

Tested:
- on x86 PC (AMD Ryzen with Manjaro Linux Gnome, Linux 5.8.16, geoclue 2.5.6 )
- PineTab


fortysixandtwo:
Hey, I'm currently trying to build this on my pinephone (running mobian/debian).

And I'm having trouble generating the pngs with the `convert` provided by imagemagick-6q.16.
Since you tested this on the PineTab (which essentially runs the same software)
I was wondering which imagemagick you are using/how you got it to build on debian (see #1065)

fortysixandtwo:
For the people who want to build this:
I had to rebase on the trunk in order to get the fixes for inkscape 1.0 usage (otherwise I would be calling inkscape with `--export-png`).

I guess I need to somehow configure the geoclue vehicle?
Because currently I'm getting no indication that navit has any location information.

Running `DISPLAY=:0 G_MESSAGES_DEBUG=all ~/install/bin/navit -d 3 2>&1 | tee navit.log` and then grepping for geoclue
I only get a single line:
```
50:info:navit:plugin_new:found plugin module file [19]: '/home/mobian/install/lib64/navit/vehicle/libvehicle_geoclue.so'
```

fortysixandtwo:
Ok, I'm very happy to report that it actually works after adding
`<vehicle name="Geoclue" profilename="car" enabled="yes" source="geoclue://" active="1"/>`
to my configuration!

jkoan:
> For the people who want to build this:
> I had to rebase on the trunk in order to get the fixes for inkscape 1.0 usage (otherwise I would be calling inkscape with `--export-png`).
> 
> I guess I need to somehow configure the geoclue vehicle?
> Because currently I'm getting no indication that navit has any location information.
> 
> Running `DISPLAY=:0 G_MESSAGES_DEBUG=all ~/install/bin/navit -d 3 2>&1 | tee navit.log` and then grepping for geoclue
> I only get a single line:
> 
> ```
> 50:info:navit:plugin_new:found plugin module file [19]: '/home/mobian/install/lib64/navit/vehicle/libvehicle_geoclue.so'
> ```

yes, sorry, i forgot to rebase before uploading. For a full config look into the issue #1032, i uploaded my config there

fortysixandtwo:
Last time I tried, my position was not actually updating (and the first time I was stationary).
Will test again soon and report my findings.

Issue #1054, opened 2020-08-29
Background usage on Android


jkoan:
### To reproduce

1. Start navit as normal
2. Set a destination and wait until the first tss is done
3. Lock the Android device

### Expected result

I expacted that the navigation keeps going and announces new maneuvers in time to react to it. But I normally don't use the display so I don't get distracted.

### Actual result

Since Android API level 26 Android restrictions Background location requests to save battery.
https://developer.android.com/training/location/background
This results in navit announcing nothing at all and keeps stuck until unlocking the display again

### Software environment

Android 10 Oneplus Nord

jandegr:
Hi,

you can add `<uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION" />
` in the manifest.

Afterwards I had to manually change the permission from `only when using the app` to `always` in the phone's android settings, making navit ask for it itself will require a little bit of coding and can still be done after you confirm the addition to the manifest + manual setting works.

I tested by upgrading, maybe for a fresh install there might no extra coding needed.

PR #1064, opened 2020-10-25
fix:android:Add possibility to use background position usage

jkoan:
This PR should fix the [issue](#1054 ) that android won't ask for Background GPS usage and ask again when restarting navit because ACCESS_BACKGROUND_LOCATION is not checked in verifyPermissions.


jkoan:
@jandegr could you take a look?

PR #1066, opened 2020-10-25
fix:build:core:Fix two issues where the wrong enum type is used

jkoan:
Fix two issues where the wrong enum type is used, but both have the same value of 0


jkoan:
as its a nobrainer in my eyes i merge it now

PR #1071, opened 2020-11-08
Add:port/android:Adaptive app icon

mvglasow:
Starting with Android 10 (or 9, which I skipped), Android requires that each app supply an adaptive icon.

An adaptive icon consists of a background and a foreground. The background can be a single color or an image and is clipped to the user-selected icon shape. The foreground is an image (usually transparent) which is displayed on top of the background. Assembly is done by the OS, which may include visual effects such as a parallax effect (foreground and background moving against each other as the device is moved).

Non-adaptive icons are scaled down to fit inside the clip area and placed on top of a white background, which looks bad for our icon.

The design size for an adaptive icon is 108×108 px; the foreground may be clipped to a radius of 33 px around the center (how much exactly is shown depends on the selected icon shape).

I have used the orange gradient from the familiar Navit logo as the background and the Navit compass logo as the foreground. When resizing the chiclet icon to 66×66, the logo itself is about 62×62 in size, so I used that for the logo, leaving the familiar margin around the logo also for adaptive icons.

Here is how it looks when selecting round icons:
Screenshot_20201108-204218_Trebuchet

The non-adaptive version of the icon still needs to be provided for Android versions prior to 8; I have left it unchanged.


jkoan (approved):
Looks really good, thx :sunglasses:

Issue #1041, opened 2020-08-11
Sanity check: revise checkstyle rules for Java code


mvglasow:
I just got a bunch of ant checkstyle errors from CI. While a style check certainly makes sense (I prefer having tidy code myself), checkstyle also flagged a few things which I do not quite consider errors or bad style:

> Member name 'context' must match pattern '^m[A-Z][a-zA-Z0-9]*$'. [MemberName]

Personally, I have long stopped prefixing member variables with `m`. IMHO the prefix is similar in spirit to what is known as Hungarian notation in C (`lpszText` rather than just `text`, to indicate that it is a **l**ong **p**ointer to a **s**tring which is **z**ero terminated), which was once widespread (Microsoft made extensive use of that) but is now mostly discouraged and rarely used—possibly connected to the rise of IDEs which offer syntax highlighting and additional information when hovering the mouse over a piece of code. As for ourselves, we are not enforcing (or even using) it in C either.

> 'if' construct must use '{}'s. [NeedBraces]

The various loop statements get flagged in a similar way. AFAIK we are not enforcing braces in C, so it doesn’t make sense in Java. I don’t see a real benefit in requiring braces unless the language does (shorter diffs when going from a single-statement conditional to an entire block are not all that important).

> Distance between variable 'cn' declaration and its first usage is 4, but allowed 3.  Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value). [VariableDeclarationUsageDistance]

That is in fact contrary to our C style guide, in which we require variable declarations at the top of each function body, before any other statement (which AFAIK is a limitation of some older C flavors we still need to support on some platforms). While such requirements do not apply to Java (or even to C code that is specific to Android), I would suggest we allow keeping variable declarations at the top of blocks. Those who prefer it in between can still do so, and in this case keeping the variable declaration close to its first usage makes sense. Also, I’d recommend sticking to one way or the other—but I wouldn’t spend too much energy about crafting checkstyle rules to enforce that.

PR #1079, opened 2020-12-20
Fix:ci:Relax checkstyle rules for Java code, see #1041

mvglasow:
This relaxes some checkstyle rules to be in line with what we require for C code (see #1041):

* MemberName: No longer enforce member names à la `mFoobar` (which is reminiscent of a now-discouraged practice in C and which we don’t enforce or even use there); now the only requirement is that member names must start with a lowercase letter (as identifiers starting with an uppercase letter are for class names) and contain nothing other than letters and numbers. Hence any of `mFoobar`, `foobar` or `fooBar` is now accepted. `Foobar` or `foo_bar` would get flagged as before.
* NeedBraces: We don’t require braces around single-statement blocks in C (and usually omit them there), so the same now applies to Java code.
* VariableDeclarationUsageDistance: No longer enforced—in C we require variable declaration at the very top due to compiler limitation, so it would be consistent to do the same in Java, though not a requirement. If someone really wants to put effort into this, they can design a rule that says “either declare variables at the top, or no more than 3 lines from first usage”, though I am not sure this can be done in checkstyle.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -59,9 +59,6 @@
             <property name="tokens"
              value="LITERAL_TRY, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_SWITCH"/>
         </module>
-        <module name="NeedBraces">
-            <property name="severity" value="error"/>
-        </module>
         <module name="LeftCurly"/>
         <module name="RightCurly">
             <property name="id" value="RightCurlySame"/>
@@ -141,7 +138,7 @@
             <property name="severity" value="error"/>
         </module>
         <module name="MemberName">
-            <property name="format" value="^m[A-Z][a-zA-Z0-9]*$"/>
+            <property name="format" value="^[a-z][a-zA-Z0-9]*$"/>
             <message key="name.invalidPattern"
              value="Member name ''{0}'' must match pattern ''{1}''."/>
             <property name="severity" value="error"/>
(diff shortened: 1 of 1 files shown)


checkstyle.xml
```diff
@@ -59,9 +59,6 @@
             <property name="tokens"
              value="LITERAL_TRY, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_SWITCH"/>
         </module>
-        <module name="NeedBraces">
-            <property name="severity" value="error"/>
-        </module>
```
jkoan, line 64: In my Opinion brackets should be required in c as well, because mixing styles is not a good Idea.

mvglasow: I see a lot of typing coming our way then… instead of

    if (foo)
        do_stuff();

I now have to type

    if (foo) {
        do_stuff();
    }

Which means more lines for the same code, and less code that will fit on a page of screen (bad for legibility).

It’s not really mixing styles: braces around single-statement blocks are optional in both C and Java, but mandatory around multi-statement blocks (as there is no other way to tell where the block ends). So the style is either “always use braces” or “use braces where necessary”.

The only drawback of omitting braces around single-block statements is that diffs can become slightly cluttered when a single-statement block is later extended to multiple statements: not only is there a line for each added statement, but also one for the block statement (with the added opening brace) and another for the newly-added closing brace. But I think this is outweighed by the benefits of less typing and better legibility.

jkoan: okay, then i am okay with "use braces where necessary"

jkoan (approved):

aerostitch (approved):

PR #1082, opened 2021-01-09
fix:build:android: Fix VersionCode after 31.12.2020

jkoan:
Currently the VersionCode has the Format yyMMddHHmm
which breaked at 01.01.2021 because the maximum allowed Version by
Google Play is 2100000000 because of the limitation of older Android
Devices.
Out currently highest VersionCode is 2012201504 and we need to keep it
continuous, so the new Format would be yyyyMMddHH.
Of cause this has the limitation that we can "only" create one
Versioncode per Hour.
But this is the the best @hoehnp and I came up yesterday.
As a result a PR will follow which will remove the android build from
the master build and instead will add it to a scheduled build (nightly
build)

### DIFF
--- a/gradle/scripts/git-scm-version.gradle
+++ b/gradle/scripts/git-scm-version.gradle
@@ -15,7 +15,7 @@ import java.time.format.DateTimeFormatter
 ext {
     git = Grgit.open(currentDir: projectDir)
     gitVersionName = git.describe(match: ["v[0-9.rc]*"])
-    gitVersionCode = Integer.parseInt(DateTimeFormatter.ofPattern("yyMMddHHmm").format(git.head().dateTime))
+    gitVersionCode = Integer.parseInt(DateTimeFormatter.ofPattern("yyyyMMddHH").format(git.head().dateTime))
 }
 
 task printVersion() {
--- a/gradle/scripts/git-scm-version.gradle
+++ b/gradle/scripts/git-scm-version.gradle
@@ -15,7 +15,11 @@ import java.time.format.DateTimeFormatter
 ext {
     git = Grgit.open(currentDir: projectDir)
     gitVersionName = git.describe(match: ["v[0-9.rc]*"])
-    gitVersionCode = Integer.parseInt(DateTimeFormatter.ofPattern("yyyyMMddHH").format(git.head().dateTime))
+    hh = Integer.parseInt(DateTimeFormatter.ofPattern("HH").format(git.head().dateTime))
+    mm = Integer.parseInt(DateTimeFormatter.ofPattern("mm").format(git.head().dateTime))
+    hhmm = Math.round((hh*4)+(mm/15)).toString()
+    yyyyMMdd = DateTimeFormatter.ofPattern("yyyyMMdd").format(git.head().dateTime)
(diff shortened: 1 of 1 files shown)


mvglasow:
So the highest permissible version code is `21 00 00 00 00`, and we are currently at `20 12 20 15 04`, i.e. we have some 87 million version codes left. Thank God this came back to bite us early enough, i.e. while the original devs are still around (the more seasoned folks here may still remember the Y2K bug, and we had a similar one hovering around).

Joking aside, there is plenty of “air” in the version codes. Starting at the last used version code and adding one per hour would give us roughly a millennium before we hit the barrier (vs. 79 years for the `yyyyMMddHH` approach).

First off, where is the issue in two (or more) git commits building with the same version code? For the individual dev it might not be an issue at all—on my other projects I increase the version code once per release, and the time between releases can be long. The only issue I can see is app stores such as Google Play and F-Droid, which might require each version to have a version code higher than the previous one.

That would essentially mean we need to increase the version code only when we are about to publish an update to one of these stores, i.e. when we release a new stable (or testing) version. How often do we do that?

Then we could go for a version code schema of `20 aa bb cc dd`, where
* `aa` is the major version plus 13 (because we are already using 12)
* `bb` is the minor version
* `cc` is the release
* `dd` can be a sub-release identifier, e.g. a nightly build between releases

That would allow for 100 nightlies per release, 100 releases per minor version, 100 minor versions per major version, and major versions up to 86. We’d then need a way to determine when a new release has happened, and increase the version code—we can either set the new version code only when we are about to release, or first release and then increase the least significant portion by 1 as a preliminary version code for the next release.

I am somewhat sceptical about limiting builds to once per night and would see that as a last resort when nothing else works. If we stick with time-based version codes, I would throttle builds to one time per version code validity.

In that case, a quick improvement would be `yyyyMMddQQ` for a scheme, where `QQ = HH * 4 + mm / 15`, i.e, a quarter of an hour. While HH will always be in the 0–23 range, QQ would be in the 0–97 range (or 0–98, if we round rather than truncate) and allow for one released version every 15 minutes.

We could take this one step further and go with `yyyyMMttFF`, where `tt = dd * 3 + HH / 8`, i.e. a third of a day, and `FF = (HH % 8) * 12 + mm / 5`, i.e. a five-minute period in a third of a day. Not as straightforward to read as a date, but it would allow for a new version every 5 minutes. This is close to what our CI pipeline can handle and should not be much of a limitation any more.

hoehnp:
@mvglasow, interesting suggestion I couldnt come up with such a "complicated" calculations :-)

jkoan:
Noe i Implemented the a VersionCode every 15min's Version as it seems to be a good compromise between Readability and usability. For dc7d5826cd44b82b73b7fa9d759f95c67400e95a:
```
jan@jan-pc:~/git/navit-new$ ./gradlew printVersion

> Configure project :navit:android
Version Name: v0.5.5-34-gdc7d582
Version Code: 2021011180

> Task :navit:android:printVersion UP-TO-DATE

BUILD SUCCESSFUL in 599ms
jan@jan-pc:~/git/navit-new$ 
```

jkoan:
@mvglasow , @jandegr any concerns about merging?

mvglasow (approved):

PR #1087, opened 2021-02-07
Fix:build/android:do not double-define getcwd on recent NDK versions

### LINKED ISSUE #1043: Android build fails on recent NDK versions
After upgrading from NDK 18.1.5063045 to 21.0.6113669, the Android build fails with an error:

```
  In file included from /home/me/src/navit/navit/support/gettext_intl/dcigettext.c:91:
  In file included from ../../../../../support/gettext_intl/gettextP.h:206:
  /home/me/src/navit/navit/support/gettext_intl/dcigettext.c:160:14: warning: this function declaration is not a prototype [-Wstrict-prototypes]
  char *getcwd ();
               ^
                void
  /home/me/src/navit/navit/support/gettext_intl/dcigettext.c:160:7: error: redeclaration of 'getcwd' must have the 'overloadable' attribute
  char *getcwd ();
        ^
  /home/me/tools/android-sdk-linux_86/ndk/21.0.6113669/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/bits/fortify/unistd.h:105:7: note: previous overload of function is here
  char* getcwd(char* const __pass_object_size buf, size_t size)
        ^
```

Apparently recent versions of the NDK (starting somewhere between 18.1 and 21.0) supply an implementation of `getcwd`, which clashes with our dummy, and we need to work around that. Anyone have an idea how to fix this?

mvglasow:
Fixes #1043

Holding this back while jkoan is still investigating the build issues on F-Droid.

I have successfully built this locally on NDK 21, then after downgrading to NDK 18 (which had not experienced the issue this commit intends to fix).

@jandegr I can’t officially add you as a reviewer here, but your input would be appreciated (e.g. if I’ve overlooked something).

### DIFF
--- a/navit/support/gettext_intl/dcigettext.c
+++ b/navit/support/gettext_intl/dcigettext.c
@@ -156,7 +156,7 @@ char *getwd ();
 # else
 #  if VMS
 #   define getcwd(buf, max) (getcwd) (buf, max, 0)
-#  else
+#  elif !(defined(__clang__) && defined(__BIONIC_FORTIFY))
 char *getcwd ();
 #  endif
 # endif


jkoan (approved):

m-kuhn:
Bionic is android specific https://en.m.wikipedia.org/wiki/Bionic_(software)

Also see the discussion linked above

PR #1089, opened 2021-02-14
make it switch to nightlayout when we are in a tunnel

OLFDB:
Nightlayout is enabled when available and streetflags & AF_UNDERGROUND is true


navit/navit.c
```diff
@@ -3504,6 +3504,27 @@ void navit_layout_switch(struct navit *n) {
                 && valid_attr.u.num==attr_position_valid_invalid) {
             return; //No valid fix yet
         }
+
+        int *streetflags = tracking_get_current_flags(n->tracking);
```
metalstrolch, line 3508: Suggest to move variable declaration to top of function. We still support platforms with old C compilers not supporting this.

navit/navit.c
```diff
@@ -3504,6 +3504,27 @@ void navit_layout_switch(struct navit *n) {
...
+
+        if (streetflags != NULL) {
+
+            dbg(lvl_debug,"streetflags: %i", *streetflags);
+
+            // We are in  a tunnel and we have a nightlayout -> switch to nightlayout
+            if ((*streetflags & AF_UNDERGROUND) && l->nightname) {
```
metalstrolch, line 3515: Suggest factoring out check for "l->nightname" or check for "AF_UNDERGROUND". Both allow to get rid of the ugly else if.

navit/navit.c
```diff
@@ -3504,6 +3504,27 @@ void navit_layout_switch(struct navit *n) {
...
+                navit_set_layout_by_name(n, l->nightname);
+                dbg(lvl_debug, "tunnel nightlayout");
+                return;
+            } else if (*streetflags & AF_UNDERGROUND) {
+                // We are already in nightmode, no matter of the sun
+                return;
+            }
+
+        } else {
+            dbg(lvl_debug,"streetflags are NULL");
```
metalstrolch, line 3525: Is this really required?

navit/navit.c
```diff
@@ -3504,6 +3504,27 @@ void navit_layout_switch(struct navit *n) {
                 && valid_attr.u.num==attr_position_valid_invalid) {
             return; //No valid fix yet
         }
+
```
metalstrolch, line 3507: Suggest to move this check below line 3531. The the check for fix is done before that.

metalstrolch, line 3507: Suggest adding a attr to disable this, as not everybody may like that switch.

metalstrolch (changes requested):

metalstrolch:
Due to problems with configuration, it seems that circle ci build check is not run on your branch. In order to inforce the ci run, I pushed your branch directly to navit-gps repo as well. Please proceed as usual.
Btw: Please check coding style by applying 
astyle --indent=spaces=4 --style=attach -n --max-code-length=120 -xf -xh
to changed files. Otherwise your changes will not pass ci's sanity check step.

metalstrolch:
Once CI gives me a image, I'll check it on my sailfish device. If that is OK. I'll merge this.

navit/navit.c
```diff
@@ -3504,10 +3505,29 @@ void navit_layout_switch(struct navit *n) {
...
+        if (streetflags != NULL) {
+
+            dbg(lvl_debug, "streetflags: %i", *streetflags);
+
+            if (*streetflags & AF_UNDERGROUND) {
+                // We are in a tunnel and if we have a nightlayout -> switch to nightlayout
+                if (l->nightname) {
+                    navit_set_layout_by_name(n, l->nightname);
+                    dbg(lvl_debug, "tunnel nightlayout");
+                    return;
```
metalstrolch, line 3523: OK, ran it and found another problem: navit_layout_switch contains code that should result in navit not calculating the layout switch more often than once per minute. See check at line 3497. Since your addition does not correctly update n->prevTs, this leads to the funny situation that the switch when entering a tunnel is delayed, the switch when leaving the tunnel usually is not as n->prevTs is not updated and stays older then 60s.
IMHO the function should be refactored as a whole, but I don't want to request doing so here. But to get this in, please treat n->prevTs right.

For me I don't care if tunnel switching should not be delayed, or will be delayed as well, but at least it should behave symmetrically when entering or leaving a tunnel.

metalstrolch: And remove the "todo" mark, as this change aims to resolve it.

metalstrolch (changes requested):

mvglasow:
I am not really familiar with this section of the code, but as far as I know, we have (or had?) a kludge of swapping day/night layouts in the southern hemisphere. Unless that’s been considered here, driving into a tunnel in Australia at night will make Navit switch to day layout. You might want to set up a test case for this (demo vehicle, route through a tunnel in the southern hemisphere, run during night hours). If that misbehaves, the day/night layout switch logic would need to be fixed.

The other one: how do we know we are in a tunnel? Usually, our location source is GPS, which doesn’t work in tunnels. There’s been a tunnel extrapolation feature for a while, but that has never worked for me (my production platform being Android). However, I expect the only impact of that issue would be making this change pointless, without this change breaking anything.

OLFDB:
> There’s been a tunnel extrapolation feature for a while, but that has never worked for me (my production platform being Android).

I have investigated the code around tunnel extrapolation and the reason why it's not working is in vehicle_gpsd.c line 210:
`    // If data->fix.speed is NAN, then the drawing gets jumpy.`
`    if (! isnan(data->fix.speed) && priv->fix_type > 0) {`
`        callback_list_call_attr_0(priv->cbl, attr_position_coord_geo);`
`    }`

_tracking_update_ never gets notified of a lost GPS fix because of the check for NAN. If there is no fix, it is NAN.

I was finally able to create a NMEA file with a route through a tunnel and to simulate the lost fix. 
The missing update is the reason for tracking->tunnel always being 0 as well. 
The tunnel extrapolation feature is working. You have to set _tunnel_extrapolation="10"_ in the navit.xml file to have the extrapolation done with the last recorded speed. 

So far I didn't see "jumpy drawings", but that should be tested more detailed.

navit/navit.c
OLFDB: Added new attributes for navit tag: 

- tunnel_nightlayout              :enable nightlayout in tunnel -  default off
- layout_daynightauto           :enable auto_switch layout - default on
- sunrise_degrees                 :change the degrees after sunrise for calculating the auto switch trigger - default -5

navit/navit.c
metalstrolch: Seems OK now.

metalstrolch (approved):

Issue #762, opened 2019-03-14
Android build heavily pollutes source tree


mvglasow:
Currently, when we build Navit for Android using `scripts/build_android.sh` (which is now our common entry point for all builds—local, CI and F-Droid), build artifacts are created all over the source tree. Eariier versions of the script would even rename, move or modify some image files (which I have since fixed).

This makes cleaning up quite a nightmare. It involves deleting a list of directories and files, as well as resetting the Git repo. If you have uncommitted changes (a common occurrence when you are developing a new feature and are trying things out), there is a high risk of accidentally losing your changes.

Any way we can improve this? I would suggest aiming for the following:
* Do not mix source files (which includes images, makefiles etc.) and build artifacts in the same directory.
* Do not modify, rename, move or delete source files during build. If you must, create a copy and work on that (the copy would then be a build artifact)—currently OK but should be kept that way.
* When generating build artifacts inside the source tree, do so in as few locations as possible and only as many as strictly necessary. One or two would be ideal. (A location can be an entire directory hierarchy, as long as the root of the tree is safe to delete after build.)

mvglasow:
Can anyone help me with that, or at least point me to the places where I would need to make these settings? It seems to be only Android that is affected (note that I do the Linux build from Eclipse, though). I’m not really a toolchain expert…

mvglasow:
I just noticed `build_linux.sh` has the following lines:

```
BUILD_PATH="linux"

[ -d $BUILD_PATH ] || mkdir -p $BUILD_PATH
pushd $BUILD_PATH

# Build everything
    echo "Building..."
cmake ${cmake_opts} ../

```

…whereas `build_android.sh` runs cmake from the current dir. I’ll try adding the same there. Then I figure Gradle needs to be told to look for the native build artifacts in a different place. There are also some calls to `make` for icons, translations and XML which would need to be sorted out. No guarantee doing so won’t break anything—if someone has any advice for me (or, unlike me, knows what they’re doing), any pointers would be appreciated.

mvglasow:
I notice the Android build runs cmake twice: once directly, close to the beginning of the script, and then again from gradle (close to the end of the script). Both instances generate a lot of files all over the source tree (as well as the other steps in between, i.e. generating images and XML files).

* What do the two `cmake` calls do/build? Do we really need both, or can we merge them into one?
* Can we cd to a build directory before invoking each? (How do I do that from inside gradle? Will doing so break the build?)

mvglasow:
[...]

PR #1094, opened 2021-02-20
Fix:build:Reduce source tree pollution by Android build

mvglasow:
As described in #762, this now moves all CMake output from the Android build to a subdir named `android-builddir`.

<del>For now, generated XML, PNG and translation files still get copied into the source tree. However, each gets written to an otherwise unused subdir (no mixing of code and generated files within the same dir), making cleanup much easier than before.</del>

With this change, the Android build leaves five elements in the source tree:
* `android-builddir/` (the build dir)
* `cmake_plugin_settings.txt`
* <del>`navit/android/assets/` (XML config files, entire dir)</del>
* <del>`navit/android/res/drawable-nodpi/` (PNG bitmaps, entire dir)</del>
* <del>`navit/android/res/raw/`(translations, entire dir)</del>

@jkoan For now, this does not change anything about the use of `build_android.sh` or the double CMake. See if you can use this for your Android build efforts—if you have already solved the Android source tree pollution issue otherwise, feel free to close this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -88,6 +88,7 @@ android {
     }
     externalNativeBuild {
         cmake {
+            buildStagingDirectory "./android-builddir"
             path 'CMakeLists.txt'
         }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -82,8 +82,8 @@ android {
             java.srcDirs = ["navit/android/src"]
             resources.srcDirs = ["navit/android/src"]
             renderscript.srcDirs = ["navit/android/src"]
-            res.srcDirs = ["navit/android/res"]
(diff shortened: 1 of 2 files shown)


mvglasow:
With the additional commit, all build artifacts stay in android-builddir. I’ve tested the resulting APK and the multiple resource paths in `build.gradle` seem to work; both DPI-specific resources (such as the app icon on Android 4.4.4) and non-DPI resources (such as all cross-platform images, e.g. menu icons) are present.

That limits source tree additions to two items:
* `android-builddir/` (the build dir)
* `cmake_plugin_settings.txt` (file with local settings which we could add to `.gitignore`)

PR #1099, opened 2021-03-06
FIX: Retry connect after 10 seconds not performed

OLFDB:
Set priv->cbt before starting the timer.


navit/vehicle/gpsd/vehicle_gpsd.c
```diff
@@ -312,10 +313,11 @@ static void vehicle_gpsd_close(struct vehicle_priv *priv) {
     }
     if (priv->gps) {
         gps_close(priv->gps);
-#if GPSD_API_MAJOR_VERSION >= 5
-        g_free(priv->gps);
-#endif
-        priv->gps = NULL;
+//if we release the gps object a reconnect is no longer working.
```
jkoan, line 316: That's true, but `vehicle_gpsd_close` is only called when reopening the connection or on destroy. Is there something i am missing here?

OLFDB: If you use gpsfake e.g. to play tracks via gpsd, you need to restart navit when changing the track file e.g.
When using a network instance of gpsd, it might not be available all the time, so from my point of view it makes sense to have the reconnect option.

jkoan (approved):

PR #1106, opened 2021-04-03
Trucknavigation

### LINKED ISSUE #717: Revisit truck routing
Today I tried truck routing with Navit for the first time, and noticed two things:

* After failing to turn where I should have, Navit insisted that I do a U-turn. With 7.5 tons, that is a really bad idea. And yes, there’s an option for that somewhere, but from a UX point of view it would be nicer if this could somehow be implied in the vehicle profile. After all, the readiness to turn around (and feasibility of doing so) varies greatly between trucks and, say, pedestrians.
* For truck routing, height and width restrictions are crucial. Not sure if these are currently being considered at all. UX-wise we could prompt the user to enter these values after switching to the truck profile, but before we do that, we might want to store the last profile selection between sessions.

OLFDB:
The navit routing engine is capable of truck navigation, taking into account sizes and weights (including axle weights) already.
This PR will enable maptool to provide the data needed by the routing engine in the map binfiles.

The truck profile in navit.xml was enhanced by supplying values for turn_around_penalty and turn_around_penalty2 to prevent uturns during routing.


navit/maptool/osm.c
```diff
@@ -597,6 +602,7 @@ static char *attrmap= {
     "w	boundary=civil		border_civil\n"
     "w	boundary=national_park	border_national_park\n"
     "w	boundary=political	border_political\n"
+    "?  boundary=low_emission_zone  low_emission_zone\n"
```
jkoan, line 605: there should be a \t instead of two spaces

navit/maptool/osm.c
```diff
@@ -597,6 +602,7 @@ static char *attrmap= {
     "w	boundary=civil		border_civil\n"
     "w	boundary=national_park	border_national_park\n"
     "w	boundary=political	border_political\n"
+    "?	boundary=low_emission_zone	low_emission_zone\n"
```
metalstrolch, line 605: May be "w" in reality, as according to OSM wiki, boundary=low_emmision_zone is only allowed for areas and multi polygon relations which we map to polygons as well. This will result in quite big polygons on map probably. But it is very interesting information.

navit/maptool/osm.c
```diff
@@ -1103,6 +1119,41 @@ void osm_add_tag(char *k, char *v) {
             flags[0] |= AF_SPEED_LIMIT;
         level=5;
     }
+    if (!g_strcmp0(k, "maxheight")) {
```
metalstrolch, line 1122: Think we need to deal with imperial units as well. See https://wiki.openstreetmap.org/wiki/Key:maxheight
Same for the other dimensions.

navit/maptool/osm.c
```diff
@@ -53,6 +53,11 @@ static struct item_bin item;
 
 
 int maxspeed_attr_value;
+double maxheight_attr_value;
```
metalstrolch, line 56: Why "double"? Since the units are converted to e.g. "cm" anyway, there's no need to have floats here. Later the attrs are added as int to the items anyway.

navit/maptool/osm.c
```diff
@@ -1103,6 +1119,41 @@ void osm_add_tag(char *k, char *v) {
...
     }
+    if (!g_strcmp0(k, "maxheight")) {
+        maxheight_attr_value = (int)(atof(v)*100);
+
+        if (maxheight_attr_value)
+            flags[0] |= AF_SIZE_OR_WEIGHT_LIMIT;
+        level = 5;
+    }
+    if (!g_strcmp0(k, "maxlength")) {
+        maxlength_attr_value = (int)(atof(v)*100);
```
metalstrolch, line 1130: The "*100" is the reason there should be no float required.

metalstrolch (changes requested):
Good work. I'm looking forward for the low emmission zones, as those can be very interesting on map.

navit/maptool/osm.c
OLFDB: My first guess was to use the osm values directly, but when I saw the unit used for the vehicle attributes are in cm I didn't change the type.

navit/maptool/osm.c
OLFDB: I have not finished working on the low emission zones and this is another PR which should be removed from truck navigation. They are not showing up in the map. Maybe you have an idea why. I use

<itemgra item_types="low_emission_zone" order="0-">
    <polyline color="#00FF00" width="10" dash="5,5"/>
</itemgra>

in layout name="polygons". Changing to "w" type doesn't solve it.

metalstrolch: I've committed the fix for low_emission_zone. Now you can have it in itemgra. However the combination of width != 1 and dash results in "funny" behaviour as graphics engine cannot dash wide lines due to the way they are drawn. I'd suggest transparent solid line.
To use it for routing however is a different story.

OLFDB: Thank you, but it's still not showing up on the map.

metalstrolch: OK, got it. low emission zones are mostly tagged on boundary relations, not on multipolygon ones. OSM wiki is not clear about this. Wil push fix. Caution: fix renames low_emission_zone to more correct poly_low_emission_zone. Adapt your layout accordingly

OLFDB: Perfect. It works fine now. Thank you.

OLFDB:
After converting some maps using maptool: how to handle e.g. below_default for the height? If we convert this height to 0 the way gets not flagged, but it will definitely lead to issues for trucks. We could use a default value of 250 so the way is excluded in truck routing.

OLFDB:
Added conditional speed limits. WIP.
condspeedlimits

OLFDB:
Added vertical meters to the heightline widget in internal gui as that might be of interest planning a route with a truck, or bike, or pedestrian.
heightlines

OLFDB:
Improvements for conditional speed limits. Backward limits are supported as well. Improved gui, now showing icons for danger, hazmat and wet condition. Hazmat conditions shown only if vehicle has hazmat set in it's profile. 

```
<osd enabled="yes" background_color="#FFFFFF00" color="#FFFFFFFF"
    type="cond_speed_warner"
   w="130" h="130" x="78%" y="-100%" font_size="250"
   speed_exceed_limit_offset="5" speed_exceed_limit_percent="10"              
   announce_on="1" label="images:maxspeed_48_48.png:maxspeed_48_48.png:empty.png:stvo_1040_30.png:stvo_1052_30.png:stvo_1053_35.png:stvo_1006_31:" name="cond_speed_warner" />
   <!--										 CONDITION        HAZARD	   WET		    DANGER         -->
```

unfallgefahr
hazard
wet

Routing recognizing low emission zones

Thanks to mvglasow for his proposals for the algorithms to find intersections of a line with a polygon. I'm still trying to get routing to recognize the low emission zones. route value for segments in lez is INT_MAX which should result in blocked route if route ends in a lez. I have tested with a small map of Berlin only, so there is only one lez in the map.
I'm reading the lez polygon in route_graph_add_street and searching intersection with current segment in route_graph_add_segment. For detection I took the line_intersection function from gui_internal.c. It detects intersections, but a route is still calculated. 

Any hint would be appreciated.

OLFDB:
Finally the routing taking into account the low emissions zones is working.
I think the PR is finished now.

https://user-images.githubusercontent.com/5459286/119258878-e46c7380-bbcb-11eb-9ffb-0385843c21c0.mov

navit/osd/core/osd_core.c
```diff
@@ -2239,7 +2239,10 @@ static void osd_nav_toggle_announcer_draw(struct osd_priv_common *opc, struct na
         gr_image = graphics_image_new_scaled(opc->osd_item.gr, path, opc->osd_item.w, opc->osd_item.h);
         if (!gr_image) {
             g_free(path);
-            path = graphics_icon_path("unknown.png");
+            if (this->active)
+                path = graphics_icon_path("sound_on.png");
+            else
+                path = graphics_icon_path("sound_off.png");
```
jkoan, line 2245: Strictly speaking this dosn't belong into this PR

navit/maptool/osm.c
```diff
@@ -139,800 +162,840 @@ struct country_table {
...
+                    getcurrentid(), result, value);
+//    else
+//        fprintf(stderr,"Node: %lli unit weight conversion converted value from metric tons: %i"ANSI_DEF"\n", getcurrentid(), result);
+
+    return result;
+}
+
+static char* cleanCondition(char* condition) {
+
+    while(*condition==32) { //remove leading whitespace TODO: do it in maptool
```
jkoan, line 780: We are already within maptool

navit/maptool/osm.c
```diff
@@ -139,800 +162,840 @@ struct country_table {
...
+        condition++;
+    }
+    while(*(condition+strlen(condition)-1)==32) { //remove trailing whitespace TODO: do it in maptool
```
jkoan, line 783: This as well

navit/osd/core/osd_core.c
```diff
@@ -2857,6 +2862,455 @@ static struct osd_priv *osd_speed_warner_new(struct navit *nav, struct osd_metho
...
 
+static char* cleanCondition(char* condition) {
+    char *ptrcond;
+    while (ptrcond = strstr(condition, ")")) {
+        *ptrcond = 32;
+    }
+    while (ptrcond = strstr(condition, "(")) {
+        *ptrcond = 32;
+    }
+    while(*condition==32) { //remove leading whitespace TODO: do it in maptool
```
jkoan, line 2873: Is this still current? see comment in maptool

navit/osd/core/osd_core.c
```diff
@@ -2857,6 +2862,455 @@ static struct osd_priv *osd_speed_warner_new(struct navit *nav, struct osd_metho
...
+        condition++;
+    }
+    while(*(condition+strlen(condition)-1)==32) { //remove trailing whitespace TODO: do it in maptool
```
jkoan, line 2876: as well

navit/route_protected.h
```diff
@@ -130,32 +155,44 @@ struct route_graph_segment {
...
  * @brief A complete route graph
  *
  * The route graph holds all routable segments along with the connections between them and the cost of
  * each segment.
  */
 struct route_graph {
-	int busy;                                   /**< Route calculation is in progress: the graph is being built,
-	                                             *   flooded or the path is being built (a more detailed status can be
-	                                             *   obtained from the route’s status attribute) */
+	int busy;                                   /**< The graph is being built */
```
jkoan, line 180: Why removing a longer more explaining comment? Is is outdated?

navit/route_protected.h
```diff
@@ -130,32 +155,44 @@ struct route_graph_segment {
...
 	struct mapset_handle *h;                    /**< Handle to the mapset */
 	struct map *m;                              /**< Pointer to the currently active map */
 	struct map_rect *mr;                        /**< Pointer to the currently active map rectangle */
 	struct vehicleprofile *vehicleprofile;      /**< The vehicle profile */
 	struct callback *idle_cb;                   /**< Idle callback to process the graph */
 	struct callback *done_cb;                   /**< Callback when graph is done */
+	struct callback *donelez_cb;                /**< Callback when graph is done */
 	struct event_idle *idle_ev;                 /**< The pointer to the idle event */
 	struct route_graph_segment *route_segments; /**< Pointer to the first route_graph_segment in the linked list of all segments */
-	struct route_graph_segment *avoid_seg;      /**< Segment to which a turnaround penalty (if active) applies */
```
jkoan, line 155: Why removing this comment?

navit/vehicleprofile.h
```diff
@@ -29,6 +31,39 @@ enum maxspeed_handling {
...
+            "Euro 3",
+            "Euro 4",
+            "Euro 5",
+            "Euro 6",
+            "Red Label",
+            "Yellow Label",
+            "Green Label",
+            "Euro 6d temp"
+};
+
```
jkoan, line 66: This seems to be really specific to Germany, what about other country's?

scripts/update_download_center.sh
```diff
@@ -35,7 +35,7 @@ echo "Setting up all Variables"
 UUID=${RANDOM}-${RANDOM}-${RANDOM}-${RANDOM}
 TMP_DIR=$(mktemp -d)
 CIRCLECI_API_BASE="https://circleci.com/api/v1.1/"
-NAVIT_DOWNLOAD_CENTER_REPO="git@github.com:navit-gps/download-center"
+NAVIT_DOWNLOAD_CENTER_REPO="git@github.com:OLFDB/download-center"
```
jkoan, line 38: please revert this

jkoan (changes requested):
In addition to the other comments i noticed that that your branch needs to be rebased, are you able to do so?

navit/attr.h
```diff
@@ -65,7 +65,7 @@ enum attr_format {
 #define AF_ROUNDABOUT_VALID	(1<<4)
 #define AF_ONEWAY_EXCEPTION	(1<<5)
 #define AF_SPEED_LIMIT		(1<<6)
-#define AF_RESERVED1		(1<<7)
```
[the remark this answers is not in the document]
OLFDB: A grep for AF_RESERVED1 didn't show any usage.

navit/icons/stvo_1053_35.svg
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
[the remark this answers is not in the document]
OLFDB: Removed text from image.

navit/vehicleprofile.h
OLFDB: The Euro x classes are valid in EU, not only in Germany. For the labels you are right. 
The purpose of the emission class was to use it for routing with the low emission zones. As the access tag for the lezs is missing often, it is not usable at the moment. That's why I added the "Enter low emission zone" value that is used for routing now.

scripts/update_download_center.sh
OLFDB: Same as above. Would like to keep it in my fork only, so my CI build is working. Is that somehow possible?

navit/osd/core/osd_core.c
OLFDB: You are right, do I need to remove it?

navit/maptool/osm.c
OLFDB: Removed the comment.

navit/osd/core/osd_core.c
OLFDB: Removed the comment.

scripts/update_download_center.sh
jkoan: Need to revert d670c8ca19fa8976e389f6595420a29a810e9ed4 for this

OLFDB:
> Good work. I'm looking forward for the low emmission zones, as those can be very interesting on map.

@metalstrolch: When testing with larger maps having more than one low emission zone, the detection of the lezs is not working properly. I think I need somehow to have a reference to the tiles containing the lez to be able to select the correct polygon to test against during route calculation. Could you help me with that?

christf:
In an attempt to test this with large maps and multiple LEZ, I rebased this today. Unfortunately the result does build for all platforms but windows and to add insult to injury, I am not sure what the error message means in this context. Sharing it anyways, maybe another set of eyeballs and brain can help :-)

https://github.com/christf/navit/tree/trucknavigation Here is the code (again, retaining the original author of the trucknavigation code). This is not yet raised as separate PR because of that compile error https://github.com/christf/navit/actions/runs/21120085136/job/60731485752

PR #1107, opened 2021-04-03
Fix:ci:Fix fdroidserver download path

### LINKED ISSUE #1105: F-Droid CI build failing blocking merges
Currently the f-droid CI build fails constantly, which causes merges to be blocked. Unfortunately I don't have a clue how f-droid works so I cannot fix that. Who can help?

Thanks!

Blocks at least #1089

mvglasow:
Fixes #1105


mvglasow:
btw, I just saw F-Droid has done the same thing, see https://gitlab.com/fdroid/fdroiddata/-/commit/c4c446b98dfa8806d48acc9d2323322017dc3ae0

PR #1109, opened 2021-04-11
Fix:vehicle:geoclue Fix cmake usage

metalstrolch:
The cmake statement for geoclue vehicle wrongly passes the header paths to the add library statement. It should pass the library paths instead.


jkoan (approved):
Thx for finding this one

Issue #1117, opened 2021-05-01
File system restrictions on Android 11


mvglasow:
After upgrading to Android 11, Navit is no longer seeing any files (navit.xml or map files) in `/sdcard/Android/data/org.navitproject.navit/files`. This has still worked in Android 10.

The logcat shows this message:

    W itproject.navit: type=1400 audit(0.0:479): avc: denied { read } for name="sdcard" dev="tmpfs" ino=6474 scontext=u:r:untrusted_app_29:s0:c244,c256,c512,c768 tcontext=u:object_r:mnt_sdcard_file:s0 tclass=lnk_file permissive=0 app=org.navitproject.navit

and a few debug messages from Navit telling us that a bunch of file access operations have failed.

Also, for file and media permission, only media access can be selected, full file access is not even available.

According to information on the net, Android 11 reversed the previous decision of locking apps out of the file system, see https://stackoverflow.com/questions/60360368/android-11-r-file-path-access and https://stackoverflow.com/questions/62782648/android-11-scoped-storage-permissions/66366102#66366102.

As far as I can see it, full filesystem access requires `MANAGE_EXTERNAL_STORAGE` permission (newly introduced in Android 11). The previous `WRITE_EXTERNAL_STORAGE` becomes ineffective (i.e. does nothing) on Android 11, though iirc this affects only apps targeting SDK 30+. `MANAGE_EXTERNAL_STORAGE` is subject to a PlayStore policy: Google will accept this permission only for certain use cases, and Navit likely does not fall into this category.

Another new restriction is that apps can no longer access each other’s private shared storage areas, or in fact anything underneath certain paths (among them `/sdcard/Android/data`) that doesn’t belong to them. Not sure if that is what is preventing us from accessing `/sdcard/Android/data/org.navitproject.navit/files`; after all, this path belongs to our app.

I tried yesterday to target SDK 30, request `MANAGE_EXTERNAL_STORAGE` and grant full filesystem permissions to Navit, but still get the exact same behavior and the same error message.

I have asked on [Stack Overflow](https://stackoverflow.com/questions/67340898/android-11-avc-denied-read-for-name-sdcard), with no answers so far.

@jkoan tells me he got similar reports by mail but also hasn’t figured out so far why this is not working. Any input is appreciated.

mvglasow:
With jkoan’s help I got the emulator up and running and set up Navit:

* Installed Navit from a locally built APK via drag and drop
[...]

PR #1125, opened 2021-06-20
Make file picker compatible with Android 11

mvglasow:
As mentioned in #1117:

* Default location for file picker is Navit’s private shared storage dir (`/sdcard/Android/data/org.navitproject.navit/files`), which is not affected by restrictions introduced with Android 11
* When the selected path contains Navit’s package ID, it is applied as-is (rather than appending `/navit`) – this is the same behavior as for paths including `/navit`, the assumption being that we are already in a Navit-specific part of the filesystem


jkoan (approved):
Thx, Looks really good

mvglasow:
Thanks for merging! Seems something failed on CircleCI though, as the wince build failed, thus the changes didn’t make it into trunk yet. Any ideas?

jkoan:
I have seen this as well, but so far no clue

jkoan:
Ah got it, SSL error because the wince image has really old dependency's:

`Unable to establish SSL connection.`

PR #1140, opened 2021-08-31
rework:graphics:win32:Rework win32 drawing

### LINKED ISSUE #575: Issues with navit/support/win32/ConvertUTF.c

jkoan:
This PR resolves #575

In order to remove the Unicode Sourcecode the used function was replaced by the Windows Function `MultiByteToWideChar`


aerostitch:
Thanks!

PR #1142, opened 2021-09-01
Fix:core: cancel drawing on resize request

metalstrolch:
This commit causes navit to cancel async drawing in case of resize
request. Drawing is started again after that. this should effectively
cause a redraw on subsequent calls to resize. Some graphics like qt5
call resize for both dimension changes if the widget gets resized.

### DIFF
--- a/navit/graphics/qt5/QNavitQuick.cpp
+++ b/navit/graphics/qt5/QNavitQuick.cpp
@@ -181,12 +181,7 @@ void QNavitQuick::geometryChanged(const QRectF& newGeometry, const QRectF& oldGe
     if (graphics_priv->pixmap == NULL) {
         graphics_priv->pixmap = new QPixmap(width(), height());
     }
-    painter = new QPainter(graphics_priv->pixmap);
-    if (painter != NULL) {
-        QBrush brush;
-        painter->fillRect(0, 0, width(), height(), brush);
-        delete painter;
-    }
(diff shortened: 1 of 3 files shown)


metalstrolch:
Added a fix to remove garbage displaying on qt5 after resize until navit finally finishes the redraw. This has nothing to do with the initial fix, except removing the garbage.

metalstrolch:
Fixes at least the oddities on screen rotation on sailfish os.

jkoan:
I recently developed some things related to graphics as well and I think this could fix the biggest problem I encountered. Will test this soon. Thanks for the great work

jkoan:
I couldn't actually reproduce any wrong behavior. Could you help me with reproducing the error on graphics other than qt5, or at least how to trigger it generally?

metalstrolch:
Simplest way on desktop machines is to resize the window using by dragging the window border while showing a big city where rendering the screen takes some time. Try Munich for example. You see the window then sometimes not being completely filled and vehicle cursor being at the wrong position compared to what the window shows. You need to generate a resize request as long as the screen renders in background.

metalstrolch:
Any objections? This one helps for Sailfish screen rotation at least

jkoan (approved):
I wasn't able to really reproduce it, but i trust you that it firxes what you tested.

PR #1143, opened 2021-09-01
Fix:maptool:osm fix multipolygon filter.

metalstrolch:
Change done at #1129 causes maptool to filter out all multipolygons
if -n is given due to not correctly initializing item_bin buffer
prior check. Sorry my fault.

### DIFF
--- a/navit/maptool/osm.c
+++ b/navit/maptool/osm.c
@@ -1684,6 +1684,7 @@ static inline void osm_end_relation_multipolygon (struct maptool_osm * osm) {
                 /*Don't write out multipolygons that will result in unknown types if -n is given.
                  *So we don't process useless multipolygons. May save a lot of time.
                  */
+                tmp_item_bin->type = types[a];
                 if(filter_unknown(tmp_item_bin))
                     continue;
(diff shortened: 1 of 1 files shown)


metalstrolch:
Seems nobody is processing maps these days as nobody cried out.

mvglasow:
I just tried to process a map (with this PR merged) and got the error

```
PROGRESS9: Processed 21793 nodes (21708 out) 0 ways 74044 relations 0 tiles 2:14 27 MB
PROGRESS9: Processed 21793 nodes (21708 out) 0 ways 0 relations 0 tiles 2:14 27 MB
process_multipolygons:process (thread 0)
process_multipolygons:finish (thread 0)
maptool: /home/michael/src/navit/navit/maptool/itembin.c:93: item_bin_copy_attr: Assertion `attr == attr_osm_wayid' failed.
```
both with OSM and Protobuf input (same map in both cases). Is this related or is it a different issue?

mvglasow:
Just checked out ea5a86bea (the last before #1129), same error. Apparently it’s not related.

Issue #1167, opened 2021-12-04
buffer overflow in transform()


bkoppelmann:
Hi,
I was running navit on a raspberry pi with pi os and compiled the latest git revision. When switching to 3D mode, navit would crash on some occasions. I isolated the problem using valgrind and found it on the following codepath:

- When `displayitem_transform_holes()` is called, we allocate a `struct point` buffer ([graphics.c:2701](https://github.com/navit-gps/navit/blob/ed26d58ddcf7d48352cdd9f673f3e9096df0c905/navit/graphics.c#L2701)) of size `count`.
- Then we call `transform()` ([transform.c:548](https://github.com/navit-gps/navit/blob/ed26d58ddcf7d48352cdd9f673f3e9096df0c905/navit/transform.c#L548)) to fill that buffer called `result`. In this function we fill the buffer in a for loop that runs `count` times. The buffer is indexed using `result_idx` which is incremented every loop iteration. However, if we are in 3d mode (indicated by `t->ddd`), we call `transform_z_clip_if_necessary()`. This can lead to the repetition of the current loop iteration by decreasing the loop variable `i` by 1. ([transform.c:573](https://github.com/navit-gps/navit/blob/ed26d58ddcf7d48352cdd9f673f3e9096df0c905/navit/transform.c#L573)).  Even though we decreased `i` we still increment `result_idx` by 1. So from the point of view of `result_idx` we are running the loop `count+1` times. Thus, we write one element past the allocated buffer. 

My fix is this:
```
diff --git a/navit/transform.c b/navit/transform.c
index 771037f8c..685d314a3 100644
--- a/navit/transform.c
+++ b/navit/transform.c
@@ -552,9 +552,10 @@ int transform(struct transformation *t, enum projection required_projection, str
     struct point screen_point;
     int zlimit=t->znear;
     struct z_clip_result clip_result, clip_result_old= {{0,0}, -1, 0, 0};
-    int i,result_idx = 0,result_idx_last=0;
+    int i,result_idx = 0,result_idx_last=0, result_idx_inc;
     dbg(lvl_debug,"count=%d", count);
     for (i=0; i < count; i++) {
+        result_idx_inc = 1;
         dbg(lvl_debug, "input coord %d: (%d, %d)", i, input[i].x, input[i].y);
 #if 0 /* doesn't work as wanted */
         if (i && input[i].x == input[0].x && input[i].y == input[0].y && result_idx && !width_result) {
[...]

PR #1170, opened 2021-12-06
Fix buffer overflow for ticket #1167

bkoppelmann:
Hi,

I came up with a proper fix for this buffer overflow. We calculate if we can fit a repetition of one point into our result buffer. If we cannot we bail from the function.

Cheers,
Bastian


navit/transform.c
```diff
@@ -570,7 +577,15 @@ int transform(struct transformation *t, enum projection required_projection, str
             clip_result=transform_z_clip_if_necessary(rotated_coord, zlimit, clip_result_old);
             clip_result_old=clip_result;
             if(clip_result.process_coord_again) {
-                i--;
+                loop_repeats++;
+                /* if we repeat an interation, we have to make sure that there is enough space in the result buffer to
+                   not overflow. */
+                if (i + loop_repeats < max_results) {
```
metalstrolch, line 583: You're checking against the wrong limit here. "i" isn't your concern, as "i" refers to your input buffer only.
If you think that both your input and output are 3 bytes long. Assume no points get skipped. the repeating the first point will be OK passing your check. But it will crash at the last point which is not repeated at all but still beyond array limits.

You should check "result_idx" against "max_result" instead, i think.
The you don't even need to count the repetitions in loop_repeats,

Line 610 in this listing including your changes is where you may want to look at. unfortunately this review tool doesn't allow me to annotate unchanged code. What I find a pity.

navit/transform.c
```diff
@@ -545,14 +545,21 @@ static struct z_clip_result transform_z_clip_if_necessary(struct coord_3d coord,
     return clip_result;
 }
 
-int transform(struct transformation *t, enum projection required_projection, struct coord *input,
-              struct point *result, int count, int mindist, int width, int *width_result) {
+int transform_point(struct transformation *t, enum projection required_projection, struct coord *input,
```
metalstrolch, line 548: I like that refactoring. That's removing many silly parameters from the most uses.

metalstrolch (changes requested):
Good find indeed. But your fix isn't safe yet. Check comments.

navit/graphics.c
```diff
@@ -2697,9 +2697,11 @@ static inline void displayitem_transform_holes(struct transformation *trans, enu
         out->ccount = g_malloc0(sizeof(*(out->ccount)) * in->count);
         out->coords = g_malloc0(sizeof(*(out->coords)) * in->count);
         for(a = 0; a < in->count; a ++) {
+            int buf_size=sizeof(*(out->coords[a])) * in->ccount[a];
             in->ccount[a]=limit_count(in->coords[a], in->ccount[a]);
-            out->coords[a]=g_malloc0(sizeof(*(out->coords[a])) * in->ccount[a]);
-            out->ccount[a]=transform(trans, pro, in->coords[a], (struct point *)(out->coords[a]), in->ccount[a], mindist, 0, NULL);
+            out->coords[a]=g_malloc0(buf_size);
+            out->ccount[a]=transform_point_buf(trans, pro, in->coords[a], (struct point *)(out->coords[a]), buf_size, in->ccount[a],
```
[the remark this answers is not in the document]
bkoppelmann: That's a good idea. I can implement that with little work :)

navit/transform.c
```diff
@@ -570,7 +577,15 @@ int transform(struct transformation *t, enum projection required_projection, str
...
             clip_result_old=clip_result;
             if(clip_result.process_coord_again) {
-                i--;
+                loop_repeats++;
+                /* if we repeat an interation, we have to make sure that there is enough space in the result buffer to
+                   not overflow. */
+                if (i + loop_repeats < max_results) {
+                    i--;
+                } else {
+                    dbg(lvl_warning, "Not enough space in buf for transform_point_buf, skipping further points");
```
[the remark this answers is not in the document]
bkoppelmann: Sure.

navit/transform.c
bkoppelmann: Yeah, it was a bit overcomplicated. Just using `result_idx` is clearer to understand.

bkoppelmann:
Hi metalstrolch,
I pushed a new version, which addresses you comments. Thanks for the review!
Cheers,
Bastian

navit/graphics.c
```diff
@@ -2691,15 +2691,26 @@ static inline void displayitem_transform_holes(struct transformation *trans, enu
...
-            out->coords[a]=g_malloc0(sizeof(*(out->coords[a])) * in->ccount[a]);
-            out->ccount[a]=transform(trans, pro, in->coords[a], (struct point *)(out->coords[a]), in->ccount[a], mindist, 0, NULL);
+            out->coords[a]=g_malloc0(buf_size);
+            transform_res=transform_point_buf(trans, pro, in->coords[a], (struct point *)(out->coords[a]), buf_size, in->ccount[a],
+                                              mindist, 0, NULL);
+            /* if we did not have enough buf space for transfrom_point_buf, we try again with double the buffer size,
+               until we succeed. */
+            while (transform_res == TRANSFORM_ERR_BUF_SPACE) {
+                buf_size *= 2;
+                out->coords[a] = g_realloc(out->coords[a], buf_size);
```
metalstrolch, line 2709: Here you should at least check if you still got memory, and break if not. When allocating in a loop and something goes nuts, then we run until all memory is allocated by us. We can protect us against crashing in that case by checking the malloc result. There are still some devices around that are very limited in ram.

metalstrolch (changes requested):
One comment left. Now it's looking quite good.

Another thing:
It is not a "blocker" for aproval so I do not formally request it. But you should think about writing a function comment header for your changed functions. Yes, Navit is not nice here, and most devs including me frequently forget this. So maybe you want to do better :)

navit/graphics.c
bkoppelmann: g_realloc() terminates the program on failure, it's not an uncontrolled crash. Do you want navit to continue and just abort this operation? Maybe issue a warning/error message?

metalstrolch: Well yes. But I forgot the strange habit of glib. The check mentioned by me is pointless as g_realloc will terminate anyway.

metalstrolch (approved):

PR #1176, opened 2022-01-01
fix:speech/speech_dispacher:use spd_say instead of spd_sayf

trldp:
The use of spd_sayf with a string literal causes the following error when building on Ubuntu Jammy.
`/<<PKGBUILDDIR>>/navit/speech/speech_dispatcher/speech_speech_dispatcher.c: In function ‘speechd_say’:
/<<PKGBUILDDIR>>/navit/speech/speech_dispatcher/speech_speech_dispatcher.c:41:5: error: format not a string literal and no format arguments [-Werror=format-security]`

The solution in this commit is to use spd_say instead of spd_sayf.


bkoppelmann:
LGTM. I was checking the API in [speechd](https://github.com/brailcom/speechd/blob/b5b45e99e588a28ab8e38c043533eeae5e5b38ce/src/api/c/libspeechd.h#L116) and `spd_sayf` expects a format string similar to printf which we were not giving. `spd_say` does not require that. So this looks reasonable.

Cheers,
Bastian

PR #1184, opened 2022-05-24
fix:maptool: Don't drop unrecognized relations to borders temp file

metalstrolch:
Cleans up place to area processing

### DIFF
--- a/navit/maptool/osm.c
+++ b/navit/maptool/osm.c
@@ -1715,7 +1715,7 @@ static inline void osm_end_relation_multipolygon (struct maptool_osm * osm) {
             //tmp_item_bin->type=type_none;
         }
         if ((!g_strcmp0(relation_type, "multipolygon") || !g_strcmp0(relation_type, "boundary"))
-                && (boundary || type!=type_none)) {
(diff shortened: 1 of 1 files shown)


jkoan:
Can you explain whats the current problem is about filtering out type_none?

metalstrolch:
We're currently not filtering out type_none. Instead, we drop everything NOT type_none into the boundaries file. Regardless if it is actually a boundary according to our needs. This has the effect that boundary processing takes far longer than required.

jkoan:
> We're currently not filtering out type_none. Instead, we drop everything NOT type_none into the boundaries file. Regardless if it is actually a boundary according to our needs. This has the effect that boundary processing takes far longer than required.

Got it! Now i also see the logic error.

jkoan:
> @jkoan , do you know why ci fails on "trunk" for linux_build during updating of the translation files to/from launchpad? Or do you think I should open a ticket about this?

i know more or less what its about. You don't need to open up a ticket, i will take care of it

metalstrolch:
Thanks.

PR #1188, opened 2022-06-12
Improvements for vehicle_demo

OLFDB:
If AF_UNDERGROUND flag is set in tracking attr_position_fix_type is 0. This enables for testing nightlayout for tunnels.
Configurable height value.
Deliver random values for speed, height, num_sats_used (vary around configured values).

### DIFF
--- a/navit/vehicle/demo/vehicle_demo.c
+++ b/navit/vehicle/demo/vehicle_demo.c
@@ -106,8 +106,7 @@ static int vehicle_demo_position_attr_get(struct vehicle_priv *priv,
         if(flags = tracking_get_current_flags(navit_get_tracking(priv->navit))) {
             if(*flags & AF_UNDERGROUND)
                 attr->u.num = 0;
-        }
-        else
+        } else
             attr->u.num = 2;
         break;
     case attr_position_sats_used:
--- a/navit/vehicle/demo/vehicle_demo.c
+++ b/navit/vehicle/demo/vehicle_demo.c
@@ -86,7 +86,6 @@ static int vehicle_demo_position_attr_get(struct vehicle_priv *priv,
     int hr,min,sec,year,mon,day;
     double lat,lng;
     int *flags;
-    struct attr speed_attr;
     switch (type) {
     case attr_position_speed:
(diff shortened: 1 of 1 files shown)


navit/vehicle/demo/vehicle_demo.c
```diff
@@ -244,7 +257,8 @@ static void vehicle_demo_timer(struct vehicle_priv *priv) {
                     ci.y = pos.y + dy * len / slen;
                     priv->direction =
                         transform_get_angle_delta(&pos, &c, 0);
-                    priv->speed=priv->config_speed;
+                    priv->speed=priv->config_speed + ((rand() % 5 + 1) * (rand() %2 == 0?-1:1));  // a little random + or - 1 to 5 km/h
+                    priv->height=priv->height + ((rand() % 50 + 1) * (rand() %2 == 0?-1:1));       // a little random + or - 1 to 50 m
```
jkoan, line 261: Is it a good idea to have a vehicle which constantly changes its height? In my mind this could introduce problems where navit could recalculate the route because the vehicle is suddenly bigger than the bridge where the vehicle is already under. I don't have anything against the random nature of the height in itself, but chat about keeping this to the init?

navit/vehicle/demo/vehicle_demo.c
```diff
@@ -282,6 +296,7 @@ static struct vehicle_priv *vehicle_demo_new(struct vehicle_methods
     ret->cbl = cbl;
     ret->interval=1000;
     ret->config_speed=40;
+    ret->height=-10;
```
jkoan, line 299: is this a sane default? i would guess that about 2-3m is more appropriate

jkoan (changes requested):

navit/vehicle/demo/vehicle_demo.c
OLFDB: This is the locations height above sealevel, not the vehicle height. The vehicles height is handled in the vehicle profile.

jkoan: Ah okay, than its valid, sure

PR #1202, opened 2022-10-15
Fix:vehicle/android:Never use fused provider for precise location

mvglasow:
Android’s fused location provider aggregates locations from multiple providers, typically network and gps.

The AOSP version seems to indiscriminately pipe through any location updates it gets from the backend providers, with no verification or filtering.

When your references are a precise GPS signal and a network location based on a cell tower, with several hundred meters of error, this may cause your position to jump wildly between two places and navigation will become unusable. Such behavior would be OK for the ”fast” provider (which establishes your initial position and gets disabled as soon as the first location from the precise provider is received), but not for the precise provider (which would normally be used for navigation).

Unfortunately, Android seems to prefer the fused location provider over any others when selecting the best location provider based on criteria, and there don’t seem to be any criteria that would effectively deselect the fused provider. That makes Navit basically useless on my new phone.

This PR modifies the way Navit selects its location providers:

* The fused provider (also the passive one, which acts in a similar manner) is never used as the precise provider.
* If Android suggests a provider other than these, based on criteria, we will stick with Android’s suggestion.
* Otherwise, we examine the list of all eligible providers (based on criteria, but eliminating fused and passive):
  * If there is only one provider, we use that.
  * If there are multiple providers and one of them is GPS, we use GPS.
  * Otherwise, we pick a provider at random (should only happen on some very exotic devices – we’re talking about a device which has no GPS, but two high-precision location sources to choose from).
* The fast provider must differ from the precise provider (selecting the second-best fast provider is better than selecting the precise provider twice)
* If Android suggests one that matches our requirements, we use that. (This would likely be the fused provider, if available, or the network provider otherwise.)
* Otherwise, we examine the list of eligible providers (based on criteria, but eliminating the precise provider):
  * If there is only one provider, we use that.
  * Otherwise, we pick a provider at random.

Tests indicate that this will fix the problem. On my new device, Navit now again uses GPS for precise location and fused for fast location.

### DIFF
--- a/navit/android/src/org/navitproject/navit/NavitVehicle.java
+++ b/navit/android/src/org/navitproject/navit/NavitVehicle.java
@@ -34,6 +34,7 @@ import android.os.Bundle;
 import android.support.v4.content.ContextCompat;
 import android.util.Log;
 
+import java.util.HashSet;
 import java.util.List;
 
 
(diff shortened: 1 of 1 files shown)


mvglasow:
`build_fdroid` failed due to what seems to be an issue with the CI image, which is also affecting other people’s builds. See #1203

Issue #1185, opened 2022-05-25
Auto layout switch not taking local time into account


OLFDB:
In navit.c navit_layout_switch the GNSS time (UTC) is taken to calculate the sunrise/set values.

```
3540 if (vehicle_get_attr(n->vehicle->vehicle, attr_position_time_iso8601,&iso8601_attr,NULL)==1) {
            currTs=iso8601_to_secs(iso8601_attr.u.str);

```
Later currTs is used to determine if time is after sunrise, or after sunset and to switch to day/night layout.

```
3607 if (HOURS(trise)*60+MINUTES(trise)<(currTs%86400)/60) {
            after_sunrise = TRUE;
        }

        if (((HOURS(tset)*60+MINUTES(tset)<(**currTs**%86400)/60)) ||
                ((HOURS(trise_actual)*60+MINUTES(trise_actual)>(**currTs**%86400)/60))) {
            after_sunset = TRUE;
        }
        if (after_sunrise && !after_sunset && l->dayname) {
            navit_set_layout_by_name(n,l->dayname);
            dbg(lvl_debug,"layout set to day");
        } else if (after_sunset && l->nightname) {
            navit_set_layout_by_name(n,l->nightname);
            dbg(lvl_debug,"layout set to night");
        }

```
This is only working correctly when the system time is set to UTC (and DST is off) and we are in the UTC time zone.

It could be solved by using Zonedetect for getting the current timezone based on the GNSS coordinates.
(Zonedetect would also be able to provide the country based on coordinates to change default country for searching destinations to have to current country by default).

At least system timezone and DST should be taken into account for currTs:

```
// add timezone offset to GNSS UTC time to calculate sunrise/set correctly
            time_t z = time(NULL);
            struct tm * pdt = gmtime(&z);
            mktime(pdt);

            // use pdt->tm_isdst to have standard local time when system uses DST
            currTs=iso8601_to_secs(iso8601_attr.u.str) + (pdt->tm_gmtoff-(pdt->tm_isdst==1?3600:0));
```

OLFDB:
The issue was caused by the vehicle/iphone. It sends localtime instead of UTC in the location event.
So what I wrote above is crap. Nevertheless there is an issue when negative rise times are returned by the sunriset function.

This can be addressed by  

       ```
 int hr = HOURS(trise);
        int mr = MINUTES(trise);
        int hs = HOURS(tset);
        int ms = MINUTES(tset);
        int minofrise;
[...]

PR #1215, opened 2023-01-05
Fix:#1185

OLFDB:
sunriset was returning negative risetimes. This is the fix.


OLFDB:
I did some further testing. Please see rise.txt and riseold.txt in attached archive for results of layout switching. Previously (risetold.txt) using the current implementation in trunk there were cases where the day layout wasn't reactivated correctly. With the new implementation (riset.txt) this is corrected.
[Archiv.zip](https://github.com/navit-gps/navit/files/10366716/Archiv.zip)

OLFDB:
I'm fine with providing a test for this. Need to take over my Xcode project into whatever is used for unit testing. Do you have some information on how unit testing is planned to be done for navit?

OLFDB:
After detailed testing against https://api.sunrise-sunset.org I have a cleaner solution and will open a new pull request.

PR #1221, opened 2023-01-20
Fix:gui/internal: Fix mg: geo coordinates parsing

lains:
In my [previous PR](https://github.com/navit-gps/navit/pull/1213), I have implemented google coordinates-type parsing, but this breaks other format parsing. There is a bug and a variable naming is misleading.
The symptom is that trying to retrieve previous bookmarks or destinations will displayed anything anymore.

This PR fixes other parsing formats while still keeping the newly added format as valid (decimal latitude and longitude separated by a comma character, without any space).


christf:
Do we still need this? There are no changes in this branch that is not on trunk.

Issue #1179, opened 2022-02-21
ODS displays incorrect date on some days


pakeha-by:
How to reproduce:
- place OSD element `<osd enabled="yes" type="text" x="0" y="0" label="${vehicle.position_time_iso8601[local;%a, %d.%m.%Y]}" />` into navit.xml
- set current date to 21/02/2022
- wait till GPS data is availabe
- observe "Friday, 18.02.2022" displayed instead of expected "Monday, 21/02/2022" (date might one day earlier or later depending on current timezone but is 3 days earlier that expected day)

Reproduced on:
- raspbian / raspberry Pi
- windows 

Investigation shows that the problem is in function `iso8601_to_secs()`, which converts datetime string in iso 8601 format to unix epoch time. 
All 2022 dates were checked, for dates between 2022-02-01 and 2022-02-28 (inclusive), result of the iso8601_to_secs() function is 3 days earlier than unix time calculated by other mean.

PR #1222, opened 2023-02-04
Fix: #1179

OLFDB:
This fixes the issue #1179. Tested with testcases and working until Y2K38 by when we would need long values to be used.


jkoan:
In 87f7cfe294cb92f183e6ddaaa709789b40914c71 i added some basic tests. Which testcases have you found to test that your code is correct?

jkoan:
okay, got it. This mostly February. cbe618c81a2ec93a223a789273b9ded9c32cd2e3

Issue #1241, opened 2023-10-08
[Bug] Cannot skip interchange/exit names using TTS and samples


adevoss:
When using .wav files for TTS, Navit is silent when there is a interchange/exit/city name in the text.

There is an option to skip street names.
Please add an option to skip interchange/exit names.

If these are all 'street names' then it's a bug.

navit.xml (speech tag):
`vocabulary_name="0" vocabulary_name_systematic="0" vocabulary_distances="1"`

so no street names should be in the text but there is this error:

`error:speech_cmdline:speechd_say:No matching samples found. Cannot speak text: 'nu rechts aanhouden   richting RING-West'`

All the samples are there except `RING-West` (street name of motorway or name of interchange)

adevoss:
Also when using

navit.xml (speech tag):
`<speech type="cmdline" data="aplay -q %s"
                sample_dir="/data/navit/dir/dir" sample_suffix=".wav" flags="1"
                vocabulary_name="0" vocabulary_name_systematic="1" vocabulary_distances="1" active="1"/>`

there are interchange/exit names in the text.

At the moment TTS with samples is sort of usable but lot of text is not spoken because of this.
Working on a PR for a fix

adevoss:
Solved by #1259

PR #1250, opened 2023-11-01
play file instead of staying silent when sample is missing

adevoss:
Prerequisites
- Configure and create TTS sample files
- Add `missing.wav` file

`missing.wav` is played when a sample is missing. So you can hear the rest of the next turn instruction instead of hearing nothing.

#1241


adevoss:
Not tested yet because of not being able to compile (#1241)

adevoss:
This approach doesn't work. Will continue to search for a working solution.

adevoss:
Replaced by #1259

PR #1254, opened 2023-12-27
Rework the Binfile driver

jkoan:
The binfile driver is probably our most important driver regarding map data. 

The driver should be split in multiple units that are as self contained as possible. This will also help with the understanding as currently the binfile map-driver mixes being a downloader, a zip implementation, a tile implementation and probably something else.

Things to take care of:

- [x] Remove Download Feature from Binfile driver
Why: It is completely undocumented and mixes data access via multiple ways. We want to focus on one thing only per Plugin

For the Next PR:
- Split Zipfile loading/access and loading of tile data
- Future extend on the documentation

### DIFF
--- a/navit/map/binfile/binfile.c
+++ b/navit/map/binfile/binfile.c
@@ -83,19 +83,6 @@ struct tile {
     int mode;
 };
 
-struct map_download {
-    int state;
-    struct map_priv *m;
-    struct map_rect_priv *mr;
-    struct file *http, *file;
-    int zipfile, toffset, tlength, progress, read, dl_size;
-    long long offset, start_offset, cd1offset, size;
-    struct zip64_eoc *zip64_eoc;
-    struct zip64_eocl *zip64_eocl;
-    struct zip_eoc *zip_eoc;
-    struct zip_cd *cd_copy, *cd;
-};
-
 /**
  * @brief Represents the map from a single binfile.
  *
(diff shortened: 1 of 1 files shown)


christf (approved):
I like the removal. To make it build, I had to rebase it on top of the latest whitespace branch.

See https://github.com/christf/navit/tree/remove_binfile_download to see what I did.

hoehnp:
i found the functions while working on my cURL based downloader. Amazing that we had code for that in the bin driver. so I also agree with its removal.

navit/map/binfile/binfile.c
```diff
@@ -1538,7 +1089,6 @@ static void map_rect_destroy_binfile(struct map_rect_priv *mr) {
     if (mr->tiles[0].fi && mr->tiles[0].start)
         file_data_free(mr->tiles[0].fi, (unsigned char *)(mr->tiles[0].start));
     g_free(mr->url);
```
christf, line 1091: this should be removed along with the url field in the struct.

christf (changes requested):
one minor gotcha: the url field is now unneded and can be let go, I think.

navit/map/binfile/binfile.c
christf: this is now dead too and I think it should be removed along with that g_free() in line 878

navit/map/binfile/binfile.c
```diff
@@ -116,18 +102,8 @@ struct map_priv {
     int search_offset;
     int search_size;
     int version;
```
christf: with the removals from this PR, that field becomes unused. Let's remove it too.

christf (changes requested):
There are still two more fields that are unused now - I feel those should be removed to.

I do not think the callback list is actually used either but if we touch that then this changes signatures and I would stay clear of that for now. In any case, the url and version fields are now without use and I think we should let them go too.

christf (approved):

PR #1264, opened 2024-06-20
os:ubuntutouch: Add build files for Ubuntu Touch clickable package

metalstrolch:
This PR does add build files to build a Click package for UBPorts Ubuntu Touch. It is mainly based on the previous work for Sailfish Os,

Additionally this PR fixes coordinates from string parsing.

Additionally this PR fixes the current Android build by selecting ruby gems by specific version during CI.

PR #1266, opened 2024-06-27
fix:Define to only search for Python2 as Navit is not compatible with Python 3

jkoan:
On Systems that only have Python3 dev files cmake will find Python 3 and will try to build against it. This will not work at all


jkoan:
This PR is Blocked by old CI Tools as i needed to update the cmake version to 3.12to have FindPython2

jkoan:
Superseded by #1264

Issue #1227, opened 2023-02-24
Not able to open old Wiki and nor able to open forum


martin-ac:
Hello everybody,
for sure I would do what is asked for above, but currently and at least in the past 7 days I'm not - and I think nobody is - able to access
- the old Navit wiki at the above given address or to open https://wiki.navit-project.org
- same with the forum at https://forum.navit-project.org/
- BUT !!! am able to open https://www.navit-project.org/ without any problem or error.

I already wrote emails about this to four project managers but didn't get answers nor did this change anything. But maybe this is because the used mail domain is 
@navit-project.org which may not be working correct too?

I tried with current versions of firefox and chromium and 4 different pc's / nb's running Kubuntu 20.04 and 22.04 where I do not have any other problem with other pages in the internet and so am pretty sure there is an error with the above mentioned pages or servers hosting them.

Could someone please check and give me a feedback?
Thanks in advance!
Martin

jkoan:
Hi @martin-ac 
Cc @navit-gps/navit 
TL;DR: The server which hosted some of our infrastructure was taken offline. This server was provided by fosshost. Fosshost is closing down. Data is backed up and will be migrated.

Long Form:
Our Hoster for the navit server has shut down our server on February the 10th. I took a backup of all data. Now we still need to migrate all the relevant data, most important wiki.

We basically have the following services:
nginx (http frontend server) -> not needed any more
matterbridge -> can probably be hostet by me personally as I also host the IRC logs
forum -> discontinue, maybe achive it in some form (internet achive; archive.org?). Use GitHub discussions, GitHub issues?
cloud -> Not need anymore, only used for Fossdem Präsentation 
trac -> migrate to github issues
wiki -> migrate all content to Read the Docs #425 also see #1195 . Open: What about old links: Add a github pages site with mappings, and point wiki.navit-project to it? Alternatively less future proof export static version of the wiki and put on github pages?
ldap_gw -> not needed anymore 
backend (DB) -> definetly take dumps, but except from that also not needed anymore
ses (Mail Gateway) -> not needed anymore, was broken anyway 
www -> already on GitHub pages, so no problems 

For email I think this is still hosted on the even older servers which some of the initial team members set up. I never had access to those servers so I can't help with that. But yes has definitely nothing to do with that.

OLFDB:
Meanwhile you can use the [webarchive](https://web.archive.org/web/20230202050307/https://wiki.navit-project.org/index.php/Main_Page).

[...]

Issue #425, opened 2018-03-31
[Proposal] Move out of Mediawiki


aerostitch:
Hi,

The mediawiki website is quite old and automated crawlers think those are parked domains (example: http://duck.debian.net/static/sp/n/navit.html). Plus it might be a bit harsh for newcomers to find the informations right away.
What I would propose is to migrate it to github pages in markdown, plug it to jekyll (the one that comes with github pages) and maintain all that in a repo like navit.github.org.
I guess we could add sections to https://github.com/navit-gps/website
I would be up to do the migration myself if you all agree on that.

Let me know what you think.
Thanks for your feedback,
Joseph

sleske:
Could you explain why you feel the need to migrate? I like the fact that the Mediawiki site is easy to modify online, many people know the interface - what would migrating to Github pages buy us?

Your points about the wiki content being old and not always newbie-friendly is true, but that would not automatically get better by moving to GitHub.

aerostitch:
Hi,

> Could you explain why you feel the need to migrate? I like the fact that the Mediawiki site is easy to modify online, many people know the interface - what would migrating to Github pages buy us?

The main thing I could think of are:
* **security:**
  * Mediawiki, like any other CMS has several major security issues reported every months. If you want to have a look at the CVE related to mediawiki over time you can have a look at: https://www.cvedetails.com/product/4125/Mediawiki-Mediawiki.html?vendor_id=2360
   * The OS, the PHP engine, database engine and the webserver required to run this website also have regular CVE and need patching on regular basis.
   * using github pages means:
     * not having to maintain those and exposing less threats (you have people whose job is to maintain their infrastructure)
     * static-generated websites don't parse url arguments, don't accept POST, PUT, DELETE requests, don't have databases and run on read-only filesystems, regeneated with each commit you push to your repository, reducing the security risks by a huge deal.
     * the content is on a git repository meaning it can easily be reviewed using the usual PR system, just using a gh-pages branch
* **ease to use:**
  * I think about everybody knows how to write markdown and to push data to a git repo either by using the github web interface or using local tools and then push to the git repository
  * most people are used to the github wikis nowadays
  * As an infrequent contributor that sees a typo you don't have to have a login in a 3rd party to edit it, you just submit a PR
  * If you want to submit a change idea and get a review by several other people you can easily, again with a PR
* **content-proximity:**
[...]

PR #1281, opened 2024-12-12
Docs 2024

jkoan:
This RP imports the current state of the wiki and does a (big) overhaul to all the content of it as well.

As discussed in https://github.com/navit-gps/navit/issues/425 we should move out of our own Mediawiki and use the ReadTheDocs doc exclusively. Also the Wiki is down since a long time, so we should bring the docs back to live: #1227


hoehnp (approved):
I think it is a good starting point and we will step by step need to update everything.

PR #1285, opened 2024-12-23
test fix for fdroid

hoehnp:
this PR attempts to fix fdroid on circleci


jkoan:
try to use the newer images from fdroid from here: https://gitlab.com/fdroid/fdroidserver/container_registry/2649805 bullseye seems to be out of support. Probably the generic one without a release name would be a better fit for us.
So registry.gitlab.com/fdroid/fdroidserver:buildserver

jkoan:
See #1316

PR #1309, opened 2025-06-13
Add the word 'voice' to gettext

adevoss:


christf:
Sorry, I do not understand, what does this achieve?

adevoss:
See #1299

jkoan:
Unfortunately this is the wrong way to achieve this. First the pot needs to be updated and after that the po files can be translated on weblate

jkoan:
I opened up the issue https://github.com/navit-gps/navit/issues/1337 because the POt is often forgotten when updating code.

PR #1315, opened 2025-10-05
fix(speed) geoclue speed conversion from mps to kph

### LINKED ISSUE #1314: vehicle speed is off by factor of 3.6
the speed of the vehicle is reported lower than actual by factor of 3.6. This may be due to an incorrect conversion between m/s and km/h units.

This can be observed with the following in the config file:
`
<osd enabled="yes" type="text" label="${vehicle.position_speed}" x="85" y="100" font="Comic Sans MS" font_size="400" w="150" h="50" text_color="#ff0000" align="4" background_color="#00000000"/>
`

christf:
gclue_location_get_speed() returns a value in meters per second. In line with the gpsd implementation, this value is converted to kilometers per hour.

fixes #1314


jkoan:
Please use the already existing constant MPS_TO_KPH

christf:
> Please use the already existing constant MPS_TO_KPH

done. Thank you for the review.

PR #1316, opened 2025-10-06
Fix:Build F-droid fixed

christf:
this fixes the F-droid build on circleCI and bumps the ndk for both android and f-droid.

supersedes #1285 

- [x] find a way to build f-droid on circleci
- [x] simplify checkout logic (the directory layout is not sound, fdroid expects the data to sit in build/org.navitproject.navit 
- [x] ensure android builds
- [x] libnavit.so - find a way to embed this file into the apk
- [x] fix fastlane
- [x] embed the code from build_android.sh into cmake
- [x] fix the content of the apks. They seem to be missing the binary
- [ ] include libnavit_core in linux artefact

### DIFF
--- a/navit/android/AndroidManifest.xml
+++ b/navit/android/AndroidManifest.xml
@@ -9,6 +9,7 @@
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION"/>
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>


jkoan:
From what I see [here](https://gitlab.com/fdroid/fdroidserver/container_registry/2649805) Trixi is not a valid option. Probably the best would be to use the one without a suffix because that way we (hopefully) don't meet to change it in the future

mvglasow:
> Probably the best would be to use the one without a suffix because that way we (hopefully) don't meet to change it in the future

Doing so would upgrade to us to a new release whenever F-Droid decides to switch. If the new release comes with breaking changes, that would break things for us instantly and we would have to react quickly.

If, on the other hand, we explicitly rely on `bookworm` and monitor for new releases, we can try the new release, adapt CI to it and switch over once we have verified everything works.

jkoan:
@mvglasow yes, if they would make sure that the old image keeps working, but this isn't the case

christf:
it seems to me, the actual error is that warning:

```
Installing into /opt/android-sdk/ndk/20.1.5948944
2025-11-30 22:54:47,282 INFO: Set up NDK r20b (20.1.5948944)
2025-11-30 22:54:47,285 WARNING: NoSuchPathError: build/org.navitproject.navit
2025-11-30 22:54:47,288 ERROR: Could not build app org.navitproject.navit due to unknown error: Traceback (most recent call last):
  File "/home/vagrant/fdroidserver/fdroidserver/build.py", line 1188, in main
    if trybuild(app, build, build_dir, output_dir, log_dir,
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/fdroidserver/fdroidserver/build.py", line 931, in trybuild
    build_local(app, build, vcs, build_dir, output_dir, log_dir, srclib_dir, extlib_dir, tmp_dir, force, onserver, refresh)
  File "/home/vagrant/fdroidserver/fdroidserver/build.py", line 472, in build_local
    common.set_FDroidPopen_env(app, build)
  File "/home/vagrant/fdroidserver/fdroidserver/common.py", line 3406, in set_FDroidPopen_env
    env['SOURCE_DATE_EPOCH'] = get_source_date_epoch(get_build_dir(app))
    ~~~^^^^^^^^^^^^^^^^^^^^^
  File "<frozen os>", line 684, in __setitem__
  File "<frozen os>", line 758, in encode
TypeError: str expected, not NoneType

2025-11-30 22:54:47,288 DEBUG: Error encountered, stopping by user request.

Exited with code exit status 1
```

or in other words
```
2025-11-30 22:54:47,285 WARNING: NoSuchPathError: build/org.navitproject.navit
```

christf:
well, clearly whatever I am doing is not working. Any hints about that build directory?

christf:
it is getting there. I broke android but now we can build f-droid. After this experience, I think we might be transposing f-droid over to gha and fix both in the process.

navit/android/AndroidManifest.xml
```diff
@@ -23,6 +23,7 @@
             android:label="@string/app_name"
             android:theme="@style/NavitTheme"
             android:launchMode="singleTask"
+            android:exported="false"
```
jkoan: I am getting this error: 12-07 08:25:20.621  3635  3635 E BaseDraggingActivity: java.lang.SecurityException: Permission Denial: starting Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=org.navitproject.navit/.Navit bnds=[296,242][549,537] mCallingUid=10319 } from ProcessRecord{8b63d1f 3635:com.android.launcher/u0a319} (pid=3635, uid=10319) not exported from uid 10721 

So this one needs to be exported.

jkoan (changes requested):

navit/android/AndroidManifest.xml
christf: thank you for testing.

christf: done

jkoan:
The export of the Intent is fixed now. Now the next problem is that libnavit.so is not included in the APK.
And on another note, the artifact APK is not signed at all. But I am doing this myself for testing with apksigner and a throw away cert. 
On the question if we can replace the android build with the fdroid build and move that one to GitHub Actions, this is exactly what my goal is as well. But that's pending until we get the APK to function.

christf:
is there a way to test if the apk is sound from the pipeline? It would be great if it would only then be "green" when it really does work. 


How can I see if libnavit.so is included? If I can replicate, then I can patch. To that end: is libnavit.so also missing from the android build? Or is it just f-droid that is broken?

Edit: After discussing how to identify it is missing (download apk, unzip it, see if libnavit.so is in it), I found it is missing from the android-build as well.

I set up waydroid on my local machine, retrieved the apk from the pipeline but I am unable to install because waydroid claims it is invalid but won't tell me in which way.
invalid

jkoan:
> I set up waydroid on my local machine, retrieved the apk from the pipeline but I am unable to install because waydroid claims it is invalid but won't tell me in which way.
> invalid
> 

This is because the APK is not signed at all. You need apksigner to sign the resulting file. We will do this on non PRs with the navit key but on PRs we just skip this. Of course we could generate a throw away key  which would mean that navit needs to be completely removed after every test. Alternately we could generate a key per PR (and cache it) or generate a general Dev key that is different from the prod key and sign everything with it. 
But this is another topic for another issue/PR. As well as that I plan to provide our dev builds as a fdroid compatible repo once android is up to date again

christf:
For fastlane, see https://github.com/fastlane/fastlane/issues/22192

CMakeLists.txt
```diff
@@ -750,13 +750,11 @@ endif(HAS_IFADDRS)
 
 if(ANDROID)
 	message(STATUS "Using generator " ${CMAKE_GENERATOR})
```
jkoan, line 761: At the end of the cmake configure there's the message "configuration incomplete" which means that the cmake generate failed. Everything else is a consequent error. The real cmake error can then be found further down: https://github.com/navit-gps/navit/actions/runs/20117876574/job/57731551857#step:8:11469
Those lines can all go away, as this was the old old android build system where we also had ant instead or gradle ;)

christf: CMake Error at cmake/navit_macros.cmake:89 (message):
    Could not find 'Android SDK and AVD Manager'.  Please install the android
    sdk and add the <android-sdk>/tools directory to the search path.
  Call Stack (most recent call first):
    CMakeLists.txt:755 (message_error)

christf:
I wonder if we are doing something more fundamentally wrong in our build, if the icons require special handling for each target.

navit/icons/CMakeLists.txt
```diff
@@ -129,12 +133,18 @@ endforeach()
...
+			DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/${IMAGE_FILE}"
+			COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/${IMAGE_FILE}"
+			"${CMAKE_CURRENT_BINARY_DIR}/${IMAGE_FILE}"
+		)
+	else()
+		add_custom_command(OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${IMAGE_FILE}"
+			DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/${IMAGE_FILE}"
+			COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/${IMAGE_FILE}" "${CMAKE_CURRENT_BINARY_DIR}/${IMAGE_FILE}"
 		)
-	list(APPEND IMAGE_TARGETS "${CMAKE_CURRENT_BINARY_DIR}/${IMAGE_FILE}")
```
jkoan, line 137: @christf i think its this line which is falsely removed. Which in turn is also why linux is way smaller because the icons don't get installed/included

jkoan (approved):

PR #1319, opened 2025-10-12
ETA calculation

### LINKED ISSUE #1318: ETA too low
When a route is calculated the ETA is too low significantly.
In the car profile I changed speeds and weights but the ETA stayed the same.

When testing, there were no traffic jams. Obviously the average speed is always lower than the speed limit.

How is the ETA calculated? Can the ETA be influenced by settings?

### LINKED ISSUE #1320: ETA does not change if max speed of vehicleprofile is lower than max speed of roadsegments
The ETA is calculated using the max speed of the roadsegment. Even if the max vehicle speed for that segment is lower.
See PR https://github.com/navit-gps/navit/pull/1319

adevoss:
Fixes #1318 and fixes #1320
Testing will show If a factor of 0.9 is correct.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,30 +18,6 @@ jobs:
       #    name: Store checkstyle report
       #    path: navit/android/checkstyle
       #    destination: reports
-  build_linux:
-    <<: *defaults
-    steps:
-      - checkout
-      - run: |
-          apt-get update && apt-get install -y git
-          if scripts/check_need_build.sh; then circleci step halt; fi
-      - run:
-          name: Id
-          command: cat /etc/*release
-      - run:
-          name: Setup requirements
-          command: |
-            bash scripts/setup_common_requirements.sh
-            apt-get install -y libpng-dev libfreetype6-dev libdbus-glib-1-dev libgtk2.0-dev curl
-      - run:
-          name: Build for Linux
-          command: bash scripts/build_linux.sh
-      - store_artifacts:
-          path: linux/_CPack_Packages
-      - run:
-          name: Update Navit-Download-Center
-          command: |
-            bash scripts/update_download_center.sh
   run_doxygen:
     <<: *defaults
     steps:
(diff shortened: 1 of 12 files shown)


christf:
what did testing show? I have found navits ETA to be a bit more optimistic than that of one other widely used navigation project. I thought this might have to do with traffic but I could be wrong.

In terms of the code, I would have a few nits (rebase on latest master and thus get rid of merge commits in the feature branch, rework wording of commit messages and scope https://cbea.ms/git-commit/, fix whitespace on indents, trailing blanks and consistent blanks around assignments) before I personally can review the contribution properly (I get easily confused when I read a commit message "do not change <entity>" for a commit that does changes in that entity).

What I see so far seems to make sense to me and should reach a better value for ETA especially for slow vehicles.

Edit: Do let me know if you need help with the git groundwork. I am also on the discord, if you like to find me there.

adevoss:
> what did testing show?

A route which I travel often takes me 55 minutes by car (without traffic jams). Without this PR Navit showed 43 minutes.
Using this PR Navit shows 54 minutes.

christf:
adevoss, if you don't mind, please have a look at the other pr and how it presents the goal of the change as well as the clarity of the commit history. I created it for two reasons:

* to show what I meant by my ask for rebase, whitespace and commit structure.
* to contribute to preparation of the code for merge.

This is what I did, while on your feature branch:

git fetch --all
git rebase -i upstream/trunk # (upstream is the name of the remote of this very git repository in my local setup)

Then git asked me about each commit and how I wanted to handle it. In this case during code review, I recognized all commits belong to the same logical unit, so I squashed them (using squash command) into one single commit. 
In the code review I had earlier seen a few whitespace problems, so I fixed those and amended the existing commit by those whitespace changes (git commit --amend).

Then I sat down to find a commit message that would describe the goal of the patch, hopefully capturing its essence in the subject line. This is important because it helps maintainability in the long run. Notice, how the commit message is in line with the commit guide for the navit project, which demands declaration of type of change and area.

Then I pushed and created the pr. It is still your code, even the authorship is maintained, only the result will be better maintainable and reviewable.

With the code in that shape, I am happy to try it out on my own device.

Please take all of this as the act of kindness, it is intended to be. I see you are creating code for this project, I want it to succeed and to help others that also want to improve it. Thank you for contributing.

adevoss:
Why not just add a commit to remove whitespaces if that bothers you?
After creating this PR I tested and found my first thought of a general correction factor was not needed.
That's all part of developing code.

Taking this more general:
Why cleaning up commits and create a new PR in your own name? That erases history and information for future developers.
If deleting commits is the way to go then every PR should have one commit.

In my opinion this is not the way to go.

christf:
> Why not just add a commit to remove whitespaces if that bothers you? After creating this PR I tested and found my first thought of a general correction factor was not needed. That's all part of developing code.
> 
> Taking this more general: Why cleaning up commits and create a new PR in your own name? That erases history and information for future developers. If deleting commits is the way to go then every PR should have one commit.
> 
> In my opinion this is not the way to go.

It actually is best practice. Someone else has written it down before, so I will refer https://thelinuxcode.com/git-squash-explained-a-guide-for-keeping-commit-history-clean/

and I will also add https://navit.readthedocs.io/en/trunk/development/commit_guidelines.html
"try to have one commit per feature".

jkoan:
But on GitHub we have the option to squash the commits when merging. Which I will do on a case by case basis. I don't really care about the commit history on the PR. But I will look at it and keep it if it's clean (do a merge commit or rebase) or if not I will squash them into one commit and provide a short commit message. 

But I also wanna say that @christf commit messages are really good so I would keep them. But I don't think that reworking the commit history of someone else is worth it and like @adevoss already said it can feel like somebody is stealing the credit which should also not be what we wanna do.

christf:
> But I also wanna say that @christf commit messages are really good so I would keep them. But I don't think that reworking the commit history of someone else is worth it and like @adevoss already said it can feel like somebody is stealing the credit which should also not be what we wanna do.

Does it? The rework retains the author, and I did not see a mention of @adevoss about credit stealing. I brought it up to make it clear that doesn't happen in my view. (found the remark now, I had overlooked it earlier. No, I do not think that the person raising a PR is taking credit for the commits contained in it. Those are visible in the git history and those retain the original author. This is visible from the other PR that contains the squashed history too)

The problems I have with a commit history that describes the journey is:
* it causes additional work when reviewing - I can't review it without rebasing and rewriting it. So it is keeping me from making a contribution. If it happens to me, it could happen to others as well.
* It causes additional work when merging (which is a problem because this project - as any other - is tight on resources)
* it delays merges and therefore important improvements are made later or not at all as a result of the above two points. We can see it in this PR. The code itself was fine but it wasn't merged for months.
* if that additional work is not done, then there is clutter in the target branch that is decreasing maintainability - and it is easy to miss especially when there is so much else to do.

PR #1324, opened 2025-11-06
maptool: Enable Town Search when processing map regions by hard-coding the country

### LINKED ISSUE #1323: Unable to find towns when using regional maps (with no country boundary in them)
When having the sachsen-anhalt from 2025-09-24 map loaded, searching for its capital "Magdeburg" does not return any results.

christf:
When processing a part of a larger map, for example a region in Germany, town search will not work because the towns are not inside a larger country polygon as that is missing from the map data. This introduces a new command line switch -C that may be used together with -U. When -C is passed, each town in an unknown country will be placed in the country <ISO-code>.

This fixes #1323

### DIFF
--- a/navit/maptool/osm.c
+++ b/navit/maptool/osm.c
@@ -2045,8 +2045,13 @@ static struct town_country *town_country_list_insert_if_new(GList **town_country
 
 static GList *osm_process_town_unknown_country(void) {
     static struct country_table *unknown;
-    if (!unknown)
-        unknown = country_from_countryid(999);
+    int country = 999;
+
+    if (!unknown) {
+        if (!!hardcoded_country)
+            country = country_id_from_iso2(hardcoded_country);
+        unknown = country_from_countryid(country);
+    }
 
     return g_list_prepend(NULL, town_country_new(unknown));
 }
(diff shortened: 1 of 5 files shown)


navit/maptool/osm.c
```diff
@@ -2018,8 +2018,14 @@ static struct town_country *town_country_list_insert_if_new(GList **town_country
...
 static GList *osm_process_town_unknown_country(void) {
     static struct country_table *unknown;
-    if (!unknown)
-        unknown=country_from_countryid(999);
+    int country = 999;
+
+    if (!unknown) {
+        if (!!hardcoded_country)
+            country = country_id_from_iso2(hardcoded_country);
+        printf("Hard-Coding country: %s(%d)\n", hardcoded_country ? hardcoded_country : "", country);
```
christf: drop this printf

christf:
<put into the issue, discussing the problem>

christf:
Do we still need this (in a cleaned up version of course)?
It might provide a fallback mechanism in situations currently unknown to us and it is not much code.

But it is additional code that should exclusively be used in this very edge case, so noone will notice if it breaks.

I tend to think we should let this go.

christf:
closing for now, we can come back to it, if we need it.

Issue #1325, opened 2025-11-14
localized maps


christf:
When travelling to cultures that use different alphabets and language, it would be helpful to offer localized maps which contains transcriptions of the relevant names.

As in the issue about town search, I have first hacked maptool to use localized names, if available. Navit deals with this just fine in terms of display, search and routing.
Based on this, there are multiple potential paths to consider:

1. navit including maptool could be enhanced to bake all localized names into the bin files. Upon display/search/route navit would then have to select which label to display/search/route.
2. navit could offer localized maps. 
3. navit could push the map processing into the client during or after download time. Whether or not to use localization and which language could be part of the config file. This would allow us to decommission much of the map infrastructure and start using pbf files from ie geofabrik.
4. maptool could offer to localize without changes in navit, this would always work with localized maps. Users interested in the feature can use maptool to generate their own bin files in their language of choice.

Each of these options comes at a price. (1) is expensive to build but it would be the most versatile. Once in that foreign location with the different alphabet, it might be helpful to be able to search locations that can be copied from signs at the side of the road. (2) is expensive to run (the mapserver is loaded as it stands. offering many languages may just not be possible). (3) pushes maptool processing into the client. While maptool already can read from stdin, it might overload clients on the weaker end of the hardware spectrum. 
(4) does not allow searching for entities using their non-localized names and pushes the burden of creating the maps onto users.

PR #1326, opened 2025-11-14
DRAFT: prefer localized german names if they exist

christf:
This implements an illustration of option 4 outlined in #1325.

Of course merging a patch that prefers german names is nothing we would want or intend. The goal is to document the general idea and keep code in public.

### DIFF
--- a/navit/maptool/osm.c
+++ b/navit/maptool/osm.c
@@ -42,9 +42,16 @@
 #include "glib.h"
 #include "item.h"
 #include "item_type_def.h"
+#include "linguistics.h"
+#include "profile.h"
 #include "maptool.h"
 #include "projection.h"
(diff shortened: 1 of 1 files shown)


christf:
As an idea, this could have a linked list with all languages. Then navit can select which name (name:<locale>) to display and fallback to the name-tag without translation.

This would work in the general case and based on NLS-settings, the user could then decide if they want localization.

christf:
taken from discord: "Maybe we just need to add a new attr type to implement it.
Like 
Array: 3
translation: "[locale]:[translates string]"
translation: "Y:***"
translation: "Z:???""

jkoan:
This one is superseded by #1556 so it can already be closed right?

PR #1334, opened 2025-12-01
RFC: clang-format introduced and showcased on vehicle.c

christf:
This introduces a .clang-format styleguide for formatting and applies it treewide. The idea is to keep the codebase in a clean and maintainable way towards a central baseline that is verifiable.

Personally, I do not mind which standard to follow, the idea is to have one.
This changeset was created by running:

```

clang-format -i vehicle.c

```

### DIFF
--- a/navit/vehicle.c
+++ b/navit/vehicle.c
@@ -565,11 +565,16 @@ static void vehicle_log_gpx(struct vehicle *this_, struct log *log) {
         this_->meth.position_attr_get(this_->priv, attr_position_height, &attr))
         logstr = g_strconcat_printf(logstr, "\t<ele>%.6f</ele>\n", *attr.u.numd);
     // <magvar> magnetic variation in degrees; we might use position_magnetic_direction and position_direction to figure
-    // it out <geoidheight> Height (in meters) of geoid (mean sea level) above WGS84 earth ellipsoid. As defined in NMEA
-    // GGA message (field 11, which vehicle_wince.c ignores) <name> GPS name (arbitrary) <cmt> comment <src> Source of
-    // data <link> Link to additional information (URL) <sym> Text of GPS symbol name <type> Type (classification) <fix>
-    // Type of GPS fix {'none'|'2d'|'3d'|'dgps'|'pps'}, leave out if unknown. Similar to position_fix_type but more
-    // detailed.
+    // it out
+    // <geoidheight> Height (in meters) of geoid (mean sea level) above WGS84 earth ellipsoid. As defined in NMEA GGA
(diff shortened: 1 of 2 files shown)


christf:
What are your thoughts? Should I go on to process all files?

navit/vehicle.c
```diff
@@ -542,77 +529,75 @@ void vehicle_log_gpx_add_tag(char *tag, char **logstr) {
...
-        logstr=g_strconcat_printf(logstr,"\t<hdop>%.6f</hdop>\n",*attr.u.numd);
+    if (attr_types_contains_default(attr_types, attr_position_height, 0) &&
+        this_->meth.position_attr_get(this_->priv, attr_position_height, &attr))
+        logstr = g_strconcat_printf(logstr, "\t<ele>%.6f</ele>\n", *attr.u.numd);
+    // <magvar> magnetic variation in degrees; we might use position_magnetic_direction and position_direction to figure
+    // it out <geoidheight> Height (in meters) of geoid (mean sea level) above WGS84 earth ellipsoid. As defined in NMEA
+    // GGA message (field 11, which vehicle_wince.c ignores) <name> GPS name (arbitrary) <cmt> comment <src> Source of
+    // data <link> Link to additional information (URL) <sym> Text of GPS symbol name <type> Type (classification) <fix>
+    // Type of GPS fix {'none'|'2d'|'3d'|'dgps'|'pps'}, leave out if unknown. Similar to position_fix_type but more
+    // detailed.
```
jkoan: This on is butchered, the variable names where on other lines to make it more readable

jkoan:
> What are your thoughts? Should I go on to process all files?

Yes, but only on the ones which are not support and not fib-1.1. Which reminds me that we should move fib to support as well

navit/vehicle.c
christf: indeed. This happens, when clang-format decides to reflow a multi-line comment due to line-length restrictions. After manually putting in the linebreaks, it will not touch these comments again, so no change to the format-file is required but rather as part of the reformatting PR, I should have checked for those reflows and made sure they are sound. Now that I know of this problem, I can do so for the next files.

christf:
> > What are your thoughts? Should I go on to process all files?
> 
> Yes, but only on the ones which are not support and not fib-1.1. Which reminds me that we should move fib to support as well

Later, this might become part of a check in the pipeline. For this, an exception already is in place for everything in the support directory. Once we move fib, it won't be subject to reflows.

jkoan:
Any objections on merging this already and doing the next files in batches in new PR's?

christf:
not at all. For the next change I would also update the sanity check script.

PR #1339, opened 2025-12-03
Fix:maptool: circular junctions allow oneway to be specified

### LINKED ISSUE #1331: incorrect routing on circular junction
navit told me to turn sharp left at this spot https://www.openstreetmap.org/way/152411244 when approaching from "Pfälzer Straße" to turn into "Hohenstaufenring" - this is incorrect because there is a circular junction there that is one-way and the route should have been to follow that circular junction and take the 4th exit.  From what I can tell, the OSM data is correct in this case and navit is drawing incorrect conclusion (not considering a circular junction is one-way only and thus sending drivers into the one direction of a one-way street)

christf:
this follows up on 191529acab24e0cea8f70351cb1ec7e0398fb0cc that implemented circular junctions to default to one-way streets. While already an improvement, it is not entirely accurate as in fact there are circular junctions that are specified as one-way, there are bidirectional circular junctions and there are circular junctions that lack the oneway tag entirely.

This implements circular junctions to default to oneway=yes, and considers a oneway-tag, should it be set.

This fixes #1331 for real, this time.

routing for circular junctions with oneway=yes:
circular_oneway=yes

routing for circular junctions with oneway=no:
circular_oneway=no

routing for circular junctions without oneway tag:
circular_no_oneway

### DIFF
--- a/navit/maptool/osm.c
+++ b/navit/maptool/osm.c
@@ -1082,7 +1082,11 @@ void osm_add_tag(char *k, char *v) {
         if (!g_strcmp0(v,"1")) {
             flags[0] |= AF_ONEWAY | AF_ROUNDABOUT_VALID;
         }
-        if (! g_strcmp0(v,"-1")) {
+        else if (! g_strcmp0(v,"0")) {
+            flags[0] &= ~AF_ONEWAY;
+            flags[0] &= ~AF_ONEWAYREV;
+        }
+        else if (! g_strcmp0(v,"-1")) {
             flags[0] |= AF_ONEWAYREV | AF_ROUNDABOUT_VALID;
         }
         if (!in_way)


christf:
I will have to say: This code has one problem known to me currently as there is a dependency on the proper ordering of tags. So far the osm-files contained tags ordered alphabetically but I do not think this is standardized. When alphabetical, junction-tags get processed before oneway-tags. This is the scenario in which it works.

If the sort order is different and oneway-tags might be handled before junction-tags, then the code falls back to the default of junctions being oneway. Again, this is an improvement over the current state, but it is still not perfect. To make it perfect, the tag parsing needs to be refactored to allow considering the oneway-tag while processing the junction-tag. Currently it is not possible to do that because when handling the junction tag, there is no way to identify whether oneway been processed and is "no", or whether it is undefined or whether it has not been processed yet but it is actually "yes".

PR #1340, opened 2025-12-04
Fixing compiler warnings

christf:
This fixes some compiler warnings and corresponding issues.


christf:
Codefactor is unhappy with the code (that existed before this PR). In this particular case I would like to be able to silence this warning. I do not think the popen() call is actually dangerous here, because the data used to construct it comes from the config file owned by the user.

PR #1349, opened 2025-12-04
Traff http plugin developed by mvglasow

### LINKED ISSUE #1359: Provide traffic by default
On targets which usually have Internet access include the traff_http and ship an disabled configuration for it.
Why in a disabled state? Navit is an offline app which in the coming releases will introduce online functionality link downloading maps or as in this issue traffic data. But we don't want to become online first. Every online action should be a deliberat choice by the user

christf:
This implements the traff_http plugin, adds documentation

I hope you don't mind me raising this PR @mvglasow  - I took the last few days to rebase and test it, I feel it should be upstreamed. I do think there is still a bit of work here, adding a tasklist.

- [x] fix crash in traffic.c:3624
- [x] fix NLS bug for win32 too or disable the plugin for windows (current approach removes the patch from win32 to get it to compile, this is not good)
- [x] fix the linker problem - the so file for traff-http must understand where libnavit_core is installed.
- [x] avoid gui freezes => find cause of slow startup. During usage it does not seem to be a problem
- [x] investigate, why sometimes(TM) minlat > maxlat and fix that
- [x] investigate if sometimes minlon > maxlon
- [x] enable traff_http on android
- [x] ensure the documentation contains "[...] stating that the server is intended for development use only, as it is still in early alpha state, has not been thoroughly tested and is therefore not yet recommended for production use."
- [ ] now that the build works, have another look at commit structure, reshape as necessary


mvglasow:
Certainly no objections from my side. (Also no guarantees – I haven’t touched this code in a while.)

navit/thread.c
```diff
@@ -0,0 +1,544 @@
...
+ * @param data Pointer to a `struct thread_main_data` encapsulating the main function and its argument.
+ */
+static void *thread_main_wrapper(void * data) {
+    struct thread_main_data * main_data = (struct thread_main_data *) data;
+    void * ret = (void *) (main_data->main(main_data->data));
+    // FIXME free up `data`
+    return ret;
+}
+
+char * thread_format_error(int error) {
```
christf: Codefactor does not like this function. The message is that this is too complex.
I disagree with codefactor on this one. Yes, it is a long function but it contains only one switch/case statement. Changing that will require it to be broken into multiple cases actually increase complexity and reduce clarity of code. @jkoan how should we proceed for this one? Do we have the option to ignore/accept some findings as false positives? How would we do it?

jkoan: We can just ignore it, we don't need to do anything about codefactor. I agree that this function is really simple. One solution which would be possible but doesn't gain anything is a map with the error code as key and the string as message. But this isn't needed

christf: Codefactor says, there should be an "ignore button" in its interface - but I do not see it. Is this a permissions-thing or am I missing something on my end? I would love to present a green build as part of a PR...

jkoan: Done. I had the option

christf:
just saw a crash.

```
error:map_textfile:get_line:Unable to get line (Erfolg)
error:map_textfile:get_line:Unable to get line (Erfolg)
error:map_textfile:get_line:Unable to get line (Erfolg)
error:map_textfile:get_line:Unable to get line (Erfolg)
error:map_textfile:get_line:Unable to get line (Erfolg)
BACKTRACE ------------
navit(+0x45281) [0x55943d325281]
/lib/x86_64-linux-gnu/libc.so.6(+0x3fdf0) [0x7f5e2ece8df0]
navit(+0x2cb91) [0x55943d30cb91]
navit(+0x2f114) [0x55943d30f114]
navit(map_rect_new+0x27) [0x55943d3353c7]
navit(+0x50c56) [0x55943d330c56]
navit(navit_draw_async+0x61) [0x55943d339361]
navit(+0x59465) [0x55943d339465]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_closure_invoke+0x140) [0x7f5e2e692950]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x2bd43) [0x7f5e2e6a6d43]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x2d769) [0x7f5e2e6a8769]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0x36) [0x7f5e2e6ae5a6]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit+0x93) [0x7f5e2e6ae663]
/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(+0x7b844) [0x7f5e2de7b844]
/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(+0x214a0d) [0x7f5e2e014a0d]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_closure_invoke+0x140) [0x7f5e2e692950]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x2bd43) [0x7f5e2e6a6d43]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x2d769) [0x7f5e2e6a8769]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0x36) [0x7f5e2e6ae5a6]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit+0x93) [0x7f5e2e6ae663]
/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(+0x95266) [0x7f5e2de95266]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_closure_invoke+0x140) [0x7f5e2e692950]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x2be70) [0x7f5e2e6a6e70]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x2d769) [0x7f5e2e6a8769]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0x36) [0x7f5e2e6ae5a6]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit+0x93) [0x7f5e2e6ae663]
/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(+0x94069) [0x7f5e2de94069]
/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(+0x140faa) [0x7f5e2df40faa]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_closure_invoke+0x140) [0x7f5e2e692950]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x2c320) [0x7f5e2e6a7320]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x2d032) [0x7f5e2e6a8032]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0x36) [0x7f5e2e6ae5a6]
/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit+0x93) [0x7f5e2e6ae663]
/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(+0x265dcc) [0x7f5e2e065dcc]
/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(gtk_propagate_event+0x15c) [0x7f5e2df3f62c]
/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(gtk_main_do_event+0x34b) [0x7f5e2df3f9cb]
/lib/x86_64-linux-gnu/libgdk-x11-2.0.so.0(+0x61a60) [0x7f5e2e4a1a60]
/lib/x86_64-linux-gnu/libglib-2.0.so.0(+0x5c385) [0x7f5e2f00b385]
/lib/x86_64-linux-gnu/libglib-2.0.so.0(+0x5e5b7) [0x7f5e2f00d5b7]
/lib/x86_64-linux-gnu/libglib-2.0.so.0(g_main_loop_run+0x12f) [0x7f5e2f00e01f]
navit(main_real+0x599) [0x55943d3035c9]
/lib/x86_64-linux-gnu/libc.so.6(+0x29ca8) [0x7f5e2ecd2ca8]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x85) [0x7f5e2ecd2d65]
navit(_start+0x21) [0x55943d302f31]
----------------------
GNU gdb (Debian 16.3-1) 16.3
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from navit...
Attaching to program: /usr/local/bin/navit, process 122717
[New LWP 122722]
[New LWP 122721]
[New LWP 122720]
[New LWP 122719]
[New LWP 122718]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
__syscall_cancel_arch () at ../sysdeps/unix/sysv/linux/x86_64/syscall_cancel.S:56

warning: 56	../sysdeps/unix/sysv/linux/x86_64/syscall_cancel.S: Datei oder Verzeichnis nicht gefunden
#0  __syscall_cancel_arch () at ../sysdeps/unix/sysv/linux/x86_64/syscall_cancel.S:56
#1  0x00007f5e2ed38668 in __internal_syscall_cancel (a1=<optimized out>, a2=a2@entry=140736313112488, a3=a3@entry=0, a4=a4@entry=0, a5=a5@entry=0, a6=a6@entry=0, nr=61) at ./nptl/cancellation.c:49
#2  0x00007f5e2ed386ad in __syscall_cancel (a1=<optimized out>, a2=a2@entry=140736313112488, a3=a3@entry=0, a4=a4@entry=0, a5=a5@entry=0, a6=a6@entry=0, nr=61) at ./nptl/cancellation.c:75
#3  0x00007f5e2eda3787 in __GI___wait4 (pid=<optimized out>, stat_loc=stat_loc@entry=0x7fffb9f337a8, options=options@entry=0, usage=usage@entry=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
#4  0x00007f5e2eda37b7 in __GI___waitpid (pid=<optimized out>, stat_loc=stat_loc@entry=0x7fffb9f337a8, options=options@entry=0) at ./posix/waitpid.c:38
#5  0x00007f5e2ecfbf7b in do_system (line=line@entry=0x7fffb9f33b50 "gdb -ex bt -ex detach -ex quit navit 122717") at ../sysdeps/posix/system.c:172
#6  0x00007f5e2ecfc11a in __libc_system (line=line@entry=0x7fffb9f33b50 "gdb -ex bt -ex detach -ex quit navit 122717") at ../sysdeps/posix/system.c:212
#7  0x000055943d3252c5 in sigsegv (sig=<optimized out>) at /home/christof/Programmierung/navit/navit/debug.c:128
#8  0x00007f5e2ece8df0 in <signal handler called> () at /lib/x86_64-linux-gnu/libc.so.6
#9  0x000055943d30cb91 in traffic_message_restore_segments (this_=this_@entry=0x55945b69ec20, ms=<optimized out>, map=0x55945b612520, route=0x55945b4fd8b0)
    at /home/christof/Programmierung/navit/navit/traffic.c:3620
#10 0x000055943d30f114 in tm_rect_new (priv=0x55945b6126c0, sel=0x55945b9193e0) at /home/christof/Programmierung/navit/navit/traffic.c:980
#11 0x000055943d3353c7 in map_rect_new (m=0x55945b612520, sel=0x55945b9193e0) at /home/christof/Programmierung/navit/navit/map.c:298
#12 0x000055943d330c56 in do_draw (displaylist=0x55945b16c6d0, cancel=cancel@entry=0, flags=flags@entry=1) at /home/christof/Programmierung/navit/navit/graphics.c:3296
#13 0x000055943d3314d9 in graphics_load_mapset
    (gra=<optimized out>, displaylist=<optimized out>, mapset=<optimized out>, trans=<optimized out>, l=<optimized out>, async=<optimized out>, cb=<optimized out>, flags=<optimized out>)
    at /home/christof/Programmierung/navit/navit/graphics.c:3499
#14 0x000055943d339361 in navit_draw_async (this_=<optimized out>, async=0) at /home/christof/Programmierung/navit/navit/navit.c:396
#15 navit_draw_async (this_=0x55945b166b70, async=0) at /home/christof/Programmierung/navit/navit/navit.c:389
#16 0x000055943d339465 in navit_scale (this_=0x55945b166b70, scale=8192, p=0x0, draw=1) at /home/christof/Programmierung/navit/navit/navit.c:750
#17 0x00007f5e2e692950 in g_closure_invoke () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#18 0x00007f5e2e6a6d43 in ??? () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#19 0x00007f5e2e6a8769 in ??? () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#20 0x00007f5e2e6ae5a6 in g_signal_emit_valist () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#21 0x00007f5e2e6ae663 in g_signal_emit () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#22 0x00007f5e2de7b844 in ??? () at /lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0
#23 0x00007f5e2e014a0d in ??? () at /lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0
#24 0x00007f5e2e692950 in g_closure_invoke () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#25 0x00007f5e2e6a6d43 in ??? () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#26 0x00007f5e2e6a8769 in ??? () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#27 0x00007f5e2e6ae5a6 in g_signal_emit_valist () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#28 0x00007f5e2e6ae663 in g_signal_emit () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
```

mvglasow:
Then `traffic.c:3620` should be where it’s at (according to line 9 of the stack trace).

christf:
blue_lines

Somehow the map looks weird with all those blue lines.

mvglasow:
**Edit: See next comment before trying any of this**

Odd indeed, looks like a bunch of streets got added as item type `street_route` – that is, if you are using the default map style. Traffic distortions should appear in orange (original style before I started work) or dark red (which I added to the custom style I used, don’t remember if I also changed that in the default style). I have used Navit with traffic quite extensively, and never encountered anything like this.

Try examining one of the blue segments (I think this can be done by right-clicking in GTK) and check its item type. (You may have multiple items stacked on top of each other, such as a street type, a route type and a traffic distortion type.) If you find a `street_route` type, then Navit is erroneously adding route segments somewhere.

To figure out if that is from the traffic plugin, you can try disabling the map (with the internal GUI this should be in **Settings > Maps > Traffic**, then disable the Traffic map) and see if the blue lines disappear.

If the blue lines disappear after that, then for some reason the plugin is assigning the wrong item type (`street_route` instead of `traffic_distortion`).

If the blue lines are still there, it is probably not the traffic plugin (or at least not the map provider) that is at fault.

You can also try to disable the Routing map in the same way – if that makes the blue lines go away, the fault is likely with the routing engine.

Background info – the traffic plugin uses parts of the routing engine to decode a pair of points (`from`/`to`) into the corresponding road segments. That means multiple instances of the routing engine being active, but that should not be an issue for Navit – unless there is some regression which messes with that separation.

To find out if there was some regression since I last touched the code, check out 329c6a22, build and test. If 329c6a22 is OK, try `git bisect`:
* `git checkout traff-http` (assuming that is the branch you are working on)
* `git bisect start`
* `git bisect bad` (marks current version as bad – `git bisect` internally just maintains two commit references, one `good` and one `bad`, and points them to the last commit you marked as `good` or `bad`)
* `git bisect good 329c6a22`
* Git will check out a commit halfway between `good` and `bad`. Build and test.
  * If the error is present in that commit, run `git bisect bad`.
  * If the error is not present, run `git bisect bad`.
  * If you can’t test the current commit (e.g. because it fails to build, or has an unrelated error), run `git bisect skip`.
* The previous step will repeat until `good` is the parent of `bad` (in which case `bad` is the commit that introduced the error), or `good` and `bad` are separated by nothing but skipped commits (in which case either `bad` or one of the skipped commits introduced the error).

mvglasow:
Looking again, the whole map style looks weird. The motorways (A2 north of Magdeburg, A14 west of it and its southeastern continuation) are just thin gray lines. Bundesstraßen alternate between red lines and yellow lines with red casing and red centerline, with no apparent changes in tagging on the map. Maybe check your map style and the underlying base map before you try any of the things I suggested above. If you disable the binfile map but leave traffic and route active and the blue lines go away, it might be your map style or the map data itself.

christf:
> Looking again, the whole map style looks weird. The motorways (A2 north of Magdeburg, A14 west of it and its southeastern continuation) are just thin gray lines. Bundesstraßen alternate between red lines and yellow lines with red casing and red centerline, with no apparent changes in tagging on the map. Maybe check your map style and the underlying base map before you try any of the things I suggested above. If you disable the binfile map but leave traffic and route active and the blue lines go away, it might be your map style or the map data itself.

if I remove that traffic.xml from ~/.navit, then the style looks ok again. So this traffic.xml somehow triggers a crash (fixed with the latest commit) but it also influences the layout somehow.

mvglasow:
> if I remove that traffic.xml from ~/.navit, then the style looks ok again. So this traffic.xml somehow triggers a crash (fixed with the latest commit) but it also influences the layout somehow.

Are you saying the roads (motorways, national roads) render differently with traffic.xml vs. without?

christf:
> > if I remove that traffic.xml from ~/.navit, then the style looks ok again. So this traffic.xml somehow triggers a crash (fixed with the latest commit) but it also influences the layout somehow.
> 
> Are you saying the roads (motorways, national roads) render differently with traffic.xml vs. without?

exactly this.
with this traffic.xml: 
[traffic.xml](https://github.com/user-attachments/files/24214250/traffic.xml)

mvglasow:
Whenever I got that kind of weird and inexplicable behavior, it was usually caused by reading from memory that isn’t yet initialized, or has been freed already. Valgrind should help shed light on this, I once wrote instructions on how to run it from within Eclipse (which is way more convenient than from the command line – but if you use gdb, maybe the command line doesn’t scare you :-)

christf:
will have a look. weird though that asan did not trigger....

also, eclipse scares me more than the cli.

christf:
well, this cost me most of my navit-time today. The cause was, the maps were created with a maptool that had its item_defs.h deviate from navits (the binary) item_defs.h. This caused all sorts of weird trouble. It wasn't the new code at all.

Valgrind does show two values being used without initialization (and it should be fixed), but this is not new to the branch.

navit/traffic/traff_http/traffic_traff_http.c
```diff
@@ -0,0 +1,651 @@
...
+        curl_easy_setopt(curl_handle, CURLOPT_URL, url);
+
+        curl_easy_setopt(curl_handle, CURLOPT_POSTFIELDS, data);
+
+        /* provide a callback and buffer for result data */
+        curl_easy_setopt(curl_handle, CURLOPT_WRITEFUNCTION, curl_result_callback);
+        curl_easy_setopt(curl_handle, CURLOPT_WRITEDATA, (void *)ret);
+
+        // FIXME provide a meaningful user agent
+        curl_easy_setopt(curl_handle, CURLOPT_USERAGENT, "libcurl-agent/1.0");
```
jkoan: How About "Navit/Major.Minor.Patch (TraffXML)" ?

navit/traffic.c
```diff
@@ -4839,44 +4858,65 @@ static void traffic_xml_start(xml_context *dummy, const char *tag_name, const ch
      */
 }
 
+static int floatparse(char *line, gdouble *lat, gdouble *lon) {
+    char *foo = g_strdup(line);
+    char *token = strtok(foo, " ");
+
+    if (token) {
+        *lat = g_strtod(token, NULL);
```
jkoan: g_strtod   is locale depended as well, but g_ascii_strtod is not. And its already in support/glib https://github.com/navit-gps/navit/blob/e85ad7de73aaf637a9acc041eaf0e849f0b751d7/navit/support/glib/gstrfuncs.c#L362

christf:
Just adding a bit of progress:
This is an area of Hannover showing congestion that is marked in orange.
distortions

juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **2 points that might be worth attention** (could be false positives, please use your judgment):

1. <strong>Refining thread safety in debug level configuration</strong>
   - [navit/debug.c:L160-L177](https://github.com/navit-gps/navit/blob/1fa9edd764eae35c227f47c782d7c6b535950476/navit/debug.c#L160-L177): `navit/debug.c:L160-L177`
   - Impact: Concurrent updates to debug levels while logging is active could lead to memory instability or unexpected crashes.
   - Suggestion: Consider acquiring the `rw_lock` in write mode before modifying the shared hash table to ensure synchronization.

2. <strong>Enhancing float-to-string conversion for negative values</strong>
   - [navit/util.c:L520-L525](https://github.com/navit-gps/navit/blob/1fa9edd764eae35c227f47c782d7c6b535950476/navit/util.c#L520-L525): `navit/util.c:L520-L525`
   - Impact: Negative coordinates might be formatted with unexpected symbols (like double signs), potentially affecting traffic API data accuracy.
   - Suggestion: Consider using absolute values for the fractional part and handling the negative sign explicitly to ensure consistent formatting.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

christf:
```
error:traffic_traff_http:traffic_traff_http_worker_thread_main:sending request: 
<request operation='SUBSCRIBE'>
<filter_list>
    <filter bbox="51.21366119 9.816023826 53.22928237 13.41331005 "/>
    <filter bbox="52.85487365 11.52271080 52.19587326 11.70257568 "/>
    <filter bbox="52.83801269 11.52468013 52.19419479 11.70454502 "/>
    <filter min_road_class="PRIMARY" bbox="51.91939544 11.25291442 52.36094284 11.97237205 "/>
    <filter min_road_class="PRIMARY" bbox="51.91770172 11.25488376 52.35926818 11.97434139 "/>
    <filter min_road_class="PRIMARY" bbox="52.13861083 11.61195945 52.14113616 11.61529636 "/>
</filter_list>
</request>
error:map_textfile:map_rect_new_textfile:error opening textfile /home/mobian/.navit/distortion.txt: Datei oder Verzeichnis nicht gefunden
error:traffic_traff_http:traffic_traff_http_process_response:TraFF request failed with status INVALID
```

christf:
in using the current state, the previous crashes and weirdness is gone but when navigating 600km, there are noticable ui freezes when fetching and parsing data, which warrants another look. Also thise "request invalid" responses are not yet understood by me.

mvglasow:
See https://gitlab.com/traffxml/traff/-/issues/26: `bbox` format is `minlat minlon maxlat maxlon`. If `minlat > maxlat`, that is an error and will cause the request to be rejected. (This might be unclear in the specs, might want to update that.) In the second and third entry of your example, `minlat` is greater than `maxlat`. You might want to fix that and also investigate why this is happening sometimes, but not all of the time, and whether it affects only latitude or also longitude. (`minlon > maxlon` is correct, but will give you the side of the globe which crosses the 180° meridian. This may go unnoticed, except that you will be receiving the wrong set of messages.)

christf:
I only saw your comment after summarizing here. I will change the implementation and have a look to identify why it is happening.

christf:
Today, I created a flame graph of navit and I checked which code is being executed in which thread.
This is the result


```
error:navit:event_main_loop_run:threadid on start 135330
error:traffic_traff_http:traffic_traff_http_on_feed_received:threadid before parsing message: 135330
error:traffic_traff_http:traffic_traff_http_worker_thread_main:threadid after releasing lock 135339`
```

This means: The function traffic_traff_http_on_feed_received including the message parsing happens in the main thread, not in the worker thread for the traff_http_plugin. This explains the gui freezes with the traff-http module enabled.
At the same time, I am not sure using an extra thread inside a plugin is the best approach. Maybe we can split the workload for parsing the individual messages into very small chunks and process these with idle_add_full() in the main thread? Also, the download itself could be built non-blocking. @mvglasow please feel free to join into these thoughts, I am not at all sure about this.


Also, this is a flame graph of navit using the traff-http plugin. The text to speech system piper is consuming significant resources and it may be useful to implement a cache for generated text.
Edit: better flamegraph with just navit:
flamegraph

mvglasow:
Bit of history here – Navit has been single-threaded all along, save for some dependencies possibly spawning their own threads. Naturally, this means extra care has to be taken to avoid locking up the UI.

In the traffic plugin, I followed the example of the routing engine: longer operations run in a loop, one work item at a time. Every couple of items, the loop is interrupted to allow UI input to be processed.

Admittedly, that is not perfect. Indeed, sometime around 2018/2019, I tried to implement multithreading in Navit, but eventually abandoned the project. If you want to tackle this, be warned that it will be a huge task. You’ll want to keep the UI separate from any major CPU eaters, and from anything which may block for an extended period of time. That comes down to some 2–3 components to keep track of:

* the UI (we want to keep it responsive)
* the routing engine (which can eat a lot of CPU power and lock up the UI)
* traffic (specifically message decoding, more specifically location decoding, which relies on the routing engine)

Unfortunately, all three need map access, so you will need a thread-safe way of accessing the map. The map drivers were not designed with thread safety in mind, so you will first need to figure out how they work and then come up with a way to make these operations thread-safe.

I eventually abandoned the project and salvaged some of the code for the `traff_http` plugin. Network requests should generally be moved off the UI thread, to prevent an unresponsive server from locking up the UI. (Android quite rigidly enforces that by killing any app which performs any network operation on the UI thread.) Moving just the network access off the main thread results in a consumer/producer model which is fairly easy to keep synchronized, especially since I wrote all of the data structures involved.

Most of my real-world traffic tests were done with TMC, where messages arrive one at a time, at a rate of less than ten messages per second, with frequent breaks where no new messages are delivered at all. My first attempts with online sources (using RoadEagle) led to some redesign here, but that may still not be perfect.

In conclusion, I would recommend sticking with the “cooperative multithreading” approach for now, maybe tweaking its parameters and/or breaking down work packages even further to allow for more frequent yields to the UI. Multithreading would for sure be a great thing, but making map access thread-safe might take a long time and I would not recommend making that a prerequisite for a new feature before it’s in place.

christf:
more on gui freezes: When it happens, navit logs this:

```
3:07:26.475|error:navit:traffic_message_add_segments:end point not found on map
23:07:26.508|error:navit:traffic_message_add_segments:end point not found on map
23:07:26.560|error:navit:traffic_message_add_segments:no segments
23:07:26.578|error:navit:traffic_message_add_segments:no segments
23:07:26.634|error:navit:traffic_message_add_segments:no segments
23:07:26.653|error:navit:traffic_message_add_segments:no segments
23:07:26.710|error:navit:traffic_message_add_segments:no segments
23:07:26.730|error:navit:traffic_message_add_segments:no segments

```

christf:
just saw this when selecting off-map position and destination:

```
19:56:11.121|error:traffic_traff_http:traffic_traff_http_worker_thread_main:sending request: 
<request operation='CHANGE' subscription_id='5337369e312527874a1e5e9f06b20291'>
<filter_list>
</filter_list>
</request>
19:56:11.238|error:traffic_traff_http:traffic_traff_http_process_response:TraFF request failed with status INVALID
```

I am not sure yet if it is a problem - capturing for now to determine if it is.

christf:
I have used this code + a few not yet committed changes during a few longer trips. It worked nicely, without gui freezes however the traffic information was not very current. One construction site must have moved on the morning of my drive (a few hourse before I passed by). It seems now I need to clean up the code a bit to make it presentable and then we need to start focussing on the completeness, accuracy and timeliness of the data.

christf:
for a few month now, this has been active in my mobile phone that I use regularly to navigate.

PR #1350, opened 2025-12-05
Feat:routing: Improve Time of Arrival Estimation developed by adevoss

christf:
The actual speed of a vehicle on a segment is lower than the maximum speed. On long routes, this difference can add up to significant deviations.

This implements an improved estimation that considers the vehicle speed to be less than the maximum possible.

I took the liberty of cleaning up whitespace and the commit history of #1319. As such, this supersedes #1319.

### DIFF
--- a/navit/route.c
+++ b/navit/route.c
@@ -99,6 +99,7 @@
 #include "mapset.h"
 #include "route_protected.h"
 #include "route.h"
+#include "navigation.h"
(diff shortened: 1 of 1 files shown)


adevoss:
> cleaning up ... commit history of #1319. As such, this supersedes #1319.
Duplicate of #1319.

Wiping commit history is not correct.

PR #1353, opened 2025-12-06
Docs:vehicle: Give example for activated, enabled demo vehicle

christf:
This should clarify usage of the demo vehicle

### DIFF
--- a/docs/user/configuration/basic/vehicle.rst
+++ b/docs/user/configuration/basic/vehicle.rst
@@ -14,7 +14,7 @@ A simple vehicle definition looks like this:
 
 Here some of the available options:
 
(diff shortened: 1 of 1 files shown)


christf:
Is there a way to see a rendered version of this docs change before merging the PR into trunk?

jkoan:
> Is there a way to see a rendered version of this docs change before merging the PR into trunk?

Yes, under Checks there's "docs/readthedocs.org:navit" when you click on it the preview doc will open.

christf:
Thank you for this, I completely had overlooked it. Now, I think the areas of my edit look fine.

jkoan:
We should probably add to this that this is a way if you want to always use debug on startup, which is valid, but you can also activate it from the menu.
When you set start and end and after that switch the vehicle in the settings it will also get activated.

christf:
nice, I did not know that. Indeed this should be added.

PR #1356, opened 2025-12-07
feat:maptool: write temporary files into instance-specific directory

### LINKED ISSUE #1328: maptool: secure temp files by randomizing names
currently no two instances of maptool can run inside the same directory as they overwrite each others temporary files. This is inconvenient and facilitates local users to change the result of the maptool process by manipulating those temporary files they know are being created.

christf:
This isolates the temporary files of different maptool instances. Once isolated, this allows running multiple instances of maptool inside the same directory at the same time. 

This fixes #1328

### DIFF
--- a/navit/maptool/tempfile.c
+++ b/navit/maptool/tempfile.c
@@ -20,11 +20,32 @@
 #ifndef _MSC_VER
 #include <unistd.h>
 #endif
+#include <sys/stat.h>
 #include "maptool.h"
 #include "debug.h"
 
+char *tempfile_obtain_prefix() {
+    static char *tmpfile_prefix = NULL;
+
+    if (!tmpfile_prefix) {
+#define tmpfile_prefix_size 64
+        tmpfile_prefix=calloc(tmpfile_prefix_size, 1);
+
+        snprintf(tmpfile_prefix, tmpfile_prefix_size, "maptool_%d.tmp", getpid());
+        if (mkdir(tmpfile_prefix, 0755)) {
+            perror("Error creating directory");
+            exit(1);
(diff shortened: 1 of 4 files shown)


jkoan:
This only isolates cords right? But there are other temp files as well. How about changing the function to create the temp files by including a directory which can be passed by a cli parameter, and if not a random one is generated

navit/maptool/tempfile.c
```diff
@@ -20,11 +20,32 @@
...
+        if (mkdir(tmpfile_prefix, 0755)) {
+            perror("Error creating directory");
+            exit(1);
+        }
+    }
+    return tmpfile_prefix;
+}
+
+void tempfile_cleanup() {
+    rmdir(tempfile_obtain_prefix());
```
christf, line 44: if the directory is not empty, it will remain on the file system. This is in line with the  current implementation that leaves debris in tmp-files in case errors occur.

christf:
> This only isolates cords right? But there are other temp files as well. How about changing the function to create the temp files by including a directory which can be passed by a cli parameter, and if not a random one is generated

no, this works for all temp-files including coords. See https://github.com/navit-gps/navit/pull/1356/files#diff-1a3eb2f5232b4d79962f2ce255c4b613e9d0cf0ad54f4c64e8a583eda4ae8e6bR48 for how the new function tempfile_obtain_prefix() is used for creating the names of every tmp-file.

christf:
manual testing revealed that search doesn't work anymore with this. Marking as WIP.

christf:
fixed with latest push.

PR #1364, opened 2025-12-15
Chore:command: Warning on String Truncation resolved

christf:
This shortens the amount of data to be considered by strncpy, avoiding the warning. There is no functional change, as the next line would have terminated eventually overflowing strings even before this - silencing the warning is still a good idea as it reduces false positives

### DIFF
--- a/navit/command.c
+++ b/navit/command.c
@@ -1472,7 +1472,7 @@ void command_evaluate(struct attr *attr, const char *expr) {
     }
     if (ctx.error && ctx.error != eof_reached) {
         char expr[32];
-        strncpy(expr, ctx.expr, 32);
+        strncpy(expr, ctx.expr, 31);
         expr[31]='\0';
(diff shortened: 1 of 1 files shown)


navit/command.c
```diff
@@ -1472,7 +1472,7 @@ void command_evaluate(struct attr *attr, const char *expr) {
     }
     if (ctx.error && ctx.error != eof_reached) {
         char expr[32];
-        strncpy(expr, ctx.expr, 32);
+        strncpy(expr, ctx.expr, 31);
         expr[31]='\0';
```
jkoan, line 1476: i know this is not on you, but shouldn't the null termination be on 32 i.e the last byte?

jkoan: ignore me, i haven't thought in my programmer brain

christf: it is a good question though - with these things one can never be too careful. Thank you for merging.

PR #1365, opened 2025-12-15
fix:traffic:Fix error on missing error message

jkoan:
If traffic.xml is missing there are two errors, which are not real errors


christf:
I agree with the change and it solves the problem of graciously skipping non-existent xml files. 
With this in mind, I wonder if the chick might be even better-suited to happen inside xml_parse_file()

PR #1371, opened 2025-12-16
Contributors updates and script speedup

### LINKED ISSUE #1368: Optimize Contributor Script
Make scripts/generate_contributors.sh faster, so it can run in seconds on CI.

POC:
```
git log  --encoding=utf-8 --full-history --date=short --use-mailmap "--format=format:%aN" #all
git log --after="2 years ago" --encoding=utf-8 --full-history --date=short --use-mailmap "--format=format:%aN" #active
```
subtract active from all to get inactive

christf:
closes #1368

### DIFF
--- a/scripts/generate_contributors.sh
+++ b/scripts/generate_contributors.sh
@@ -9,44 +9,11 @@ set -e
 # Note: it uses git's mailmap functionnality to get a clean list of users
 # ###########################################################################################################
 
-declare -A CONTRIBUTORS=()
-declare -a authors=()
-TWO_YEARS_AGO=$(date +%s --date="2 years ago")
-retiredTitleWritten=false
+recentcontributors=$(git log --after="2 years ago" --encoding=utf-8 --full-history --date=short --use-mailmap "--format=format:%aN"  | sort  -fu)
+allcontributors=$(git log  --encoding=utf-8 --full-history --date=short --use-mailmap "--format=format:%aN"|sort -fu)
 
-git log --encoding=utf-8 --full-history --date=short --use-mailmap "--format=format:%ad;%aN" |
-{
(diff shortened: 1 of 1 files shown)


AUTHORS
```diff
@@ -1,62 +1,83 @@
...
+Ettore Atalan
+Helder Santana
+Hosted Weblate
 jkoan
-Johan Fitié
-Joseph Herlant
 lains
-Lionel AINS
 metalstrolch
-mvglasow
```
christf, line 16: do we really want this?

jkoan: In the current state yes, but we should maybe look at other types of contributions like issue management, translations, ... 
But what we won't be able to do is to manually add or manage a list of contributions

scripts/generate_contributors.sh
```diff
@@ -9,44 +9,11 @@ set -e
...
-  done
-  # We are still in the subprocess scope so we can print the sorted list of
-  # retired contributors
-  IFS=$'\n' sorted=($(sort <<<"${authors[*]}"))
-  printf "%s\n" "${sorted[@]}" >> AUTHORS
-}
+echo "# Active contributors:" > AUTHORS
+echo "$recentcontributors" >> AUTHORS
+echo -e "\n# Retired contributors:" >> AUTHORS
+echo -e "$recentcontributors\n$allcontributors" |sort | uniq -u >> AUTHORS
```
jkoan, line 19: With `uniq -u` I even learned something new today. I was first struggling to understand why it was correctly working, because I only used `sort -u` in the past. And with that we would have had duplicates between the two sections

christf: to me, sort -f was new today. :)

PR #1370, opened 2025-12-16
Chore:i18n: break build on outdated msgid on ci_sanity_checks

christf:
For good coverage of translations, navit must have up-to-date po.in files. This breaks the ci sanity check, in case there are new or changed msgids as part of a compile process, indicating that po files have not properly updated.

If this occurs on, say, file de.po, then copy de.po from the build/po/de.po location to po/de.po.in and commit, to resolve.

closes #1337 

- [x] this depends on https://github.com/navit-gps/Dockerfiles/pull/15 
- [x] execute this sanity check as part of gha
- [x] ensure cmake and the other build dependencies are installed in the image running this script

PR #1372, opened 2025-12-17
Feat:i18n: Update PO files to current strings

christf:
This updates all po files to the current state of the code base. The new ci sanity check will help us ensure they are kept in sync for the future.

This is just the pot update from #1370 so others can start translating while the check is being put in place.


jkoan:
Weblate will actually do this automatically when the pot is updated, so we only need to update this on our end

christf:
sorry, I do not understand the statement. What do you mean by "this" (both occurances)?

christf:
seeing that the structural topics were closed in #1370 I am closing this PR.

PR #1377, opened 2025-12-18
Set gdb-friendly compiler options when debugging

christf:
when working with gdb, it is extremely helpful if information about line numbers has been baked into the binary. Also, reducing optimizations so the progress is in line with line numbers of code helps for clarity. Complete backtraces help too.


jkoan:
CMake has https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html and also https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_CONFIG.html

PR #1378, opened 2025-12-18
Whitespace Changes

christf:
this formats the code currently being worked on in a few other PRs to have a clean foundation for discussion.


christf:
rebasing this on top of #1382 because this depends on a fix of file.h that was made there.

Actually, no, let's do this the other way around. This now contains all the whitespace changes which should not be controversial. Will rebase #1382 which contains a refactoring and user-visible change in behavior ontop of this PR.

edit: done.

navit/attr_def.h
```diff
@@ -30,26 +30,26 @@
 
 /* common */
 #ifndef ATTR
-#define ATTR(x) extern ##x;
+#define ATTR(x) extern##x;
 #endif
 #ifndef ATTR2
-#define ATTR2(x,y) extern ##y;
+#define ATTR2(x, y) extern##y;
```
jkoan: this is interesting as i needed to research this, and it zthurn out, both expand to externX for ATTR(X) without scpace. I think from a formating perspective this is correct, but i am not sure if it should expand to `extern x`

navit/command.c
```diff
@@ -292,52 +305,56 @@ static int command_object_remove_attr(struct context *ctx, struct attr *object,
...
 static void command_get_attr(struct context *ctx, struct result *res) {
     int result;
-    struct result tmp= {{0,},};
-    enum attr_type attr_type=command_attr_type(res);
+    struct result tmp = {
+        {
+            0,
+        },
+    };
+    enum attr_type attr_type = command_attr_type(res);
```
jkoan: Thats a extreme case where i would probably would like a middle option like 
```
{
    {0,},
}
```
But i am fine with this as well

navit/item.h
```diff
@@ -46,33 +45,31 @@ extern int default_flags[];
...
-                               ||  (item).type == type_living_street  \
-                               ||  (item).type == type_street_construction  \
-                               ||  (item).type == type_path \
-                               ||  (item).type == type_street_parking_lane \
-                               ||  (item).type == type_footway )
+#define item_is_street(item)                                                                                           \
+    (((item).type >= type_street_nopass && (item).type <= type_roundabout) || (item).type == type_street_service ||    \
+     ((item).type >= type_street_pedestrian && (item).type <= type_track_grass) ||                                     \
+     (item).type == type_living_street || (item).type == type_street_construction || (item).type == type_path ||       \
+     (item).type == type_street_parking_lane || (item).type == type_footway)
```
jkoan: i think this one was more readable before, any chance to have clang-format leave this one untouched?

navit/item.h
```diff
@@ -46,33 +45,31 @@ extern int default_flags[];
...
+    (((item).type >= type_poi_lake && (item).type <= type_poi) ||                                                      \
+     ((item).type >= type_poi_land_feature && (item).type <= type_poi_zoo) ||                                          \
+     ((item).type >= type_poi_gc_multi && (item).type <= type_poi_cafe) ||                                             \
+     ((item).type >= type_poi_peak && (item).type <= type_poi_image) ||                                                \
+     ((item).type >= type_poi_townhall && (item).type <= type_poi_ruins) ||                                            \
+     ((item).type >= type_poi_post_box && (item).type <= type_poi_tennis) ||                                           \
+     ((item).type >= type_poi_vending_machine && (item).type <= type_poi_shop_shoes) ||                                \
+     ((item).type >= type_poi_tree && (item).type <= type_poi_shop_photo) ||                                           \
+     ((item).type >= type_poi_pub && (item).type <= type_poi_bahai) ||                                                 \
+     ((item).type >= type_poi_customh && (item).type <= type_poi_customf))
```
jkoan: this one is fine, if we can do it like this for item_is_street this would be great

navit/route.c
```diff
@@ -3965,131 +3948,111 @@ static struct item *rp_get_item(struct map_rect_priv *mr) {
...
-    "utf-8",
-    rm_destroy,
-    rm_rect_new,
-    rm_rect_destroy,
-    rm_get_item,
-    rm_get_item_byid,
-    NULL,
-    NULL,
-    NULL,
+    projection_mg, "utf-8", rm_destroy, rm_rect_new, rm_rect_destroy, rm_get_item, rm_get_item_byid, NULL, NULL, NULL,
```
jkoan: this and route_graph_meth surprises me, why is this in one line again?

navit/main.c
```diff
@@ -89,355 +88,353 @@ void setenv(char *var, char *val, int overwrite) {
...
-    {"NAVIT_LIBPREFIX",   "*/",   NULL,            NULL,          NULL,           NULL},
-    {NULL,                NULL,         NULL,            NULL,          NULL,           NULL},
+static char *environment_vars[][6] = {
+    {"NAVIT_LIBDIR", ":", ":/" LIB_DIR, ":\\lib", ":/lib", ":\\lib"},
+    {"NAVIT_SHAREDIR", ":", ":/" SHARE_DIR, ":", ":/share", ":"},
+    {"NAVIT_LOCALEDIR", ":/../locale", ":/" LOCALE_DIR, ":\\locale", ":/locale", ":\\locale"},
+    {"NAVIT_USER_DATADIR", ":", "~/" HOMECONFIG_DIR, ":\\data", ":/home", "~\\navit"},
+    {"NAVIT_LOGFILE", NULL, NULL, ":\\navit.log", NULL, ":\\navit.log"},
+    {"NAVIT_LIBPREFIX", "*/", NULL, NULL, NULL, NULL},
+    {NULL, NULL, NULL, NULL, NULL, NULL},
```
jkoan: in this case i also think that the old had more readability

navit/main.c
```diff
@@ -89,355 +88,353 @@ void setenv(char *var, char *val, int overwrite) {
...
+#if defined(HAVE_API_WIN32_CE)
+#warning HAVE_API_WIN32_CE is defined
+#endif
+#if defined(HAVE_API_WIN32)
+#warning HAVE_API_WIN32 is defined
+#endif
+#error Exactly only one directive amongst HAVE_API_WIN32_CE or HAVE_API_WIN32 should be defined when preprocessor reach this section of code
+#endif
+#endif
+#endif /* _WIN32 || _WIN32_WCE */
```
jkoan, line 434: i don't really know if i like the old style, but it's clearer which if belongs to what in that case. But fine with the new one, only suggestion to add missing comments on the endif's

navit/item.h
christf: Indeed, I like the previous version for this segment more as well. I have played around with it a bit, but there does not seem to be an approach that is local to this area. In a fixup commit I pushed a suggestion. I like it a bit better but this area remains tricky because clang-format tries to use the whole line width. Changing the sorting helps a bit but it still is not great. If you like it, the next rebase will automatically squash it. If you don't, it is just as simple to remove this commit again.

navit/main.c
christf: the problem with the old style is that it breaks the logic depending on the toolchain. Apparently the piece of cursed knowledge (preprocessor directives cannot be indentend and compilers will break if you try) now is not much of a thing any more since clang-format has a directive for formatting preprocessor directives. I was hunting bugs involving indentation with a version of proc from the early 2000s last year. I will re-add the indentation in another fixup commit as our tooling is too young to legally drink and therefore it should be fine.

navit/main.c
christf: agreed. fixup commit pushed.

navit/route.c
christf: I agree. There is https://github.com/llvm/llvm-project/issues/96664 for that situation and clang-format seems unable to control this aspect. This is a bit annoying.

navit/item.h
christf: this one is not uglified because the && binds stronger than || and the clauses are sufficiently long...

navit/main.c
jkoan: its much better now, but do you know why its now right aligned instead of left aligned?

navit/route.c
jkoan: maybe we can try with struct dot attributes https://stackoverflow.com/questions/8047261/what-does-dot-mean-in-a-struct-initializer
But this would require c99 and currently we haven't set any particular c style

navit/command.c
jkoan: for me it looks worth now:

```
struct result tmp = {
    {
        0, },
    };
```

christf: yeah, this still needs some tlc

navit/main.c
christf: That is because the AlignArrayOfStructures Parameter is set to Right. I do not mind if it is left, but leaned towards right because it might look better if numbers with decimals are involved. Do you want it left-aligned?

navit/command.c
christf: So, I dug around in the c99 standard and learnt that this bracketed initialization is already C99. Also, iiuc, it is equivalent to ```
struct result tmp = { 0 }; ``` which looks a bit cleaner if formatted like in the latest push. Shall we go with that?

navit/route.c
christf: since c99 is used in other places, would it be a problem if we use it here too? Pushed another bit that looks pleasing to my eyes in that area

christf:
it seems to me, the decisions left are:
* shall we go with left or right alignment?
* is C99 ok?
* is the bit in item.h acceptable?

navit/route.c
jkoan: no its not a problem at all. My hint to c99 was more that we don't enforce a specific c standard yet, which we should do in the future

navit/main.c
jkoan: I tested this locally and like Left more

navit/command.c
jkoan: yes, works for me

jkoan:
> shall we go with left or right alignment?

I tested locally and like Left more

> is C99 ok?

Yes

> is the bit in item.h acceptable?

Yes

jkoan:
For 0e04c29536470deebc99e6121deebcd1bc361180 you Say _Without it, the compiler complains about missing types._ so we have a todo there?

christf:
> For [0e04c29](https://github.com/navit-gps/navit/commit/0e04c29536470deebc99e6121deebcd1bc361180) you Say _Without it, the compiler complains about missing types._ so we have a todo there?

no, because this pr contains the commit 0e04c29536470deebc99e6121deebcd1bc361180

jkoan:
> > For [0e04c29](https://github.com/navit-gps/navit/commit/0e04c29536470deebc99e6121deebcd1bc361180) you Say _Without it, the compiler complains about missing types._ so we have a todo there?
> 
> no, because this pr contains the commit [0e04c29](https://github.com/navit-gps/navit/commit/0e04c29536470deebc99e6121deebcd1bc361180)

ah i understand now *it* refferes to this commit, not the removed header

PR #1379, opened 2025-12-18
Fail sanity check on errors

christf:
if the sanity check fails, it is helpful if the pipeline execution task
is also marked with a red icon. This returns a non-null exit code in
case the sanity check fails to communicate this failure as part of the
CI run


christf:
evidence of the check working: https://github.com/navit-gps/navit/actions/runs/20354846550/job/58487911470?pr=1379

scripts/ci_sanity_checks.sh
```diff
@@ -86,8 +86,10 @@ for f in $(git diff --name-only refs/remotes/origin/trunk | sort -u); do
             rc=$?
             if [[ $rc -ne 0 ]]; then
                 echo "[ERROR] Your ${f} file doesn't validate against the navit/navit.dtd using xmllint" >&2
+								exit 3
```
christf: fix whitespace

jkoan:
This is outdated right? As #1388 is merged which includes this as well

PR #987, opened 2020-04-06
[WIP] Update:Android:GUI/Internal: Show contextual GUI menu when receiving geo intent

lains:
I my previous PR #812, the default action when receiving a geo intent (for example by scanning a geo QR code), is to start navigating to the provided location.
This PR aims to give the user more control over which action is started when navit receives a geo intent.
For this (and only implemented when using the internal GUI), I bring up the contextual submenu for a geo position in the internal GUI, and let the user select what to do with this position (add bookmark, set as waypoint, destination, search for nearby POIs), in the same way as when the geo pos is entered manually in the internal GUI.

This is only for Android, and if this feature is not implemented in the active GUI plugin (only implemented in internal GUI for now), the behaviour will revert back to set destination to the GPS coords.

PR #1381, opened 2025-12-20
rebasing the work by lains from #987

christf:
this contains lains work from #987 and rebases it on top of latest trunk. Minor cleanups were already applied, as well as fixing the build for linux. No testing was done.


jkoan:
I have tested that this works well

Issue #1380, opened 2025-12-19
more on ETA


christf:
current implementation in route.c always uses the route weight to obtain the vehicle max speed. This is not in line with the current documentation that says, this value is exclusively used if maxspeed_handling=2.
While this could have an easy fix (in docs), I do not think that this is the proper way to go.

Vehicles can move at certain speeds in certain environments. So each environment needs to be able to specify "a bike can go 25km/h on asphalt", "a bike can go "3km/h on stairs"  or "a car can go 80km/h on highways". There already is a speed parameter and a maxspeed parameter in the roadprofile.

maxspeed sets the maximum allowed speed on that road type independent if vehicle capabilities in absence of OSM provided speed limit. It is redundant to specify it in each vehicleoption, but that is what we currently do. 

speed sets the value to be used for eta calculation. If a road has a government issued speed limit of 50km/h then the maxspeed should be set to 50 (either per default or map data).
For cars, the speed parameter should be lower as argued by @adevoss in #1319 maybe to 45. There should not be a need for factors in obtaining good values because the data can be directly controlled. 

If a pedestrian cannot go faster than 4km/h, great, set the speed attributes of all road types for them to that value. maybe consider direction for stairs later (but I am getting ahead of myself).

In addition there should be a likelihood of selecting a path. This is done by the weight as per docs. This needs to be kept separate from eta calculation because, well, it is a different thing. Maybe going slow is great for EV on  that preferred road because using it allows to save energy.

So, here is my proposal:

1. remove maxspeed_handling (0,1,2) setting from the vehicleprofile entirely.
2. change the implementation to obtain the vehicle speed from the minimum of these 2 values: vehicleprofile->roadprofile->speed and the maxspeed (as obtained from default for this roadprofile or osm data as described above) parameter. use that to calculate eta.

On top of all of this: route_weight is currently unused except for the incorrect usage as default speed. A route weight is unitless, a speed has a unit of distance / time. We could remove that too to avoid further confusion.

christf:
more ideas: 

* include terrain in the eta calculation... going downhill by bike is significantly faster than uphill... maybe that is yet another issue.
* consider turning penalty. A change in direction comes at the cost if speed. A route with many turns will be slower than an equal length path without turns
* consider a penalty for traffic lights

adevoss:
> current implementation in route.c always uses the route weight to obtain the vehicle max speed.

Agree that this should be changed. See #1320.


[...]

PR #1382, opened 2025-12-20
Illustrate thoughts on ETA from #1380

christf:
This outlines the thoughts in  #1380

- [x] clean up docs
- [x] clarify route_weight usage /remove it from docs and code
- [x] fix typos from commit messages
- [x] fix whitespace

### DIFF
--- a/navit/attr_def.h
+++ b/navit/attr_def.h
@@ -126,7 +126,6 @@ ATTR(pitch)
 ATTR_UNUSED
 ATTR_UNUSED
 ATTR(route_status)
-ATTR(route_weight)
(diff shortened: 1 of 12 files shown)


navit/attr_def.h
```diff
@@ -126,10 +126,8 @@ ATTR(pitch)
 ATTR_UNUSED
 ATTR_UNUSED
 ATTR(route_status)
-ATTR(route_weight)
```
jkoan, line 129: Don't just remove it, this throws of the indexes and they are used for example in maps which then makes the map files incompatible. Replace with ATTR_UNUSED

PR #1384, opened 2025-12-21
Use speed instead of route_weight for vehicle speed

adevoss:
Use speed instead of route_weight for vehicle speed


adevoss:
Didn't see https://github.com/navit-gps/navit/issues/1382 before creating this PR.
This PR is already incorporated in https://github.com/navit-gps/navit/issues/1382

Issue #1373, opened 2025-12-17
Container image for maptool


jkoan:
We should provide a container image for at least maptool. It could for example be used in https://github.com/navit-gps/gh-actions-mapserver/ or by users which want to self build maps.

PR #1387, opened 2025-12-23
Fix:Build: Linux libnavit_core included in build

### LINKED ISSUE #1386: Build:Linux properly link to libnavit_core
this library is currently not linked. As such, the linux build is broken and needs to be fixed.

christf:
This installs libnavit_core and therefore fixes #1386. However, libnavit is only found when `export LD_LIBRARY_PATH=/usr/local/lib64/navit/`

I am not sure if this is intentional or not.


jkoan:
No it's not intended to need an LD_LIBRARY_PATH. 
In the last release the lib wasn't actually there and both navit and maptool are about the size of the libnavit bigger, so we probably need to add something else or move the libnavit one directory up so it's in the distributions lib dir. But this doesn't feel right when we already have a navit directory.
So we probably have to set rpath to something like ${ORIGIN}/../lib64 where lib64 my be lib as well depending on target.

See here https://discourse.cmake.org/t/help-with-rpath-and-install-prefix/9095 and https://jeremimucha.com/2021/04/cmake-fundamentals-part8/

So this should help:

```
file(RELATIVE_PATH relativeRpath
    ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_BINDIR}
    ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}
)
set(CMAKE_INSTALL_RPATH $ORIGIN $ORIGIN/${relativeRpath})
```

jkoan:
I tested this and for some reason the rpath just contained a wrong path followed by a bunch of colons. 
In the root CMakeLists.txt we already determine the LIB_DIR maybe we can put the rpath variable into the existing if statements? This way it would also be logically coupled. 
If the rpath variable needs to be in navit/CMakeLists.txt you could add a custom variable. But I think this should not be necessary

christf:
did you test the pr or your suggestion from the previous comment? seeing an invalid path with many ../ was, what I got from applying the comment which was my starting point. The docs you mentioned are a good read, thank you for that :)

While there is still something not quite clear to me (the reason for the weird rpath), the PR as is, yields a working installation for me on linux with a navit binary that runs without having to specify LD_LIBRARY_PATH

I am happy to improve it though and agree that this magic (whatever ends up being required - we might want versioning too) should be done on the top level.

jkoan:
After cpack or install the rpath will be set (which i haven't done before), still the path is not optimal
```
an@jan-pc:/tmp/navit-g$ readelf -d ./navit/bin/navit | grep -i runpa
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/../../../../../../usr/local/lib64/navit]
```
With this patch [rpath.patch](https://github.com/user-attachments/files/24336957/rpath.patch) the path is as exacted:
```
jan@jan-pc:/tmp/navit-g$ readelf -d ./navit/bin/navit | grep -i runpa
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:'$ORIGIN/../lib64']
```

jkoan:
> After cpack or install the rpath will be set (which i haven't done before), still the path is not optimal
> ```
> an@jan-pc:/tmp/navit-g$ readelf -d ./navit/bin/navit | grep -i runpa
>  0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/../../../../../../usr/local/lib64/navit]
> ```
> With this patch [rpath.patch](https://github.com/user-attachments/files/24336957/rpath.patch) the path is as exacted:
> ```
> jan@jan-pc:/tmp/navit-g$ readelf -d ./navit/bin/navit | grep -i runpa
>  0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:'$ORIGIN/../lib64']
> ```


I think there's still a slight error on the quoting, I think the ' needs to go

christf:
with the patch, after make install the path revealed by ldd is fine. However after make, before make install ldd on build/navit/navit links to /usr/local/lib64 instead of the in-tree libnavit_core.

This could challenge the usefuless of local builds. Maybe we can still do even better.

jkoan:
could you please recheck with a clean build and `readelf -d build/navit/navit | grep -i runpa`? For me the build dir rpath (runpath) is [...]/build/navitwhich is correct.
Maybe you still had `LD_LIBRARY_PATH` set? Because modern compilers use runpath instead of rpath the LD_LIBRARY_PATH Env Variable may overwrite the runpath

christf:
I did re-check and I have to say, I feel embarrassed. A few days back I manually set LD_LIBRARY_PATH and by a fluke ran this test in that particular shell. So, indeed the patch is working as expected. Because, I pushed it, it bears my name as author - feel free to got commit --reset-author if you like. It is your patch after all.

jkoan:
Yesterday I was working on #1373 and noticed that got my patch wrong, because I built a deb packageb with cpack.
We don't actually need the runpath to contain $ORIGIN/../lib64 but $ORIGIN/../lib64/navit which was an error on my part. 
So the patch is even simpler because we don't need to set it in ever LIB_DIR if case individually but can use $ORIGIN/../${LIB_DIR} 
Only for the last if case where LIBDIR gets set I am not sure what the rpath should be.

jkoan:
Its now working. Tested with the WIP work on #1373 where I generate a deb and install it. @christf any objections?

christf:
on the contrary. I can confirm this works for me.

PR #1389, opened 2025-12-27
Chore:Dockerfile: Add Dockerfile for maptool Container

jkoan:
This is a minimal patchset to build maptool into a container image (i used Docker for my testing). it depends on #1387 

Its working by generating a deb package which is then copied from the build container into the runtime container.
The build is done like so: `docker build -f contrib/container/maptool/Dockerfile --tag maptool .` It needs to be run from the root of the repo.

What we should do in the next PR's is :
- Github Actions build and push to registry :heavy_check_mark: 
- Extend the cpack deb instructions so that we can build navit.deb and maptool.deb in two packages, so this can be used for nightly repos
- use the container image at https://github.com/navit-gps/gh-actions-mapserver/

Usage looks like this:
```
jan@jan-pc:~/git/navit$ docker run -ti --rm maptool -h

maptool - parse osm textfile and convert to Navit binfile format

Usage (for OSM XML data):
bzcat planet.osm.bz2 | maptool mymap.bin
Usage (for OSM Protobuf/PBF data):
maptool --protobuf -i planet.osm.pbf planet.bin
Available switches:
-h (--help)                       : this screen
-3 (--32bit)                      : set zip 32 bit compression
-6 (--64bit)                      : set zip 64 bit compression (default)
-a (--attr-debug-level)  <level>  : control which data is included in the debug attribute
-c (--dump-coordinates)           : dump coordinates after phase 1
-D (--dump)                       : dump map data to standard output in Navit textfile format
-e (--end) <phase>                : end at specified phase
-E (--experimental)               : Enable experimental features (Move coastline data to order 6 tiles. Makes map look more smooth, but may affect drawing/searching performance.)
-i (--input-file) <file>          : specify the input file name (OSM), overrules default stdin
-k (--keep-tmpfiles)              : do not delete tmp files after processing. useful to reuse them
-M (--o5m)                        : input data is in o5m format
-n (--ignore-unknown)             : do not output ways and nodes with unknown type
-N (--nodes-only)                 : process only nodes
-P (--protobuf)                   : input data is in pbf (Protocol Buffer) format
-r (--rule-file) <file>           : read mapping rules from specified file
-s (--start) <phase>              : start at specified phase
-S (--slice-size) <size>          : limit memory to use for some large internal buffers, in bytes. Default is 1GB.
-t (--timestamp) <y-m-dTh:m:s>    : Set zip timestamp
-T (--threads) <count>            : Set number of threads (for some operations)
-w (--dedupe-ways)                : ensure no duplicate ways or nodes. useful when using several input files
-W (--ways-only)                  : process only ways
-U (--unknown-country)            : add objects with unknown country to index
-x (--index-size)                 : set maximum country index size in bytes
-z (--compression-level) <level>  : set the compression level
Internal options (undocumented):
-b (--binfile)
-B 
-m (--map) 
-O 
-p (--plugin) 
-u (--url) 
```

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,0 +1,32 @@
+name: Build and Push Docker Image
+
+on:
+    push:
+        branches:
+            - trunk
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v6
+
+            - name: Set up Docker Buildx
+              uses: docker/setup-buildx-action@v3
+
+            - name: Log in to GitHub Container Registry
+              uses: docker/login-action@v3
+              with:
+                registry: ghcr.io
+                username: ${{ github.actor }}
+                password: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Build and push Docker image
+              uses: docker/build-push-action@v6
+              with:
+                context: .
+                file: contrib/container/maptool/Dockerfile
+                push: true
+                tags: ghcr.io/${{ github.repository }}/maptool:trunk


CMakeLists.txt
```diff
@@ -3,6 +3,10 @@ cmake_minimum_required(VERSION 3.18)
 set(MACOSX_BUNDLE_GUI_IDENTIFIER "org.navitproject.navit")
 set(CMAKE_EXECUTABLE_ENABLE_EXPORTS "true")
 set(MACOSX_BUNDLE_BUNDLE_NAME "Navit")
+set(CPACK_PACKAGE_CONTACT "navit@navit-project.org")
+set(CPACK_DEB_COMPONENT_INSTALL ON)
+#set(CPACK_DEBIAN_MAPTOOL_PACKAGE_DEPENDS "")
```
hoehnp, line 8: cant we just omit this line, since it is anyway commented out.

contrib/container/maptool/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# Copy the current repo into the container
+COPY . .
+
+# Install dependencies
+RUN ./scripts/setup_common_requirements.sh && apt-get install -y libpng-dev libfreetype6-dev libdbus-glib-1-dev libgtk2.0-dev curl
+
+# Build the project
+RUN mkdir build && cd build && \
+    cmake -Dsvg2png_scaling=-1 -Dsvg2png_scaling_flag=-1 -Dsvg2png_scaling_nav=-1 -Dsvg2png_scaling_textures=-1 .. && \
+    make maptool&& \
```
christf, line 17: can we set -O and  -j 4 or even -j 8 here?

contrib/container/maptool/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    cpack -G DEB
+
+# Runtime container
+FROM debian:stable-slim
```
christf, line 21: personally, I subscribe to reproducible builds, which would require to set a specific version here. In absence of renovate / dependabot, I think this is fine for now as is, and maybe we should consider using bots and in the future shifting over to more specific tags to help with reproducability.

CMakeLists.txt
jkoan: Yes, we could, and I actually wanted to, but forgot to remove it. But it's in my opinion not really important as we should work on proper cpack deb support anyway. This implements only the bare minimum required to get a usable den for maptool

contrib/container/maptool/Dockerfile
jkoan: What do you mean by -O? Do you mean compiler optimisations like -O2?
For -j yes good point, I will use -j$(nproc)

contrib/container/maptool/Dockerfile
jkoan: Definitely a good point, but as you also concluded, currently it's not really feasible. But I agree it would be a good thing to work on in the future. But when speaking of reproducible builds I think we have more homework to do ;-)

contrib/container/maptool/Dockerfile
jkoan: I looked it up before, but haven't found anything. Looking further I found it. It's an output sync. This sounds helpful. But since when does this exist? It seems to not be really well known

christf: no, I actually meant the make flag -O for output-sync as you now have found :)
It has existed for a long time, I think decades.

I like to use it when compiling with multiple jobs because it helps to understand better what is broken. However it may mean that an error message might be significantly earlier in the scrollback buffer. I think it has existed for a long time. It is not critical though. Let's do -j <something great> for now.

CMakeLists.txt
christf: just created an issue for this.

christf (approved):

christf:
Is anything keeping us from merging this?

PR #1400, opened 2026-01-17
Feat:Navit: Set C Standard to C99

### LINKED ISSUE #1398: Set C Standart the Project uses
This will make it clearer which features we can use. We should take a look at what our minimum common denominator is at the moment and set this in cmake

christf:
this fixes #1398

I am not sure if C99 is correct. If we go to 11, we get additional warnings.

### DIFF
--- a/navit/support/glib/gutils.c
+++ b/navit/support/glib/gutils.c
@@ -88,8 +88,7 @@
 #if NOT_NEEDED_FOR_NAVIT
 #  include <direct.h>
 #endif /* NOT_NEEDED_FOR_NAVIT */
-#  include <windows.h>
-typedef void IShellExecuteHookA;
+struct IShellExecuteHookA;
 #  include <shlobj.h>
    /* older SDK (e.g. msvc 5.0) does not have these*/
 #  ifndef CSIDL_MYMUSIC


jkoan:
This only sets target navit. We also have maptool which should be added, and we need to look if the lib's need to be added as well. Also https://cmake.org/cmake/help/v3.18/prop_tgt/C_STANDARD_REQUIRED.html#prop_tgt:C_STANDARD_REQUIRED should probably be set

jkoan:
while looking at the last commits from iwyu i found this, maybe this is also what we need: https://github.com/include-what-you-use/include-what-you-use/commit/d873c86084741a38ed331d22971eced665268282

christf:
oh. interesting. I found that iwyu seems to work for our code base as is. When problems occur, then includes are missing entirely and the tool cannot conjure them up. So, until now I have not hit an obvious blocker in using iwyu but I guess ymmv and now I am not sure if we should proceed with c99 or something newer.

jkoan:
That's not what i meant, i just meant that iwyu is also a c Project which has introduced the same feature like we want to. I didn't mean to say that something about the usage of iwyu is not working

navit/CMakeLists.txt
```diff
@@ -216,6 +216,7 @@ if (USE_LIBGNUINTL AND NOT HAVE_GLIB)
 	ADD_DEPENDENCIES(support_glib support_gettext_intl)
 endif()
 
+set_property(TARGET navit PROPERTY C_STANDARD 99)
```
jkoan: This won't work for all compilation units. See my last comment in the thread. We need to do it like it was done here:
https://github.com/include-what-you-use/include-what-you-use/commit/d873c86084741a38ed331d22971eced665268282

christf: I would think that the build failing now for maptool is a symptom of the setting being in effect ;)

christf:
I do not think we can do this at the moment. Let me close it for now.

PR #1401, opened 2026-01-18
Feat:ci_sanity_check: po-Update simplified

christf:
By copying the newly-built po files into the source tree, this simplifies the assessment to be done by humans which po files might have changed for which reason. This simplifies the workflow as git commit -a -m can later be used to accept all the changes

### DIFF
--- a/scripts/ci_sanity_checks.sh
+++ b/scripts/ci_sanity_checks.sh
@@ -13,9 +13,10 @@ set -u
 
 return_code=0
 GIT_DIFF_OUTPUT="${1:-/dev/stdout}"
+KEEPTEMP=${2:-false}
 
 cleanup() {
-	rm -rf build-locale-$$
+	$KEEPTEMP || rm -rf build-locale
 }
 trap cleanup EXIT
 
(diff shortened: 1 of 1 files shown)


jkoan:
One idea while you're at it. Just upload the patch as an artifact to the ci so the user can just download, review and apply it

christf:
I am not sure about that. Currently it is possible to locally execute the sanity check. If it would not take 22s to run, I would do it as commit hook. I have played around with it and can optimize to 5s given caching.
This would then place the changes directly.

I think, it could be more desirable to keep the option of running locally instead of binding the shell script to gha.
So, maybe that is another script that collects the data and presents it? Maybe changing the error message "do run ci sanity check locally to obtain the patch" could do the job too? It certainly would be faster than downloading a patch and manually applying it...

jkoan:
> I am not sure about that. Currently it is possible to locally execute the sanity check. If it would not take 22s to run, I would do it as commit hook. I have played around with it and can optimize to 5s given caching. This would then place the changes directly.
> 
> I think, it could be more desirable to keep the option of running locally instead of binding the shell script to gha. So, maybe that is another script that collects the data and presents it? Maybe changing the error message "do run ci sanity check locally to obtain the patch" could do the job too? It certainly would be faster than downloading a patch and manually applying it...

My idea was to just add a parameter to the script to enable the patch generation to a desired path. Then upload that patch as artifact from the workflow. This way you can still run the script locally without issues.
Quick POC (untested): [ci_sanity_checks.patch](https://github.com/user-attachments/files/24708181/ci_sanity_checks.patch)

christf:
applied with a minor change

PR #1407, opened 2026-01-24
Optimize docs

jkoan:
This PR Fixes the pdf output and loads of formatting

### DIFF
--- a/docs/development/targets/Ångström_development.rst
+++ b/docs/development/targets/Ångström_development.rst
@@ -115,11 +115,11 @@ bitbake
 | bitbake 1.8.12 didn't work as well as 1.8.10.
 | ====Libs==== For my build I had to change the plugins block in
   navit.xml to make the libs found:
-| 
+|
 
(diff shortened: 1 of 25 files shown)


docs/development/targets/Ångström_development.rst
```diff
@@ -115,11 +115,11 @@ bitbake
 | bitbake 1.8.12 didn't work as well as 1.8.10.
 | ====Libs==== For my build I had to change the plugins block in
   navit.xml to make the libs found:
-| 
+|
```
christf: I think this is no longer necessary since the linker changes we did in november/december last year, is it? Should this be clarified now?

jkoan: I think Ångström is defunct anyway. Don't know if anyone still uses it any way. So i don't know if this is even the same issue.
It's just there as reference if relevant in the future. That also why its not listed anywhere on https://navit.readthedocs.io/en/trunk/user/platforms/index.html

PR #1420, opened 2026-01-31
Chore: Switch from pocketpc-cab to gcab

christf:
Caveat: I do not have pocketpc to try any of this.


christf:
it would be nice to not have this cab inside the zip but present it as a separate artefact,

PR #1413, opened 2026-01-27
Whitespace

christf:
This should hopefully mark the remaining bit of the formatting changes.

PR #1423, opened 2026-02-01
Chore: svg_debug: sprintf replaced with snprintf to make Codefactor happy

christf:


christf:
this is now part of #1413

PR #1426, opened 2026-02-01
chore(build): process most c and h files with iwyu

christf:
- [ ] apply iwyu to plugin.c (do this in #1439 )
- [x] apply iwyu to all header files (done but no changes required, interestingly)
- [x] fix windows build
- [x] drop the commits that are intended to be dropped

### DIFF
--- a/navit/search_houseno_interpol.c
+++ b/navit/search_houseno_interpol.c
@@ -18,21 +18,16 @@
  */
 
 #include "search_houseno_interpol.h"
+#include "attr.h"
+#include "attr_type_def.h"
 #include "coord.h"
-#include "country.h"
 #include "debug.h"
-#include "geom.h"
 #include "item.h"
-#include "linguistics.h"
 #include "map.h"
-#include "mapset.h"
-#include "projection.h"
-#include "search.h"
 #include "transform.h"
-#include "util.h"
 #include "xmlconfig.h"
 #include <glib.h>
-#include <math.h>
+#include <math.h>  // IWYU pragma: keep for sqrt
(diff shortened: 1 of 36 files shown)


CMakeLists.txt
```diff
@@ -34,6 +34,11 @@ set(NAVIT_VERSION_MAJOR "0")
 set(NAVIT_VERSION_MINOR "5")
 set(NAVIT_VERSION_PATCH "7")
 
+find_program(iwyu_path NAMES include-what-you-use iwyu REQUIRED)
+if (iwyu_path)
+  set(CMAKE_C_INCLUDE_WHAT_YOU_USE "iwyu" )
```
jkoan: Shouldn't this be iwyu_path instead of a hardcoded string?

christf: absolutely! Sorry for this.

navit/search_houseno_interpol.c
```diff
@@ -18,21 +18,16 @@
...
 #include "map.h"
-#include "mapset.h"
-#include "projection.h"
-#include "search.h"
 #include "transform.h"
-#include "util.h"
 #include "xmlconfig.h"
 #include <glib.h>
-#include <math.h>
+#include <math.h>  // IWYU pragma: keep for sqrt
```
jkoan, line 30: Do you know why iwyu will remove that one?

christf: No, and I did not further investigate after I found in their documentatiin that iwyu was grown from specific google code bases and therefore in the real world needs help by setting pragmas.

Issue #1424, opened 2026-02-01
allow static code analysis to be run as part of ci_sanity check and locally


christf:
codefactor runs on the codefactor platform only. Maybe we can find a tool that works on the developers machine.

Supermagnum:
I use Linux mint, its Ubuntu based. 
I suggest:
cppcheck --enable=all --library=gtk --library=posix \
    --std=c99 --suppress=missingIncludeSystem \

And:
gcc -Wall -Wextra -Wpedantic -fanalyzer \
    $(pkg-config --cflags glib-2.0) \
    -fsyntax-only 

scan-build gcc -c aprs.c $(pkg-config --cflags glib-2.0)

flawfinder

christf:
adding flags to gcc should be fairly easy. Let's try this first....

Supermagnum:
These are the complete list of tools that I can use:
Static Analysis
cppcheck — general static analyzer
clang-tidy — comprehensive linter
flawfinder — security vulnerability scanner
scan-build — Clang static analyzer wrapper
gcc -fanalyzer — GCC's built-in static analyzer

Runtime/Dynamic Analysis
Valgrind — memory leaks, invalid access, uninitialized values
AddressSanitizer (-fsanitize=address)
UndefinedBehaviorSanitizer (-fsanitize=undefined)
ThreadSanitizer (-fsanitize=thread) — if multithreaded

Dead Code Detection
cppcheck --enable=unusedFunction
gcc -Wunused-function -Wunused-variable

Additional Useful Warnings
-Wconversion — implicit type conversions that could lose data
-Wshadow — local variable shadows another

PR #1427, opened 2026-02-02
Enable static code analysis during build

christf:
this enables one of the options named in #1424 as part of the regular build.

I am not sure we want this turned on always, Let me read your thoughts please.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,13 @@ IF(NOT CMAKE_BUILD_TYPE)
 	SET(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." FORCE)
 ENDIF(NOT CMAKE_BUILD_TYPE)
 
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+	find_program(iwyu_path NAMES include-what-you-use iwyu)
+	if (iwyu_path)
+		set(CMAKE_C_INCLUDE_WHAT_YOU_USE ${iwyu_path} )
+	endif()
+endif()
+
 # binary name
 set(NAVIT_BINARY navit CACHE STRING "Navit binary name")
 # install path
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -873,7 +873,6 @@ CHECK_FUNCTION_EXISTS (_atoi64 HAVE__ATOI64)
 
 CONFIGURE_FILE("${CMAKE_CURRENT_SOURCE_DIR}/config.h.in"
 			   "${CMAKE_CURRENT_BINARY_DIR}/config.h" )
-
 # Compile with -Wall -Wextra. We need all the help we can get from the compiler :-).
 # Disabled warnings:
 # -Wno-missing-field-initializers: Used a lot, does not seem problematic.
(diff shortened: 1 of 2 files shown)


christf (commented):
On discord, we discussed we would want this on only for debug builds. Somehow gh is preventing myself to request changes but that is exactly what I would like to do to keep a task.

jkoan:
I think we should add all flags starting with "-W" only to CMAKE_C_FLAGS_DEBUG so its only relevant for debug builds

CMakeLists.txt
```diff
@@ -866,27 +873,28 @@ CHECK_FUNCTION_EXISTS (_atoi64 HAVE__ATOI64)
...
-add_compiler_flag_if_available("-Wmissing-prototypes")
-add_compiler_flag_if_available("-Wstrict-prototypes")
-add_compiler_flag_if_available("-Wformat-security")
+add_debug_compiler_flag_if_available("-Wall")
+add_debug_compiler_flag_if_available("-Wno-unused-parameter")
+add_debug_compiler_flag_if_available("-Wno-sign-compare")
+add_debug_compiler_flag_if_available("-Wno-missing-field-initializers")
+add_debug_compiler_flag_if_available("-Wundef")
+add_debug_compiler_flag_if_available("-Wcast-align")
+add_debug_compiler_flag_if_available("-Wpedantic")
```
christf: seeing the warnings, now I do not think -Wpedantic is a good idea any more.

christf:
... should we change ci to also do a debug build to obtain the warnings and maybe use werror to break it?

if we want to go down this route, the warnings need to be adressed in here too.

christf:
There is still this:
Warning: include-what-you-use reported diagnostics:
error: unknown argument '-fanalyzer'; did you mean '-Xanalyzer'?

And someone(TM) really should go through all of these ;)

jkoan:
> There is still this:
> Warning: include-what-you-use reported diagnostics:
> error: unknown argument '-fanalyzer'; did you mean '-Xanalyzer'?
> 
> And someone(TM) really should go through all of these ;)

If I remember correctly iwyu is based on clang so it probably also only accepts those flags. The interesting question is how it would be handled if gcc is used as compiler but iwyu is also enabled

christf:
> There is still this: Warning: include-what-you-use reported diagnostics: error: unknown argument '-fanalyzer'; did you mean '-Xanalyzer'?
> 
> And someone(TM) really should go through all of these ;)

this is done now but it required a wrapper that would drop that parameter. This should also handle the gcc case.

cmake/iwyu-wrapper.sh
```diff
@@ -0,0 +1,22 @@
...
+
+n=$#
+while [ "$n" -gt 0 ]; do
+	arg=$1
+	shift
+	case "$arg" in
+		-fanalyzer) ;;
+		*) set -- "$@" "$arg" ;;
+	esac
+	n=$((n - 1))
```
christf: this I would like to have another look at. maybe there is a better way

jkoan:
How about just using clang with iwyu? Just throw a warning that "-fanalyzer" got disabled because iwyu is active together with a helpful line how to switch it back on.

Alternatively is there something we could do to replace fanalyzer? Maybe we could use clang for our default Linux build and use gcc only when necessary (wince, tomtom, android(?)) this way we could completely rely on clang tools for more modern tool usage like we already started with clang-format. Like clang-tidy and Scan-Bild. This way we could remove the need for the gcc option and even get more advanced tooling

PR #1418, opened 2026-01-31
Add APRS and APRS SDR plugins

Supermagnum:
APRS plugin: real-time tracking of APRS stations on map. Bell 202/AX.25 decoding, SQLite storage, moving objects display.

APRS SDR plugin: RTL-SDR hardware access with Bell 202 demodulation for APRS packet reception.

PR #1419, opened 2026-01-31
Feature/driver break

Supermagnum:
Driver_break plugin
Rest period management for vehicles (cars, trucks, hiking, cycling).
EU Regulation EC 561/2006 compliance for commercial vehicles.
DNT/pilgrimage route support. Map-based POI discovery.
Energy based routing.
ECU support.

Co-authored-by: Cursor <cursoragent@cursor.com>

PR #1436, opened 2026-02-10
Add APRS and APRS SDR plugins

Supermagnum:
This PR contains only APRS-related changes, separated from mixed concerns in the original PR.

## Changes
- APRS plugin with modular SDR support
- APRS SDR plugin for RTL-SDR hardware
- APRS documentation and test results
- APRS-specific fixes and enhancements

This is a cleaned version of PR #1418 with only APRS commits.

Made with [Cursor](https://cursor.com)


README.md
```diff
@@ -37,6 +37,22 @@ Current vehicle position from \
 — the location service of several mobile platforms. \
 — a UDP server (friends tracking) (experimental).
 
+Plugins
+=======
```
christf, line 41: please remove those conflict markers

Supermagnum:
Currently the tests are failing. 
Will push when those is sorted.

CLAUDE.md
jkoan, line 1: Same as in #1419 : 
Just no, we won't encourage users to run Code through AI Datacenters by providing files. We have the docs where we provide users as well as developers with all the informations they need. 
If something is missing in there feel free to add it as separate PRs, but dont bloat this PR.

Supermagnum: deleted

Supermagnum:
Note that the DSP functions in  https://github.com/Supermagnum/navit/tree/feature/aprs-clean/navit/plugin/aprs_sdr are very finicky. I will try to get rx working with my libresdr, but that can be difficult because the driver for that is made for Gnuradio.

Supermagnum:
On why this PR cannot be split to reduce the line count

A reasonable first reaction to the size of this PR is that it should be split into a core APRS change and a separate APRS-SDR change so each is smaller. We investigated this carefully and the conclusion is that splitting would not reduce the total number of lines, because the bulk of the diff is not plugin code and is not optional.

Here is where the lines actually come from:

- Translation files (the largest part by far). Adding the APRS menu entries introduces new translatable strings. The project's own sanity-check stage requires every language file to be regenerated so it stays in sync with those strings. When the regeneration runs, it does not only add the handful of new APRS strings; it also reorders existing entries across all of the language files to match the updated extraction order. This produces a very large, mechanical diff. We confirmed this is mandatory and not an artifact of an out-of-date branch: regenerating against the current tip of trunk produces essentially the same large translation diff. This part of the change belongs entirely to the core APRS side, since the menu strings live in the shared configuration, not in the SDR plugin.

- Documentation. The plugin ships a full set of user documentation pages. These are legitimate additions and account for a few thousand lines on their own.

- The actual plugin source and tests. This is the smallest of the three buckets.

Why splitting does not help:

- The translation regeneration is driven by the core menu strings, so it cannot move to the SDR branch. Whichever branch carries the menu entries also carries the full translation diff. Splitting would leave the core branch almost exactly as large as the combined PR is now.

- The SDR plugin on its own is small. Separating it would create one tiny branch and one branch that still contains nearly all of the lines. It changes how the work is organized but not the total size.

- The line count here is required content, not accidental bloat. We verified there is no unrelated upstream history, no stale-base inflation, and no large generated files committed by mistake (one stray generated configuration file was found and removed). After that, the remaining size is simply what it costs to add a plugin that introduces new translatable menu strings in a project that regenerates all language files.

In short: the size is dominated by mandatory translation regeneration and documentation, both of which stay together regardless of how the plugins are divided. Splitting the plugins would improve nothing about the line count and would only add coordination overhead between two dependent branches.

PR #1437, opened 2026-02-10
Feature/code quality

Supermagnum:
This addresses the concern about mixed concerns in the original  #1418 PR.

### DIFF
--- a/navit/navit.dtd
+++ b/navit/navit.dtd
@@ -127,6 +127,7 @@
 <!ATTLIST osd y CDATA #IMPLIED>
 <!ATTLIST osd src CDATA #IMPLIED>
 <!ATTLIST osd command CDATA #IMPLIED>
+<!ATTLIST osd data CDATA #IMPLIED>
 <!ELEMENT itemgra (polygon|polyline|text|circle|icon|image|arrows|spikes)*>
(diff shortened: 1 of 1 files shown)


navit/.clang-format
christf, line 1: please remove this

christf:
I have a few nits with this one:
* the first commit does not refactor existing code in line with its commit message but it adds a few k LOC. I do not agree with this.
* the last commit (chaning the clang-format) should be dropped from this PR.
* sanity check fails and as such this does not compile.

I do like the second commit that completes the dtd by specifying the data element for the osd. It would be great if this would have been in the commit message to allow me to find it.

Supermagnum:
> see comment.

Those should now be resolved

christf:
Great, now we have the content. Commit structure needs a cleanup and then it is good to go in my point of view. Do you need help with that?

Supermagnum:
> Great, now we have the content. Commit structure needs a cleanup and then it is good to go in my point of view. Do you need help with that?

No.

christf:
now the only remaining nit is the commit message because there is nothing about trailing newline validation in here - but that can be fixed on merge. Thank you for your work!

PR #1440, opened 2026-02-11
minor build fixes

christf:
this isolates two patches from #1400 as they are useful stand-alone, adding a missing include and using G_PI instead of self-defined M_PI constant.

### DIFF
--- a/navit/maptool/osm.c
+++ b/navit/maptool/osm.c
@@ -50,11 +50,6 @@ struct relations;
 struct relations_func;
 struct zip_info;
 
-#ifndef M_PI
-#    define M_PI 3.14159265358979323846
-#    define M_PI_4 0.785398163397448309616
-#endif
-
 static int in_way, in_node, in_relation;
 osmid nodeid, wayid;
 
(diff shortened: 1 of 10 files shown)


navit/maptool/osm.c
```diff
@@ -49,9 +50,9 @@ struct relations;
 struct relations_func;
 struct zip_info;
 
-#ifndef M_PI
-#    define M_PI 3.14159265358979323846
-#    define M_PI_4 0.785398163397448309616
+#ifndef G_PI
+#    define G_PI 3.14159265358979323846
+#    define G_PI_4 0.785398163397448309616
```
jkoan: just include [navit/support/glib/gtypes.h](https://github.com/navit-gps/navit/blob/1d7af68bd686e878e01e9f1929ab96385bd16f34/navit/support/glib/gtypes.h#L111)? Aka glib/glib.h

christf: on that note, when do I include "glib.h" and when <glib.h>?

jkoan: For everything that should be system provided i use `<glib.h>`, so for everything in support as well. But i don't think it doesn't matter much. https://stackoverflow.com/a/3162067

christf: well, there is a ./navit/support/glib/glib.h - this is what I assume is being included on `#include "glib.h"`

jkoan: You're right. I missed that that was already there

christf:
nice, removing the defines for G_PI has worked on all platforms - so let me rebase to squash the fixup commit and then we can call it a day.

PR #1442, opened 2026-02-12
Update Android Build 

jkoan:
- [x] update SDK including AGP 
- [x] update NDK
- [x] fix fdroid 
- [x] fix apk filename


navit/android/AndroidManifest.xml
```diff
@@ -1,14 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
     package="org.navitproject.navit"
-    android:sharedUserId="org.navitproject.navit"
```
jkoan, line 4: Revert and set sharedUserMaxSdkVersion=32

christf:
Should we use renovate for such updates in the future?

jkoan:
> Should we use renovate for such updates in the future?

I have already heard about it but never used it. So I don't know if it would work. This also wasn't hard as within android studio there are assistants and context help for all the things, sprinkle in some ai for some less important deprecation warning questions and that's about 2h of work including builds and tests

navit/android/AndroidManifest.xml
```diff
@@ -1,14 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="org.navitproject.navit"
+    xmlns:tools="http://schemas.android.com/tools"
     android:sharedUserId="org.navitproject.navit"
```
jkoan: To be able to install two apps with the two different app IDs we need to only set this is on the main (old) app id. That way both app IDs don't need to use the same signing key

navit/support/glib/ghash.c
```diff
@@ -32,6 +32,8 @@
 
 #include "glib.h"
 #include "galias.h"
+#include "gprimes.h"
+#include "gatomic.h"
```
christf, line 34: this does not look like it has the right order

navit/support/wordexp/glob.c
```diff
@@ -31,7 +31,7 @@
 /*
  * @brief searches for all the pathnames matching pattern according to the rules
  * which is similar to the rules used by common shells.
- * here: expanding of �*� and �?� only in filenames
+ * here: expanding of �*� and �?� only in filenames
```
christf: there is a charset problem here

navit/support/wordexp/glob.c
```diff
@@ -97,6 +97,7 @@ int glob(const char *pattern, int flags,
 #include <dirent.h>
 #include <string.h>
 #include <fnmatch.h>
+#include <stdlib.h>
```
christf, line 101: system includes should come after the project-specific ones.

.github/workflows/build.yml
```diff
@@ -140,8 +140,6 @@ jobs:
       #      ln -sf navit/navit.dtd navit.dtd
       #      mkdir test-results
       #      ./gradlew lint test
-      - name: Run Javadoc
```
christf, line 143: why remove that?

jkoan: Because genarating the JavaDoc is broken with AGP 9.0+ due to the new DSL

navit/android/AndroidManifest.xml
jkoan: Implemented in 5348345e57f8969baf9fa06121354272c35f6880

gradle/scripts/git-scm-version.gradle
```diff
@@ -1,34 +1,44 @@
...
-    }
-    dependencies {
-        classpath 'org.ajoberstar.grgit:grgit-core:5.3.3'
+def execOps = project.services.get(org.gradle.process.ExecOperations)
+
+ext {
+    // git describe
+    def stdoutName = new ByteArrayOutputStream()
+    execOps.exec {
+        it.commandLine 'git', 'describe', '--tags', '--match', 'v[0-9.rc]*', '--dirty'
```
christf: this glob looks a  bit weird and by the form of it it seems to be an invitation for misinterpretation.

Let me elaborate: 
The glob as written matches all tags that 
* starts with a "v"-character
* followed by a character that is either:
- a number
- a literal dot
- a "r"
- a "c"
* followed by any other string


If we find all our version tags start with a v and are followed by a digit, then let's just match that.

"v[0-9]*"

this is much clearer.

If we want it to match the trinity of major/minor/patch, followed by anything else, then we could do

"v[0-9]\.[0-9]\.[0-9]*"

this would, however break if we ever released v0.10.0.


Personally, I think "v[0-9]*", while not very specific, is probably fine and it also works for release candidates or betas etc as long as their first two characters match "v[0-9]"

navit/support/glib/gatomic.h
```diff
@@ -0,0 +1,13 @@
+//
+// Created by jan on 18.02.26.
+//
+
+#ifndef NAVIT_GATOMIC_H
+#define NAVIT_GATOMIC_H
+
+#include <glib.h>
```
christf: I think this should include "gtypes.h" instead of "<glib.h>" https://github.com/christf/navit/actions/runs/23414504713/job/68107450904

navit/support/glib/gprimes.h
```diff
@@ -0,0 +1,13 @@
+//
+// Created by jan on 18.02.26.
+//
+
+#ifndef NAVIT_GPRIMES_H
+#define NAVIT_GPRIMES_H
+
+#include <glib.h>
```
christf: also here, I think this should include "gtypes.h"

build.gradle
```diff
@@ -61,22 +44,21 @@ android {
...
-                arguments '-DDISABLE_CXX=y', '-DUSE_PLUGINS=n', '-DBUILD_MAPTOOL=n', '-DXSL_PROCESSING=y', '-DSAMPLE_MAP=n', '-DGRADLE_BUILD_DIR='+layout.buildDirectory.file('generated/navit_generated').get().getAsFile()
+                arguments '-DDISABLE_CXX=y', '-DUSE_PLUGINS=n', '-DBUILD_MAPTOOL=n', '-DXSL_PROCESSING=y', '-DSAMPLE_MAP=n', '-DGRADLE_BUILD_DIR='+layout.buildDirectory.file('generated/navit_generated').get().getAsFile(),"-DANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES=ON"
             }
         }
     }
     buildTypes {
         release {
             minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
```
christf, line 54: https://android.googlesource.com/platform/sdk/+/master/files/proguard-android-optimize.txt says to "test thoroughly if we want to go this route". Are we equipped to do this? What are the benefits of the optimized versions? Reading through https://stackoverflow.com/questions/35321742/android-proguard-most-aggressive-optimizations/35509428#35509428 seems to suggest, there is trouble with "overloadaggressively" but I am not sure I am putting it into correct context.

build.gradle
```diff
@@ -105,57 +90,54 @@ android {
...
-        // create tasks to generate Javadocs
-        task("generate${variant.name.capitalize()}Javadoc", type: Javadoc) {
-            source = android.sourceSets.main.java.srcDirs
-            destinationDir = file("navit/android/build/outputs/docs/javadoc/")
-            title = rootProject.name
-            options.memberLevel = JavadocMemberLevel.PACKAGE
-            verbose = true
-            failOnError false
-        }
+    lint {
```
christf: this looks like a duplicate

build.gradle
```diff
@@ -105,57 +90,54 @@ android {
...
+        abortOnError = true
+        disable 'UnusedResources'
+    }
+    buildFeatures {
```
christf: this looks like a duplicate

build.gradle
```diff
@@ -48,11 +32,10 @@ android {
         }
     }
     defaultConfig {
-        compileSdk 33
-        compileSdkVersion 33
+        compileSdk = 36
         applicationId "org.navitproject.navit"
-        minSdkVersion 10
-        targetSdkVersion 33
+        minSdk 21
```
christf, line 37: do we really want this bump? I guess we are effectively disconnecting old androids from upgrades by this.

build.gradle
jkoan: To my knowledge the optimisations are for java only, so a quick check that all java callbacks are there should be enough. But good call that we should test it again.

build.gradle
christf: as discussed out-of-band: we want this bump. sdk21 is android 5 from 2015, so we still support devices with software that was current 10 years ago.

navit/support/glib/fake.c
```diff
@@ -65,7 +65,7 @@ g_private_new_navit (void)
 	if (pthread_key_create(&key, NULL)) {
 		fprintf(stderr,"pthread_key_create failed\n");
 	}
-	return key;	
+	return &key;
```
christf: won't that address be invalid after the function ends? It gets defined here and it gets destroyed when the function is removed from the stack.

build.gradle
```diff
@@ -5,39 +5,23 @@ apply plugin: 'checkstyle'
...
-            name 'Google'
-        }
+        mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.1.3'
-
-        // NOTE: Do not place your application dependencies here; they belong
-        // in the individual module build.gradle files
+        classpath 'com.android.tools.build:gradle:9.1.0'
```
christf, line 11: why use 9.1 here but 9.3.1 above?

build.gradle
```diff
@@ -105,57 +90,48 @@ android {
...
-        }
-    }
+clean {
+    delete "./android-builddir"
 }
 
 dependencies {
-    implementation fileTree(include: ['*.jar'], dir: 'navit/android/libs')
-    androidTestImplementation('com.android.support.test.espresso:espresso-core:3.0.2', {
+    androidTestImplementation('com.android.support.test.espresso:espresso-core:2.2.2', {
```
christf: interesting, why the downgrade?

build.gradle
jkoan: i don't see a 9.3.1 anywhere, that the only place i see a AGP version. What am i missing?

build.gradle
jkoan: Probably by mistake, fixed

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,5 +2,5 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-all.zip
-distributionSha256Sum=a8da5b02437a60819cad23e10fc7e9cf32bcb57029d9cb277e26eeff76ce014b
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.3.1-all.zip
```
christf, line 5: this is where I saw the 9.3.1 - should we harmonize that?

jkoan: Ah okay. Then no: https://developer.android.com/build/releases/about-agp#updating-gradle

navit/support/glib/fake.c
christf: So, I have looked at this some more and there is one more topic. In this branch, an address is returned. In the windows case above, it is an index. I do not have deep understanding of the code in this area but it does seem to me that in both cases either an address or an index should be returned.

jkoan: I haven't found a good solution for this so far, so I will actually merge this as it's now, because from my testing it somehow works as is

jkoan:
I looked at fake.c again and fixed the reported issue. @christf I would appreciate you're review. If you have no objections I would like to merge soon

christf:
I like the changeset and I think the bit around the invalid return value is resolved. There is a new topic around malloc() potentially failing. When that happens all bets are off anyways. gmalloc() terminates the application when this happens so I guess it might be ok.

christf (approved):

PR #1460, opened 2026-02-17
feat(graphics): accuracy of pitch increased

### LINKED ISSUE #585: Pitch rounding errors
The ```pitch``` variable in the ```navit``` struct should be changed from an ```int``` to a ```float``` data type so that pitch correction (see https://github.com/navit-gps/navit/pull/303 and https://github.com/navit-gps/navit/pull/306) does not have to round the result. Currently, this is causing some problems with the pitch in high resolution displays where the correction causes the stored ```pitch``` value to be a low value.

F.ex. on a Galaxy S9+ with a display resolution of 2960x1440px the default pitch of 20 gets stored as ```20*sqrt(240*320)/sqrt(2960*1440)=2.68…``` which gets rounded to 3. In reverse, this effectively means a pitch of ```3*sqrt(2960*1440)/sqrt(240*320)=22.34…``` which gets rounded to 22.
This is a 10% error.

Definition: https://github.com/navit-gps/navit/blob/trunk/navit/navit.c#L166
Round function should be removed at: https://github.com/navit-gps/navit/blob/trunk/navit/navit.c#L2620 and https://github.com/navit-gps/navit/blob/trunk/navit/navit.c#L2860

Not sure at what other locations changes should be made.

christf:
this allows to store pitch with increased accuracy and closes #585


navit/navit.c
```diff
@@ -2913,8 +2913,8 @@ int navit_get_attr(struct navit *this_, enum attr_type type, struct attr *attr,
         attr->u.num = this_->osd_configuration;
         break;
     case attr_pitch:
-        attr->u.num = round(transform_get_pitch(this_->trans) * sqrt(this_->w * this_->h)
-                            / sqrt(240 * 320));  // Pitch corrected for window resolution
+        attr->u.num = transform_get_pitch(this_->trans) * sqrt(this_->w * this_->h)
```
jkoan: Num is a long type, numd would be a double

jkoan (changes requested):

navit/graphics/qt5/event_qt5.cpp
```diff
@@ -113,57 +112,51 @@ static struct event_watch* event_qt5_add_watch(int fd, enum event_watch_cond con
...
-    event_qt5_main_loop_run,
-    event_qt5_main_loop_quit,
-    event_qt5_add_watch,
-    event_qt5_remove_watch,
-    event_qt5_add_timeout,
-    event_qt5_remove_timeout,
-    event_qt5_add_idle,
-    event_qt5_remove_idle,
-    event_qt5_call_callback,
+    event_qt5_main_loop_run, event_qt5_main_loop_quit, event_qt5_add_watch,
```
christf: meh

navit/graphics/qt5/graphics_qt5.cpp
```diff
@@ -151,21 +151,13 @@ static void font_destroy(struct graphics_font_priv* font) {
...
-static const char* fontfamilies[] = {
-    "Liberation Sans",
-    "Arial",
-    "NcrBI4nh",
-    "luximbi",
-    "FreeSans",
-    "DejaVu Sans",
-    NULL,
+static const char *fontfamilies[] = {
+    "Liberation Sans", "Arial", "NcrBI4nh", "luximbi", "FreeSans", "DejaVu Sans", NULL,
```
christf: meh

christf:
well, I picked this up because I saw that issue.... but now coming back to it, I see the gh-issue but I fail to see an issue. So what, if this is not accurate?
Doing int arithmetics is significantly faster on slow hardware. Maybe the best solution is to leave it as is?

Issue #1474, opened 2026-03-02
Problems on Android


clarified:
I have been using navit on an old wince system for many years so I am pretty familar with navit.
I decided to buy a Samsung Android tablet primarily to run navit since it was reported to have excellent gps capabilities.

I installed navit from FDroid, but was surprised to discover that it was in a parlous state. If I had not been an experienced user, and had used adb before, it would have been impossible to get it working.

1) I tried to follow the instructions and download a map. The download stalled with no timeout or warning. Of course, I suspected that the upstream servers were no longer working, but that would have been the first hurdle for a new user. As I normally compile my own maps with maptool, I _eventually_  managed to install a map using adb. I had hoped to put it on an sdhc card, but the instructions to place it under
/sdcard/navit seemed to be wrong: it did not work. Much,much later, after I had unzipped the apk and looked at the navit.xml, it seems that the directory should have been /sdcard/Download/ and be of the form osm_bbox_*.bin" ! Who thought that one up and how is the end user supposed to know?
I still haven't tried to load from the external sdhc card, although that ought to be the simplest option.

2) It was only after I had extracted the right navit.xml from the apk ( I did not notice the various versions
for different screen resolutions at first), edited it and pushed with adb:

`db -d push ~/mapping/navit/android/navit.xml_android /storage/emulated/0/Android/data/org.navitproject.navit/files/navit.xml`

that I began to begin to get something more or less working. 

3) I needed my bookmarks from the wince navit. Eventually I also pushed that with adb to the same
place: `/storage/emulated/0/Android/data/org.navitproject.navit/files/`

4) I have had problems with the screen layout. At times the 2D icon appeared, yet the screen shows a tilted 3D layout. At other time, the icon showed 3D, but the screen was always 2D. I have not investigated that, nor tried to find the layout files to inspect what might be wrong.

These are just initial comments before trying to use navit in earnest. The main point is that an ordianry user would have no chance of getting it working. About shows 0.5.7.arm64-v8a+git:v0.5.7-dirty...

I aapreciate the huge amount of work that has gone into navit

clarified:
Now that I have tried to use the android navit in a car, I can report additional problems.

1) The map will only display in "3D" which I happen to dislike and does not seem to work with my contour map also installed. The bug seems to be that when I select 2D in the settings->display, this lasts only a second or so. I suspect that the old 3D state returns when new locations come in. This is just speculation. When I tried the program in demo mode before trying in a car, I could select 2D and it persisted. 

[...]

PR #1488, opened 2026-03-10
fix(core): Size of center_timeout aligned with u.num to be long

christf:
In #1474 it was reported that reducing the timeout parameter makes auto zoom behave differently. This aligns the size of the underlying datastructures to help


christf:
in #1474 it was discussed that the datatype change is not really the root cause of the weirdness. Do we still keep the patch? I am fine with either.

jkoan:
When looking at other cases we have multiple instances where we read a attr and store if in a long just to unpack it into a int again later, so i think it it fine like its now

PR #1489, opened 2026-03-10
Feature/navit dash

Supermagnum:
Summary
Documented the Navits dbus standard  for use by external dashboards (instrument clusters, head units, loggers, etc.).   Also added example svg icons.

Compatible with:
https://github.com/Supermagnum/navit/tree/feature/driver-break

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -18,7 +18,7 @@
 /* Versions */
 #cmakedefine GIT_VERSION "@GIT_VERSION@"
 #ifndef GIT_VERSION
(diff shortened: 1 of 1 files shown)


Supermagnum:
It builds perfectly on my machine, and I am unable to locate the errors that causes builds to fail.
This also has a plugin and it passes all tests.
https://github.com/navit-gps/navit/pull/1419
Does anyone know what's causing this?

christf:
why not use the dbus plugin for this?

Supermagnum:
> why not use the dbus plugin for this?

I will look into that.

PR #1490, opened 2026-03-10
fix(gui_internal): pass HAVE_SYSTEM as preprocessor flag to support c…

christf:
…alling browser

this defines HAVE_SYSTEM as preprocessor directive if the corresponding cmake variable is set

### DIFF
--- a/docs/user/configuration/Gui_internal.rst
+++ b/docs/user/configuration/Gui_internal.rst
@@ -32,6 +32,16 @@ with
 
 inside the <gui .../> tag.
 
+The internal gui has a feature to call a browser on OSM nodes. To use it,
+place a shellscript called navit-browser.sh into your PATH.
+
+The shellscript should contain a call to the browser that is meant to be
+executed, similar to this
+
+.. code-block: bash
+   #!/bin/bash
+   xdg-open $@
+
 Menu
 ----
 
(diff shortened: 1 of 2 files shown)


christf:
As a build, maybe the browser should be configurable and this setting documented.

jkoan:
i know that i have used the system function multiple times, this was always working. Is there a warning/error somewhere i missed?

christf:
without the patch, navit tells me "Error: External commands were disabled during compilation, cannot call" when attempting to call a browser. This is because of the ifdef-guards in this particular place here https://github.com/navit-gps/navit/blob/1dfa645751fba6efb00410617955b3d2dec226d0/navit/gui/internal/gui_internal.c#L845

I am not sure if we should be guarding in all places or in none. This patch just ensures the define is properly passed.

docs/user/configuration/Gui_internal.rst
```diff
@@ -32,6 +32,16 @@ with
...
 
+The internal gui has a feature to call a browser on OSM nodes. To use it,
+place a shellscript called navit-browser.sh into your PATH.
+
+The shellscript should contain a call to the browser that is meant to be
+executed, similar to this
+
+.. code-block: bash
+   #!/bin/Bash
+   firefox $@
```
christf: if we document like this, it should be xdg-open, instead of any specific browser.

PR #1492, opened 2026-03-11
Feature/navit safety

Supermagnum:
Navit Safety Plugin

The Navit Safety plugin addresses POI reliability and resource planning in remote, arid, or sparsely serviced terrain. In populated areas with good map coverage, finding fuel, water, or shelter is straightforward. In remote or depopulating regions, POI data cannot be trusted at face value: a fuel station on the map may have closed years ago; a spring may be seasonal or dry. This plugin treats POI reliability as a first-class input to resource planning and provides location-aware remote mode with conservative buffer thresholds.

The plugin handles:

    POI confidence scoring — Sources are ranked by reliability; only confirmed or reasonably recent POIs count as reliable resupply points.
    Location-aware remote mode — Automatic detection via POI density, Köppen climate classification, and known sparse-region polygons; manual override available.
    Resource planning — Fuel and water buffers, lookahead planning, carry capacity, and handling of unconfirmed stops for desert driving and foot travel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,16 +37,6 @@ Current vehicle position from \
 — the location service of several mobile platforms. \
 — a UDP server (friends tracking) (experimental).
 
-Plugins
-=======
-
-Navit includes several plugins that extend its functionality:
-
-- **Driver Break Plugin** — Vehicle pause and rest stop management with POI discovery and energy-based routing
-- **APRS Plugin** — Real-time APRS station tracking with packet decoding and map display
-- **APRS SDR Plugin** — RTL-SDR hardware support for direct APRS packet reception
-
-For complete plugin documentation, see the `Navit User Manual <https://navit.readthedocs.io/en/latest/user/plugins/index.html>`_.
 
 Translation
 ===========


scripts/ci_sanity_checks.sh
```diff
@@ -86,10 +86,10 @@ for f in $(git --no-pager diff --name-only refs/remotes/origin/trunk | sort -u);
             fi
         fi
 
-        # Formats any *.h *.c and *.cpp files
+        # Formats any *.h *.c and *.cpp files (per navit programming guidelines: astyle)
         if [[ "${f: -2}" == ".h" ]] || [[ "${f: -2}" == ".c" ]] || [[ "${f: -4}" == ".cpp" ]]; then
             echo "[INFO] Checking for indentation and style compliance on ${f}..."
-            clang-format -i "${f}"
+            astyle --indent=spaces=4 --style=attach -n --max-code-length=120 -xf -xh "${f}"
```
christf, line 92: please, no. We just replaced that. The fix needs to happen to the docs instead.

christf:
there is more than a safety plugin in this PR. is this intentional?

PR #1496, opened 2026-03-13
Patch:core:Some debug statements

adevoss:
fix debug statements

### DIFF
--- a/navit/navigation.h
+++ b/navit/navigation.h
@@ -34,6 +34,9 @@
  * meters per second. */
 #define MPS_TO_KPH 3.6
 
+#define INVALID_ANGLE 361
+#define ROUNDABOUT_EXTRA_LENGTH 50
+
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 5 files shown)


navit/navigation.h
```diff
@@ -34,6 +34,9 @@
  * meters per second. */
 #define MPS_TO_KPH 3.6
 
+#define INVALID_ANGLE 361
+#define ROUNDABOUT_EXTRA_LENGTH 50
```
christf, line 38: what does ROUNDABOUT_EXTRA_LENGTH do?

christf:
I think, overall this is good, just the bit with the ROUNDABOUT_EXTRA_LENGTH still eludes me

navit/navigation.h
adevoss: I don't know.

adevoss:
Just moved that line of code from `navigation.c` to `navigation h`. All constants in one place.

PR #1497, opened 2026-03-14
code in phrase.c is not used

adevoss:
code in phrase.c and phrase.h is not used


jkoan:
Thank you for your contribution :)

PR #1259, opened 2024-04-06
Use TTS when sample is missing

adevoss:
When using samples there always is text that can't be spoken: City names. It is impossible to create samples for all of them.
Instead of being silent, TTS is used to speak missing samples.
If you have a complete set of samples, text will be spoken using samples and city names will be spoken using TTS.

Help wanted:
- [x] Parse arguments of speech program once  (at startup) instead of at each instruction.
- [x] Fixes #1538
- [x] Pass sanity check script
- [x] Test speaking street names
- [x] Add optional `data_tts` to `speech` attribute in `navit.dtd`
- [x] Add optional `play_multiple_samples` to `speech` attribute in `navit.dtd`
- [x] ~~using 'TTS only' causes a crash~~

PR #1498, opened 2026-03-14
Add:docs:sample files list (nl) to use for speech

### LINKED ISSUE #1482: No complete list of the samples required. There is one for the Dutch language
In chapter 4.1.1.1:
"Unfortunately, there is no complete list of the samples required."

For the Dutch language I've compiled a list of filenames of samples required. I would like to add this to the documentation.
No idea how to do that.

Also I would like to add an explanation how to compile such a list in other languages:
- install a TTS program (if #1259 is merged)
- create a speech tag in navit.xml
- set debug level to "debug" in navit.xml

Repeat the steps below until all samples are created
- set a destination and make the trip
- check the log file for "Can't speak the text ..." / if #1259 is merged: "Spoken by TTS: ..." (except city names)
- create missing sample file(s)

adevoss:
Edit documentation and add a Dutch sample files list to use for speech.

Fixes #1482

### DIFF
--- a/docs/user/configuration/Configuration_Full_list_of_options.rst
+++ b/docs/user/configuration/Configuration_Full_list_of_options.rst
@@ -2427,6 +2427,18 @@ tags.
    |              |       |              | tion <Config |              |
    |              |       |              | uration>`__. |              |
    +--------------+-------+--------------+--------------+--------------+
+   | data-tts     |       |              | Same as      | `            |
+   |              |       |              | 'data'.      |            ` |
+   |              |       |              | When using   |              |
+   |              |       |              | sample files |              |
+   |              |       |              | use this     |              |
+   |              |       |              | option to    |              |
+   |              |       |              | configure    |              |
+   |              |       |              | TTS and use  |              |
+   |              |       |              | 'data' to    |              |
+   |              |       |              | configure    |              |
+   |              |       |              | sample files.|              |
+   +--------------+-------+--------------+--------------+--------------+
    | cps          |       | 0 - ∞        | Speed at     | ``cps="15"`` |
    |              |       |              | which Navit  |              |
    |              |       |              | should speak |              |
@@ -2483,7 +2495,7 @@ tags.
    +--------------+-------+--------------+--------------+--------------+
    | sample_dir   |       |              | For          | ``sampl      |
    |              |       |              | spe          | e_dir="/path |
-   |              |       |              | ech_cmdline, | /to/waves"`` |
+   |              |       |              | ech_cmdline, | /to/files"`` |
    |              |       |              | defines path |              |
    |              |       |              | where        |              |
    |              |       |              | pre-recorded |              |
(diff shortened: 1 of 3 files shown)


docs/user/configuration/advanced/introduction.rst
```diff
@@ -23,7 +23,10 @@ Note that Navit internally handles all text in UTF-8 encoding. If you use a file
...
 
-Note that if any file that is needed to compose the complete phrase is missing then Navit will be silent. In that case a warning will be printed. Unfortunately, there is no complete list of the samples required. However, all the navigation text is contained in the translation files (.po files), so you can get a rough list.
+Note that if any file that is needed to compose the complete phrase is missing then Navit will be silent. In that case a warning will be printed. There is a list of the samples required for the Dutch language. Lists for other languages can be created by reading the log file. Look for lines containing "Cannot speak text" and create the missing audio files. The list for Dutch can be used as a starting point (the list for the numbers is the same in every language). Other texts are probably different because of the order of words in a sentence.
+
+WARNING: THERE ARE ANNOUNCEMENTS THAT CONTAIN GEOGRAPHICAL NAMES. SO THERE IS NO PRACTICAL USE FOR THIS FEATURE AT THIS MOMENT SINCE IT IS IMPOSSIBLE TO CREATE AUDIO FILES FOR ALL THE NAMES.
```
christf, line 28: I do not think this is adequate. As doumented below, to make this feature useful, vocabulary_name_systematic needs to be set. Did that work for you?

docs/user/configuration/Configuration_Full_list_of_options.rst
```diff
@@ -2427,6 +2427,18 @@ tags.
    |              |       |              | tion <Config |              |
    |              |       |              | uration>`__. |              |
    +--------------+-------+--------------+--------------+--------------+
+   | data-tts     |       |              | Same as      | `            |
```
christf: I think this bit should be part of #1259

docs/user/configuration/advanced/introduction.rst
adevoss: Yes it works for me.

How I understand the docs: `vocabulary_name_systematic` announces street names. Street names are names like "Main street" or "Elvis Presley road".
When I drive I don't need to hear the name of the street. All I want to hear is: "Turn right after 200 meter". The names of the current and next street are on my OSD.

Using #1259 you can set `vocabulary_name_systematic="1"` if you want. All the street names are spoken by TTS (if you don't have a sample file for that street name).
Never tested this but I will test this soon and let you know the result.

Even with `vocabulary_name_systematic="0" there are names of town in the navigation instructions.
These instructions are something like this: "Take the next exit towards Amsterdam".
If you use the the `trunk` branch and sample files then this instruction is not played. Navit stays silent.
Only because one word ("Amsterdam") is missing.

Using #1259 all the available samples are played. Text without sample files is spoken by TTS.
So taking the example above and you have a complete set of sample files: Samples are played for the first part, followed by TTS speaking "Amsterdam".

Instead of Navit staying silent the whole instruction is played.

PR #1503, opened 2026-03-17
chore:author (me) listed twice

adevoss:


jkoan:
This wouldn't work. This file gets automatically generated by scripts/generate_contibutors.sh. We need to add a mapping to the .mailmap

christf:
the line in the .mailmap should look similar to this:

`adevoss <arnobugs@hotmail.com>                                Arno <arnobugs@hotmail.com>`

the first column is the "What should be replaced"-bit and the second column contains the text that is replacing the content from the first column. You can pick your name and target email address.

jkoan (approved):

Issue #1439, opened 2026-02-10
refactor includs in plugin.c


christf:
plugin.h is included twice here. Also it is included in many other c files (without that PLUGIN_C define). this throws off the debugger and is prevents usage of include guards.

This should be refactored for clarity.

PR #1506, opened 2026-03-17
Refactor/plugin header split

pgrandin:
Addresses #1439 
Our current code was maybe slightly abusing the list macros here.


navit/plugin_impl.h
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * You should have received a copy of the GNU Library General Public
+ * License along with this program; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA  02110-1301, USA.
+ */
+
+/**
+ * @file plugin_impl.h
+ * @brief Plugin implementation macros — included only by plugin.c.
```
jkoan, line 22: It its only included inside plugin.c why do we need a header for it?

pgrandin: It felt cleaner that way, keeping the .c file focused on the actual logic instead of dealing with macros. The impl header also needs to `#undef` and redefine several macros before re-expanding `plugin_def.h` a second time so that's ~60 lines of macro machinery that could clutter `plugin.c` if inlined. 

What do you think?

christf:
wow, thank you for this! :-)

navit/plugin_impl.h
```diff
@@ -0,0 +1,90 @@
...
+ * This header provides the macro definitions that expand PLUGIN_FUNC* and
+ * PLUGIN_CATEGORY into function bodies and global variable definitions.
+ * It must be included exactly once, after plugin.h, in plugin.c.
+ */
+
+#ifndef NAVIT_PLUGIN_IMPL_H
+#define NAVIT_PLUGIN_IMPL_H
+
+#include "plugin.h"
+#include <glib.h>
```
christf, line 33: I think including "glib.h" is better here, if we want to continue supporting platforms without native glib.

navit/plugin_impl.h
christf: I think, having it is fine and for my taste plugin.c still contains too many macros/defines (which is no way speaking to this PR but rather to the structure that is in place).

christf: I did notice there is still some kind of problem around plugin_def.h. If that one receives an include guard, linking fails. I have not been able to look into this further

PR #1513, opened 2026-03-18
Fix:docs:Route status is replaced by navigation status

adevoss:
This route_status example is replaced by navigation status. The latter is easier to use so why not delete this example?


christf:
I feel, if the application has a behavior, then it is good practice to have documentation for it.

If we all can agree that the route status is obsoleted by the navigation status (I am not sure, simply have not looked deeply yet), then I feel we can remove both: the feature and the documentation at the same time.
Also, I think that if we can find a feature that is not needed, we absolutely should remove it.

PR #1515, opened 2026-03-20
Patch:osd:debug statements and rearranged code

adevoss:
While figuring out variables for OSD in `navit.xml` I added debug statements and rearranged some code.
No functional changes.


adevoss:
#1522 includes this PR

PR #1517, opened 2026-03-23
Fix:docs:move icons from docs to icons directory

adevoss:


jkoan:
The moved files fall into two categories, either they are examples how a osd ison will look on screen or they a image elements from other themes, so the move does not make sense for the standard theme. 
But I definitely agree that filenames and location within the docs can be optimized.

PR #1523, opened 2026-03-30
Patch:core:Reformat LOC

adevoss:


jkoan:
@adevoss have you seen that your changes don't pass the formatting sanity check?

adevoss:
Yes I did. What could cause this?

The only thing I did is joining two lines of code to make it one line. The two lines are one statement.
The code compiles.

jkoan:
Because of the line length limit. The CI runs scripts/ci_sanity_checks.sh at first to check if the code style rules are met. The Codestyle rules are ensured with clang-format which has its configfile in navit/.clang-format (which gets automatically loaded when running clang-format)

adevoss:
Thanks. Then I will close this PR.

Issue #1132, opened 2021-07-26
Can't download map using Navit menu on Android


marcin-github:
Navit-0.5.5 from GPlay.
Since many weeks I can't download maps using builitn option. Download starts and after a couple of seconds transfer is stalled. Navit retries to download but it always finish after small amount of data.

mvglasow:
Yes, we are currently having issues with the map server. Unfortunately, the maintainer of the server seems to have left the project. @jkoan is working on a solution.

Currently your options are:
* look https://github.com/jkoan/gh-actions-mapserver (under Releases), maybe one of the map files there suits your needs
* A map of Europe is available as a bit torrent at magnet:?xt=urn:btih:29e637e2dd8e376c85e1abe3445606e7277db38f&dn=europe-210503.bin (19.49 GB in size); I am currently seeding

marcin-github:
Thank you for information

hoehnp:
As I feel a little

Issue #1112, opened 2021-04-21
Map download at Planet Extractor not possible, almost same errors as in issue #1072


martin-ac:
Hi, again like in end 2020 there are problems with the download of maps with Planet Extractor at "http://maps9.navit-project.org/" as well as "http://maps3.navit.project.org"
Partly I do not even get error messages, partly - from maps9 - i receive them in german language and translated them to english for this issue:

At "maps3.navit-project.org" (Map Version 210326)

for download of Europe -> WesternEurope -> Germany -> Northrheinwestfalia the error message is
xyz.bin.part couldn'd be saved because the source file couldn't be read.
Try again later or get in contact with the server administrator.

same for Germany 
same for Germany+Austria+Switzerland

for WesternEurope:
Internal Server Error
The server encountered an internal error or misconfiguration and was unable to complete your request.
Please contact the server administrator at webmaster@maps3.navit-project.org to inform them of the time this error occurred, and the actions you performed just before this error.

More information about this error may be available in the server error log.
Apache/2.4.29 (Ubuntu) Server at maps3.navit-project.org Port 80

At "maps9.navit-project.org" (Map Version 210326)

for download of Europe -> WesternEurope -> Germany -> Northrheinwestfalia
download doesn't start or stops after few (4) megabytes

same for Germany 
same for Germany+Austria+Switzerland

for WesternEurope:
Serverfailure!
The request can not be answered because of an internal error.
Errormessage:
End of script output before headers: index.cgi
If you think this is a server malfunction please inform the webmaster.
Error 500
maps9.navit-project.org
Apache/2.4.33 (Linux/SUSE) 

To me it looks like currently any download from both of the servers to not work no matter what size the maps have.
Does this error (always) occur only after may be faulty data in maps data updates? 
Maybe others have faced the same errors but do not know where to report this to because
the only information about the source of the web interface in the upper right corner is outdated:
it points to Tordt Schmidt / SMT-WebServices and Tordt answered to my mail that he doesn't do anything for the servers since more but several years!

Could please again someone take care of this? Many thanks in advance and please give me feedback as soon as I should test again!!!

Fussel09:
I have the same problem. I tried different browsers (Firefox with and without ad- and script blocker, old Chrome), but none of them works. The download fails instantly or after 4 to 8 MB.
Please help.

Budman1758:
For what its worth I just downloaded a map of the entire United States (6.3Gig) without any problem at all. 
Microsoft Edge browser.

[...]

PR #1525, opened 2026-04-05
fix:cURL based map download

### LINKED ISSUE #1311: Navit won't download maps on android
On navit android app I try to download maps but it's stuck it won't download maps is this a known issue with the navit android app.

hoehnp:
The code is a new iteration of my code from #1134. Since the code in #1134 was branched long ago, I restarted my work based on trunk with some further improvements, e.g. now Threaded parallel downloading of several maps. The PR should also fix #1311, #1132 and #1112.


CMakeLists.txt
```diff
@@ -531,6 +528,21 @@ if (GETTEXT_FOUND)
...
+    set_with_reason(USE_DOWNLOAD "libcurl found" TRUE)
+    list(APPEND NAVIT_LIBS "-lcurl")
+    SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -u download_map")
+    set(USE_DOWNLOAD 1)
+endif(CURL_FOUND)
+
+if (cJSON_FOUND)
+    set(HAVE_CJSON 1)
+    include_directories(${CJSON_INCLUDE_DIRS})
+    list(APPEND NAVIT_LIBS ${CJSON_LIBRARIES})
```
jkoan: Same here: target_link_libraries

hoehnp:
@jkoan: Do we want to also modify the Android code to run the functions in here or do we do this in another PR?

navit/network.c
```diff
@@ -0,0 +1,643 @@
+#include "curl/curl.h"
```
jkoan: Should be a system header include, not local.

navit/network.c
```diff
@@ -0,0 +1,643 @@
...
+#include <cjson/cJSON.h>
+#include <math.h>
+#include <string.h>
+#include <navit/debug.h>
```
jkoan: Local not system include

navit/gui/internal/CMakeLists.txt
```diff
@@ -2,6 +2,11 @@ if (GUI_INTERNAL_VISUAL_DBG)
 	add_definitions ("-DGUI_INTERNAL_VISUAL_DBG")
 endif(GUI_INTERNAL_VISUAL_DBG)
 
-module_add_library(gui_internal gui_internal.c gui_internal_bookmark.c gui_internal_command.c gui_internal_gesture.c
+if(USE_DOWNLOAD)
+	set(map_downloader_sources gui_internal_map_downloader.c )
+endif(USE_DOWNLOAD)
+
+
+module_add_library(gui_internal ${map_downloader_sources} gui_internal.c gui_internal_bookmark.c gui_internal_command.c gui_internal_gesture.c
```
jkoan, line 10: map_downloader_sources can be empty in this case. It would be cleaner if we would add a variable for all gui_internal files and add downloader to if when enabled

navit/gui/internal/gui_internal_map_downloader.c
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * @returns nothing
+ */
+
+void gui_internal_download_update(struct gui_priv * this) {
+    dbg(lvl_debug, "downloading status = %i\n", dl_info.downloading);
+    if(dl_info.downloading == 1) {
+        event_add_timeout(500, 0, callback_new_1(callback_cast(this->download_cb), this));
+    } else {
+        //if(pthread_join(dl_thread, NULL)) {
```
jkoan, line 87: cleanup dead code starting here

navit/network.c
```diff
@@ -0,0 +1,643 @@
...
+#include <glib.h>
+#include "unistd.h"
+#include "stddef.h"
+#include "navit.h"
+#include <navit/coord.h>
+#include <navit/color.h>
+#include <navit/item.h>
+#include <navit/point.h>
```
jkoan: Local not system include

navit/network.c
```diff
@@ -0,0 +1,643 @@
...
+#include <pthread.h>
+#include "gui/internal/gui_internal.h"
+#include "gui/internal/gui_internal_map_downloader.h"
+#include "gui/internal/gui_internal_widget.h"
+#include "gui/internal/gui_internal_priv.h"
+#include "network.h"
+
+#include "sys/stat.h"
```
jkoan, line 21: newline makes a second import groups, remove unless there's a reason to do so in multiple groups

navit/network.h
```diff
@@ -0,0 +1,40 @@
...
+} MapData;
+
+int progressCallBack(struct map_download_info *dl_info, double dltotal, double dlnow, double ultotal, double ulnow);
+void add_to_listdata(char *name, long long size, GArray *maps_data, int pos); 
+long long get_size(char *key, int* maps_count, MapData *maps_size);
+size_t write_callback(void *contents, size_t size, size_t nmemb, void *userp);
+void format_filesize(long long map_size, char *buffer);
+void * download_map (struct gui_priv *this, struct widget *wm, struct map_download_info * dl_info);
+void * download_map2 (void * data);
+void * download_map3(void * data);
```
jkoan: Bad naming of those functions, what should they do?

navit/network.c
```diff
@@ -0,0 +1,643 @@
...
+ *
+ * @param dl_info a pointer to a map_download_info where the download updates will be stored
+ * @param dltotal number of total bytes to download, required by libcurl
+ * @param dlnow number of bytes currently downloaded, required by libcurl
+ * @param ultotal number of total bytes to upload, required by libcurl
+ * @param ulnow number of bytes currently downloaded, required by libcurl
+ *
+ * @returns Returns 0 by default, until we implement 'cancel'
+ */
+
```
jkoan, line 35: remove empty line

navit/network.c
```diff
@@ -0,0 +1,643 @@
...
+void * download_map3(void * data) {
+    // check name
+    // if name directly in list, skip it and only take elements which are not part of others
+    // build download list
+    // download list one by one
+    // make dl_info and call download_map2
+
+    char buffer[256];
+    FILE *f;
+    GThreadPool *pool = g_thread_pool_new(
```
jkoan, line 168: out support/glib does not have this, so it will fail on all targets which use it instead of system glib

navit/network.c
```diff
@@ -0,0 +1,643 @@
...
+    cJSON *assets = NULL;
+    cJSON *asset = NULL;
+
+    response = malloc(1);
+    response[0] = '\0';
+
+    curl_global_init(CURL_GLOBAL_DEFAULT);
+    curl = curl_easy_init();
+    if(curl) {
+        curl_easy_setopt(curl, CURLOPT_URL, "https://api.github.com/repositories/384098365/releases/latest");
```
jkoan, line 293: use https://api.github.com/repos/gh-actions-mapserver/navit/releases/latest its easier to understand

navit/network.c
```diff
@@ -0,0 +1,643 @@
...
+        curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, write_callback);
+        curl_easy_setopt(curl, CURLOPT_WRITEDATA, &response);
+        curl_easy_setopt(curl, CURLOPT_USERAGENT, "libcurl-agent/1.0");
```
jkoan, line 296: change it to something like navit-downloader/1.0

navit/gui/internal/gui_internal.c
```diff
@@ -1724,63 +1707,42 @@ void gui_internal_cmd_map_download(struct gui_priv *this, struct widget *wm, voi
     on.type = off.type = attr_active;
     on.u.num = 1;
     off.u.num = 0;
-    wb = gui_internal_menu(this, wm->name ? wm->name : _("Map Download"));
-    w = gui_internal_box_new(this, gravity_top_center | orientation_vertical | flags_expand | flags_fill);
-    w->spy = this->spacing * 3;
+    wb = gui_internal_menu(this, "Map Download");
```
jkoan: why remove the translated String?

navit/network.c
```diff
@@ -0,0 +1,643 @@
...
+            add_to_listdata(continent, size_continent, menu_data, pos);
+            size_planet += size_continent;
+        }
+    }
+
+    // Add Planet
+    add_to_listdata("planet", size_planet, menu_data, 0);
+
+    // --- Output to TSV ---
+    FILE *tsv2 = fopen("maps/menu2.tsv", "w");
```
jkoan, line 561: i really don't like the idea that you write a file into the filesystem to read it later, this dosn't make sense

navit/network.c
```diff
@@ -0,0 +1,643 @@
...
+                if(key[k] == '-') key[k] = '_';
+            }
+
+            maps_size[maps_count].name = key;
+            maps_size[maps_count].size_val = size->valueint;
+            maps_count++;
+        }
+    }
+
+    // --- Define Lists ---
```
jkoan, line 351: starting here the functions dosn't make sense. What's that supposed to do. You already have your List from the release artifacts, just use them directly

navit/gui/internal/gui_internal.c
```diff
@@ -1724,63 +1707,42 @@ void gui_internal_cmd_map_download(struct gui_priv *this, struct widget *wm, voi
...
     }
 
     download_enabled.type = download_disabled.type = attr_update;
     download_enabled.u.num = 1;
     download_disabled.u.num = 0;
-    wma = gui_internal_button_map_attr_new(this, _("Download Enabled"),
-                                           gravity_left_center | orientation_horizontal | flags_fill, map,
-                                           &download_enabled, &download_disabled, 0);
-    gui_internal_widget_append(w, wma);
+    f = fopen("maps/menu2.tsv", "r");
```
jkoan, line 1722: again, no

navit/gui/internal/gui_internal.c
```diff
@@ -1724,63 +1707,42 @@ void gui_internal_cmd_map_download(struct gui_priv *this, struct widget *wm, voi
...
+       
+        char* buffer2 = g_strdup_printf("%s", g_strjoin(NULL, g_strsplit(buffer,"\t",0)[0], " ", g_strsplit(buffer,"\t",0)[1], NULL));
+        wl=gui_internal_button_new_with_callback(this, buffer2, NULL,
+                    gravity_left_center|orientation_horizontal|flags_fill, gui_internal_cmd_map_download_do, dl_info);
+
+	gui_internal_widget_append(row, wl);
+	g_free(buffer2);
+
     }
+    error = pthread_create(&download, NULL, update_download_table, NULL);
```
jkoan, line 1744: this results in an empty page for the first load, where the uses sees a bland screen only one the user reenters the same screen again theirs something. Need to add "Loading..." and auto reload.

jkoan (changes requested):
When testing there where a lot of design flaws and crashes. 
Pleas look at how the code should work and see what all the functions should do. It looks like multiple functions are iterations of one another.

In my idea there would only be one functions which loads a direct link from github (the url https://api.github.com/repos/navit-gps/gh-actions-mapserver/releases/latest) and all other Links are generated from there, so you have something like a object with a name, link and size which gets stored in a list, displayed on the downloader overview and used for the curl download. No code that should probably combine continents/regions, no convoluted file which gets generated, but not in time, ...
Just [KISS](https://en.wikipedia.org/wiki/KISS_principle)

Note: I made comments on lines where i saw ting, but those are not comprehensive, as i just later saw what i wrote above

navit/network.c
hoehnp: why should be cjson local?

navit/gui/internal/gui_internal_map_downloader.c
hoehnp: certainly I will still clean up dead code.

navit/network.c
hoehnp: if we do not write to the file system, we have issues, in case of no internet connection.

navit/network.c
jkoan: this comment was for navit/* not the others

navit/network.c
jkoan: when we don't have a internet connection we cant download either, so this is fine. But sure we should also add a timeout and "Could not download index" or something

hoehnp: my answer would be that still the menu should not be empty. I can only achieve this by caching the data on file system. Earlier, if i remember correctly, we hard coded the numbers in the source.

navit/network.c
hoehnp: then the quoting is pretty misleading.

navit/network.c
hoehnp: Indeed, in the Android Code we even have the file sizes hard coded in the Java code: https://github.com/navit-gps/navit/blob/94b70ea5991f3dee8d974373fb20a10fa33991bf/navit/android/src/org/navitproject/navit/NavitMapDownloader.java#L71

@jkoan: Is this the better solution?

navit/network.c
hoehnp: I need those lists to aggregate the data for the different countries, continents, planet.

navit/gui/internal/gui_internal.c
hoehnp: the empty page on first load is an issue which should still be resolved, I agree.

navit/gui/internal/gui_internal.c
hoehnp: actually I temporary disabled it for further debugging, since the original line results in only "." being displayed.

PR #1531, opened 2026-04-13
Chore:Docs(itemgra): item type list displayed as list

christf:
This fixes the formatting for the list of itemgra options


christf:
see https://navit--1531.org.readthedocs.build/en/1531/user/configuration/basic/layout.html for the page that contains the fix.

PR #1542, opened 2026-05-10
Feat:core:Spoken destination

adevoss:
When calculation of the route starts the destination is spoken.

Help wanted:
- [x] Add "destination" to translations
- [x] Add "set to" to translations
- [ ] Destination is spoken before map is shown. Would be nice if that was the other way around
  Works for setting a new destination
  Does not work for a previous set destination (crash)


christf:
I tried this tdoay, it works for me. To make any announcement, I am running a script that executes the playback command in the background. For this reason the map is shown before the announcement for me.

adevoss:
What did you try? This PR or something else?
In the PR there is no script to execute in the background.

navit/navit.c
```diff
@@ -1605,9 +1606,17 @@ static void navit_mark_navigation_stopped(char *former_destination_file) {
...
+    char *destination_announcement;
+
     if (c) {
         this_->destination = *c;
         this_->destination_valid = 1;
+        this_->destination_description = g_strdup(description);
+
+        // Announce destination description
+        destination_announcement =
+            g_strdup_printf("%s %s %s", _("destination"), _("set to"), this_->destination_description);
```
jkoan: Free missing

christf:
I ran the code of this PR with my configuration from the navit xml that calls a shell script which executes the tts system of my choice and then  runs aplay in the background. The shell script terminates immediately, while speech synthesis and playback is still running. Then the map renders after that and then the synthesis is done and audio is emitted. Worked beautifully.

adevoss:
> navit xml that calls a shell script which executes the tts system of my choice and then runs aplay in the background.

Do you mean in `navit.xml` you have a speech tag like `<speech data="/path/to/script.sh %s" ... />` instead of `<speech data="espeak-ng %s ...">` ?
In the script an audio file is generated which has the name of the instruction (i.e. "go left on main street.wav") and then is played by aplay.

Is this correct?
I would like to test your approach.

christf:
> > navit xml that calls a shell script which executes the tts system of my choice and then runs aplay in the background.
> 
> Do you mean in `navit.xml` you have a speech tag like `<speech data="/path/to/script.sh %s" ... />` instead of `<speech data="espeak-ng %s ...">` ? In the script an audio file is generated which has the name of the instruction (i.e. "go left on main street.wav") and then is played by aplay.
> 
> Is this correct? I would like to test your approach.

this almost exactly. I do not use espeak-ng but piper-tts

it executes this:
model=/home/$USER/tts/de_DE-thorsten-high.onnx

piper -m ${model} -c ${config}  --output-raw <<< "$@" | aplay -r 22050  -c 1 -f S16_LE &
disown
exit

adevoss:
Installed piper-tts. Works great for generating audio files. Generating takes about 8.5 seconds on a Pi 4B and 7.5 seconds on a Pi 5.
Too long, in my case, to use with Navit.

Used piper-tts to generate the list of audio files (#1498) and now I can select an extra voice (#1299).

christf:
We'll do something about those delays when implementing #1577  :-)

navit/navit.c
```diff
@@ -1626,9 +1627,17 @@ static void navit_mark_navigation_stopped(char *former_destination_file) {
...
```
jkoan:
```suggestion
            g_strdup_printf(_("destination set to %s"), this_->destination_description);
```

This makes it better to translate than two strings, as in other languages it may be that the word order is different

navit/navit.c
```diff
@@ -1881,6 +1896,13 @@ static void navit_add_former_destinations_from_file(struct navit *this_) {
             route_set_destinations(this_->route, pc, count, 1);
         this_->destination = pc[count - 1];
         this_->destination_valid = 1;
+        this_->destination_description = g_strdup(active_destination_description);
+
+        // Announce destination description
+        destination_announcement =
+            g_strdup_printf("%s %s %s", _("destination"), _("set to"), this_->destination_description);
```
jkoan:
```suggestion
            g_strdup_printf(_("destination set to %s"), this_->destination_description);
```

navit/navit.c
```diff
@@ -1626,9 +1627,17 @@ static void navit_mark_navigation_stopped(char *former_destination_file) {
...
+
     if (c) {
         this_->destination = *c;
         this_->destination_valid = 1;
+        this_->destination_description = g_strdup(description);
+
+        // Announce destination description
+        destination_announcement =
+            g_strdup_printf("%s %s %s", _("destination"), _("set to"), this_->destination_description);
+        navit_say(this_, g_strdup(destination_announcement));
```
jkoan, line 1639:
```suggestion
        navit_say(this_, g_strdup(destination_announcement));
        g_free(destination_announcement);
```

navit/navit.c
```diff
@@ -1881,6 +1896,13 @@ static void navit_add_former_destinations_from_file(struct navit *this_) {
             route_set_destinations(this_->route, pc, count, 1);
         this_->destination = pc[count - 1];
         this_->destination_valid = 1;
+        this_->destination_description = g_strdup(active_destination_description);
+
+        // Announce destination description
+        destination_announcement =
+            g_strdup_printf("%s %s %s", _("destination"), _("set to"), this_->destination_description);
+        navit_say(this_, g_strdup(destination_announcement));
+
```
jkoan, line 1904:
```suggestion

        g_free(destination_announcement);
```

navit/navit.c
```diff
@@ -1881,6 +1896,13 @@ static void navit_add_former_destinations_from_file(struct navit *this_) {
...
         g_free(pc);
     }
```
jkoan, line 1907:
```suggestion
    }
    g_free(active_destination_description);
```

jkoan (changes requested):
Also po files need to be regenerated

adevoss:
When using a separate thread for the audio the map is shown (calculating route) and the destination is spoken at the same time (see first comment).

PR #1544, opened 2026-05-13
Patch:core:Update navit.dtd with existing attributes of speech element

adevoss:
These attributes are used in code but missing from `navit.dtd`.


adevoss:
There is another attribute: active.

Issue #1549, opened 2026-06-06
Get navit ready for EV-navigation


christf:
this needs two things:

* charging points and their meta data available as POI
* navigating to a POI "along the way within x km"

there may be more.

jkoan:
Another thought, add file type to vehicle. Also add fule tank capacity for the routing engine. This could even be something like Ah for electric cars, kcal for bicycle or l/km for gas based cars, so depending on the fule type the fill-up station can automatically be selected

christf:
for this, I feel there should be a way to obtain updates on the fuel state from the vehicle. For example, when going downhill, EV will charge the battery.

PR #1550, opened 2026-06-06
feat: Charging_Station display on map and searchable as POI

christf:
This parses OSM data to put charging stations for electric vehicles onto the map. It also introduces another menu item in the poi search. Charging stations can be searched with poi search

There are a few questions for the design we might want to consider:
* Do we want this to handle Electricity as a fuel type instead? If so, should we filter this separately?
* Electric vehicle drivers will likely want to filter for more details, including:
  * Maximum Charging Current
  * plug format (Chademo / CCS /...)
  * maybe even price
* do we need this in more GUIs?

I am not sure what a good design would be to make these accessible, so.... happy to have your inputs. In the meantime, this helps with #1549


christf:
Is red the right color or should this be blue like a gas station?

christf:
Thinking about this some more, I wonder if there is benefit of having slower chargers on the map. Maybe we should just add the ones that we know can deliver more than 200KW?
Maybe give them different colors or other indicators?

jkoan:
I think it depends on the use case on a long road trip you probably want the fastest charger possible (which your car supports). On a comutea slow charger where you can park and load your car the whole day while at work is probably more appropriate. So I think we need both.
For the road trip case we could probably even hook up the loading speed into the route calculation. Like if I take a detour to a charger that's twice the power will I save time compared to the one directly on the route?

christf:
so, having separate filters is the way to go then? (slow chargers / fast chargers)?

navit/gui/internal/gui_internal_poi.c
```diff
@@ -94,6 +95,7 @@ static enum item_type selectors_OtherTypes[] = {type_point_unspecified,
                                                 type_poi_land_feature - 1,
                                                 type_poi_rock + 1,
                                                 type_poi_fuel - 1,
+                                                type_poi_charging_station - 1,
```
jkoan: this is wrong. Looking at https://github.com/christf/navit/blob/4b3e4ded18ca63d3e50f19bf1d82fa5388e3b61c/navit/gui/internal/gui_internal_poi.c#L431-L434 i works like the even numbered types are the start and the odd ones are the end.

navit/gui/internal/gui_internal_poi.c
```diff
@@ -39,7 +39,8 @@ struct selector {
     enum item_type *types;
```
jkoan:
```suggestion
    enum item_type *types; // even items are start, odd ones are end of selection
```

navit/gui/internal/gui_internal_poi.c
```diff
@@ -39,7 +39,8 @@ struct selector {
...
 };
 static enum item_type selectors_BankTypes[] = {type_poi_bank, type_poi_bank, type_poi_atm, type_poi_atm, type_none};
-static enum item_type selectors_FuelTypes[] = {type_poi_fuel, type_poi_fuel, type_none};
+static enum item_type selectors_FuelTypes[] = {type_poi_fuel, type_poi_fuel, type_poi_charging_station, type_none};
```
jkoan:
```suggestion
static enum item_type selectors_FuelTypes[] = {type_poi_fuel, type_poi_fuel, type_none};
```

navit/gui/internal/gui_internal_poi.c
jkoan:
```suggestion

```

jkoan: Add 

```
type_poi_charging_station - 1,
type_poi_charging_station + 1,
```
after `type_poi_atm + 1,`

navit/icons/charging_station.svg
jkoan, line 1: doesn't fit the style of the other icons:

Image

maybe just the fuel symbol but with a yellow lightning bolt?

navit/gui/internal/gui_internal_poi.c
jkoan: Fuel should be fossil fuel only to not mix up the variants

navit/icons/charging_station.svg
christf: done.

christf:
pushed fixup commits so it is easier to see what is new.

jkoan (approved):

PR #1555, opened 2026-06-15
fix(#1323): For Germany, allow town search, utilize amtlicher_gemeind…

christf:
…eschluessel to assign country

This adds german towns to the search index properly without hard-coding. Tested on Niedersachsen and Sachsen-Anhalt.


christf:
I am not sure if this should close the draft PR for hard-coding the country #1324 
While that one is universal but requires manual intervention and will put towns into the wrong country, if the cutout of the larger map is unfortunate, this one is accurate but only works for Germany.

jkoan (approved):
LGTM!

PR #1556, opened 2026-06-16
Allow to specify language of labels displayed on map and available in town search

### LINKED ISSUE #1325: localized maps

christf:
This does two things:

* adjust maptool to place all language-specific names inside the bin maps
* adjust navit to be able to work with these language-specific names according to language preference that get to be specified like this lang_pref="en,de,native" for preferring english names over german names over whatever native language the map has.

This allows somewhat localized maps (depending on translation data).

Closes #1326 
Closes #1325


navit/attr_def.h
```diff
@@ -408,6 +408,15 @@ ATTR(exit_to)
 ATTR(street_destination_forward)
 ATTR(street_destination_backward)
 ATTR(outputdir)
+ATTR(label_l10n)
+/**
```
jkoan: Reminder for myself: lookup if this comment shifts internal IDs

navit/maptool/osm.c
```diff
@@ -125,6 +125,87 @@ enum attr_strings_type {
...
+        if (n < 0) {
+            itembin_warning(ib, 0, "name:%s translation could not be formatted, dropping\n", e->lang);
+            continue;
+        }
+        if ((size_t)n >= sizeof(buf)) {
+            itembin_warning(ib, 0, "name:%s translation is %d bytes, exceeds %zu-byte limit, dropping\n", e->lang, n,
+                            sizeof(buf) - 1);
+            continue;
+        }
+        item_bin_add_attr_string(ib, attr_label_l10n, buf);
```
jkoan: Maybe we should add this to the special types instead of as a raw sting? 
Or we put the array directly into the label (and others). When we write the string like this it should automatically work for old navit versions:
"native\0de:abc\0en:cde\0...\0" unaware navit versions would then only see native.

navit/attr_def.h
christf: I'll remove it. if we write a comment like this for every attr, then the file will become unreadable.

christf:
When pondering whether to include all languages, I worried, whether the bins would get too large.
For the three maps (greece, saxony-anhalt, lower-saxony) it did not make a measurable difference to include all name tags or just the native one.

navit/maptool/osm.c
christf: I have not seen a problem when running a new map with an unaware navit. I remember a warning but that was it. I am not sure myself of the best datastructure here.
The labels are used a lot, so something efficient in access would probably be good. Beyond this ... I am John Snow.

christf: So, I have thought about this and tested 2 versions of navit with a map of niedersachsen that was built by a maptool with this patch. Test 1 was navit with the patch and lang_pref="en,de,native" - it shows "Brunswick" for Braunschweig on the map and I can search for "Braunschweig" as well as "Brunswick".

Test 2 was running navit from the asan branch (which does not have the l10n patch). It shows the german name and ignores the english one, but it allows to find Braunschweig by its english name. Fascinating.

This is what we want, right? What would be the benefit of changing the layout of the attribute as above? One disadvantage I could see: It is unclear, where this array actually ends, so we would have to work with some sort of sentinel or double \0, which is entirely possible but I think we'd have to have a great reason to do that kind of thing.

navit/attr_def.h
christf: there is another comment a few lines up the file.... if the comment shifts the internal IDs, we'd already be in trouble....

navit/maptool/osm.c
christf: > Maybe we should add this to the special types instead of as a raw sting? Or we put the array directly into the label (and others). When we write the string like this it should automatically work for old navit versions: "native\0de:abc\0en:cde\0...\0" unaware navit versions would then only see native.

Coming back to another aspect of this: "native" seems redundant.native is the fallback always.

christf: "native" was removed and is now default if neither of the more specific language can be found

christf:
this displays all languages when searching, which is a bit confusing. Let me look at that some more.

navit/maptool/osm.c
christf: as discussed on discord, we are not certain which of the approaches is really better and agreed to leave the datastructure as is for now.

jkoan (commented):
Looks good so far, i still need to actually test this

christf:
reverting to draft to communicate that I am still working on codefactor. 
Also I would like to investigate the order in which search results are being presented to the user.

christf:
Since I wasn't sure how far along review is, I pushed some changes as fixup commits. The docs needed an upgrade because the recent cleanup had caused conflicts with this PR.

PR #1558, opened 2026-06-17
This adds a few fixes found after enabling asan on maptool and internal gui and gtk gui

christf:
This allows the build to complete and run of internal gui and gtk gui when ASAN is active and fixes all problems identified by asan. There is a change to the shutdown behavior to tear down and free all resources in an effort to make asan output contain less false positives, so it is more useful in the future.

It stands up to my tests:
* running maptool using this patches on niedersachsen and sachsen-anhalt map (I added the amtlicher_gemeindeschluessel commit for this testing because I wanted the search-ability for the other verifications)
* running navit with the patches using the maps generated by maptool from previous step
* searching an address, routing to it
* panning
* for gtk and internal gui

With these scenarios and navit linking in asan, no crashes (no oob operations, no overflows) were observed. There are still leaks, but they seem to be outside of navits scope. Also, this PR already is quite large.


navit/start_real.c
```diff
@@ -232,13 +232,11 @@ int main_real(int argc, char *const *argv) {
     }
     event_main_loop_run();
 
-    /* TODO: Android actually has no event loop, so we can't free all allocated resources here. Have to find better
```
jkoan, line 235: Have you tested android so you can remove this?

jkoan (commented):
except the one question everything looks really good!

navit/start_real.c
christf: Good question, I was debating to remove it or not myself too. No, I did not test whether cleanup works on android (I have no android device).
In the end, I opted for the removal because I found that often TODO comments never get actioned. Github issues / problem with visible symptoms get actioned. So, this task may have been important to someone in the past, but since it has not been implemented in years, maybe it is now ok to remove it. But I won't argue about it. If you want to keep it, I will put it back in.

Here is my pro/contra list:

pro removal:
* the comment bears no consequence but it adds noise to the codebase (I am firmly in the clean code camp).
* There is no gh issue about it. If anyone was unhappy with resource cleanup, we'd probably know

contra removal:
* someone might come and implement it after seeing that TODO.

navit/xmlconfig.c
```diff
@@ -674,6 +674,7 @@ static void start_element(xml_context *context, const gchar *element_name, const
         else
             parent_attr = &new->parent->element_attr;
         new->element_attr.u.data = new->object_func->create(parent_attr, attrs);
+        attr_list_free(attrs);
```
christf: there is a problem here - I think we should force vehicle plugins to dup thisor hand over ownership to the vehicle plugins when loading instead of referencing to it. (Or we could live with the leak but I am not a fan of that).... will need to do some more work on this.

navit/start_real.c
jkoan: The comment in itself is not what i meant, but the `#ifndef HAVE_API_ANDROID` which just doesn't do the cleanup on android which matches the comment. I tested it in the emulator within Android Studio and indeed navit crashes now. 
One idea would be to remove the while cleanup from here and handle it inside the event loop, as android doesn't have a real one it will not be called, resulting in a working android app. But we should implement a android event loop as well.

christf: well, that needs to be fixed. What would make the solution "move to eventloop" superior to retaining the ifdefs here?

christf:
apologies for the large-ish rebase - fixing that problem I pointed out took some re-ordering of commits and a bit new code. Reverting to draft for now until I think I will have something presentable.

navit/start_real.c
jkoan: this would actually clean it up on android as well ;-) But i tested that addinf the ifdef back brings android back in working order, so lets do this for now

jkoan (approved):

PR #1565, opened 2026-06-29
feat(navit): display map while panning via prefetched wide display list

### LINKED ISSUE #1354: wish: redraw map when panning
title says it all. It would be great if panning would cause the map to be redrawn to increase UX.

christf:
This commit prefetches a display list that is PAN_PREFETCH_SCALE_FACTOR (default 3) times the screen size in each dimension, centred on the current view. The wider list means the user can drag by up to one full screen width in any direction before the revealed edge has no data.

closes #1354

### DIFF
--- a/navit/navit.c
+++ b/navit/navit.c
@@ -77,6 +77,10 @@ struct vehicle;
 
 /* define string for bookmark handling */
 #define TEXTFILE_COMMENT_NAVI_STOPPED "# navigation stopped\n"
+/* How many times wider/taller than the screen the map source rectangle is made when building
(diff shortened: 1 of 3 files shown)


christf:
I have tried this on my phone and on my laptop computer, both running linux. This works when drag_bitmap=0.

When drag_bitmap = 1, this will not change the behavior of navit because the amount of data rendered does not change with this PR. This might be considered incomplete, I consider it to be a good thing as it provides a way to run navit on older hardware - those will be the ones with drag_bitmap =1.

PR #1579, opened 2026-07-03
Map Compression with LZMA for size benefit

christf:
This allows us to use LZMA in maptool and navit, utilizing multiple cores during compression.

Despite the multi-core aspect, this about doubles the runtime of maptool for me for a size gain of 37%


navit/maptool/misc.c
```diff
@@ -300,15 +304,189 @@ int phase4(FILE **in, int in_count, int with_range, char *suffix, FILE *tilesdir
...
+    worker_count = n_threads;
+    tile_queue = g_async_queue_new();
+    tile_done_queue = g_async_queue_new();
+    worker_threads = g_malloc0(sizeof(struct tile_process_thread) * n_threads);
+    for (i = 0; i < n_threads; i++) {
+        worker_threads[i].queue = tile_queue;
+        worker_threads[i].compression_method = compression_method;
+#ifdef HAVE_LZMA
+        if (compression_level > 0) {
+            uint64_t mem = lzma_easy_encoder_memusage(compression_level);
```
jkoan: why is lzma hardcoded here. My understanding is that this method should be used for both methods, hence the compression_method parameter.

navit/maptool/maptool.c
```diff
@@ -526,11 +536,19 @@ static int parse_option(struct maptool_params *p, char **argv, int argc, int *op
     case 'x':
         p->max_index_size = atoi(optarg);
         break;
-#ifdef HAVE_ZLIB
+    case 'C':
+        if (!strcmp(optarg, "lzma")) {
+            if (LZMA_AVAILABLE)
+                p->compression_method = 14;
```
jkoan: Please move both magic values to a define, enum or something.

CMakeLists.txt
```diff
@@ -348,6 +349,11 @@ else(ZLIB_FOUND)
 	message(STATUS "using internal zlib")
 	set_with_reason(support/zlib "native zlib missing" TRUE)
 endif(ZLIB_FOUND)
+if(LIBLZMA_FOUND)
+	set(HAVE_LZMA 1)
+	include_directories(${LIBLZMA_INCLUDE_DIR})
```
jkoan, line 354: maybe use target_include_directories?

navit/maptool/zip.c
```diff
@@ -147,8 +245,78 @@ void write_zipmember(struct zip_info *zip_info, char *name, int filelen, char *d
         dbg_assert(fwrite(&cd_ext, sizeof(cd_ext), 1, zip_info->dir) == 1);
         zip_info->dir_size += sizeof(cd_ext);
     }
+    if (comp_data != data)
+        g_free(comp_data);
+    g_free(reuse_buf);
+}
+
+void write_zipmember_raw(struct zip_info *zip_info, char *name, int filelen, char *compressed_data, int compressed_size,
```
jkoan, line 236: this function is nearly 1:1 copy of write_zipmember but with compression done externally, could we slim down write_zipmember so it calls write_zipmember_raw when its done compressing?

jkoan (changes requested):

CMakeLists.txt
christf: I am not sure this PR is the right place. We want lzma to be linked into every binary we build, same as zlib,which also uses include_directories and png. So this just continues the existing structure.

target_include_directories is the right step in general, imho - but then we should rework the whole CMakeList.

Let's do this in another PR in which we can switch to target_include_directories, target_link_libraries, lean more into find_package and use that to build the includes... maybe more - happy to discuss.

navit/maptool/zip.c
christf: sorry about that,

navit/maptool/misc.c
christf: I had to think about that one a bit because at first I thought this would not actually be a problem because of that ifdef HAVE_LZMA.
However, even in the zlib case, this lzma-code is being called if general lzma support is baked in, so, yes indeed this needs a change. Will do on next push.

navit/maptool/maptool.c
```diff
@@ -928,9 +940,10 @@ int main(int argc, char **argv) {
     linguistics_init();
 
     memset(&p, 0, sizeof(p));
-    p.zip64 = 1; /* default to 64 bit zip */
-#ifdef HAVE_ZLIB
-    p.compression_level = 9;
+    p.zip64 = 1;                                    /* default to 64 bit zip */
+    p.compression_method = ZIP_COMPRESSION_DEFLATE; /* default to zlib */
+#if defined(HAVE_ZLIB) || defined(HAVE_LZMA)
+    p.compression_level = 6;
```
christf, line 957: I am not entirely sure if we want it done that way.
Using lzma level 9 is hardly providing much space savings above level 6 but it does increase the time of processing significantly. For zlib, again, space difference is also small.

At the same time, doing different default for zlib / lzma does not seem right. The proposal here is to switch for both, just putting a lantern on it so it is a conscious choice.

navit/maptool/misc.c
christf: I think I have resolved this now. If lzma capabilities are baked into the binary, there is code added that checks whether the current compression method is lzma and in this case sets up the arena allocator.

And it sits in that function because it must be set up before starting the worker threads.

navit/map/binfile/binfile.c
```diff
@@ -321,13 +321,15 @@ static unsigned char *binfile_read_content(struct map_priv *m, struct file *fi,
...
-    case 8:
+    case ZIP_COMPRESSION_DEFLATE:
+        /* fallthrough */
+    case ZIP_COMPRESSION_LZMA:
         offset += lfh->zipxtraln;
-        ret = file_data_read_compressed(fi, offset, lfh->zipsize, lfh->zipuncmp);
+        ret = file_data_read_compressed_method(fi, offset, lfh->zipsize, lfh->zipuncmp, lfh->zipmthd);
         break;
     default:
         dbg(lvl_error, "map file %s: unknown compression method %d", fi->name, lfh->zipmthd);
```
jkoan: Here we handle one case and two others in the compressed function, wouldn't it be cleaner to handle all in one place? This way we would also avoid the two swich-case and the fallthrough

christf: of course. Somehow I must have missed that bit. I pushed the change as a single fixup commit for later squash - and also removed the now unused file_Data_read_compressed() function.

christf:
pushed a fixup commit for easier review of the change. Happy to squash once you agree to the approach of the refactor.

PR #1585, opened 2026-07-16
Doc dedup

christf:
This consolidates duplicate documentation for developers.


docs/user/community/Main_Page_support_box.rst
jkoan: File can be deleted, it was a include in the wiki and is not needed anymore

docs/user/community/Reporting_Bugs.rst
jkoan, line 1: Can't be found from the Menu, only by links. We should add it to the menu

docs/user/community/contacts.rst
```diff
@@ -3,103 +3,54 @@
...
-generally hang out.
+We are on Discord. In addition to the text-chats we also have some Broadcast
+Channels where we share relevant information like new releases.
 
--  irc://irc.freenode.org/navit
-   `WebChat <http://webchat.freenode.net/?channels=navit>`__ `IRC
-   Log <http://irclogs.navit-project.org/>`__
 
-
-Rules
```
jkoan, line 38: Maybe leave rules 1&2? It's relevant for discord as well

jkoan (changes requested):

docs/user/community/Reporting_Bugs.rst
christf: I thought so too at first. I did not add it on its own because there is a "reporting bugs" section on the contacts page, which is on the menu and this section then links to the Reporting_Bugs.rst.

When I looked at this, I considered, putting the content of Reporting_Bugs.rst into the contacts-page, but didn't do so because the contacts page should be about "how do I get in contact" and the mechanics of reporting a bug is somewhat different and quite verbose (180LOC for reporting bugs) in comparison to contacts (55 LOC).

Now, I have layed out the reasoning for why I chose this path, let me also say that I feel in no way attached to it. If you want this changed, let's see how. As far as I can see, in this case, just adding Reporting_Bugs to the toctree is not going to be the only change and I'd need some more guidance about what you have in mind.

docs/user/community/contacts.rst
christf: I have taken the liberty to slightly reword the rules without changing the meaning. Also, I thought this might be better if it is not just within the discord section but it is probably relevant to all communication channels, therefore I gave it a place in the toc next to discord and github.

christf:
ready for another review. I rebased this, removed the main page and brought back the rules as per review comments. For the third review comment I am not sure what to do.

docs/user/community/Reporting_Bugs.rst
jkoan: My intention is actually just to have it appear in the toctree so I doesn't need to be stumbled upon. The question is where if should go. I suggest right below Submitting Translations

christf: done.

jkoan (approved):

PR #1595, opened 2026-07-26
Stability

christf:
This adds the next batch of stability fixes I found while replaying nmea tracks, running with asan and watching out for assertions.

### DIFF
--- a/navit/coord.c
+++ b/navit/coord.c
@@ -85,10 +85,14 @@ void coord_rect_destroy(struct coord_rect *r) {
 }
 
 int coord_rect_overlap(struct coord_rect *r1, struct coord_rect *r2) {
-    dbg_assert(r1->lu.x <= r1->rl.x);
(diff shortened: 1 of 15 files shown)


navit/coord.c
```diff
@@ -85,10 +85,14 @@ void coord_rect_destroy(struct coord_rect *r) {
 }
 
 int coord_rect_overlap(struct coord_rect *r1, struct coord_rect *r2) {
-    dbg_assert(r1->lu.x <= r1->rl.x);
```
jkoan, line 88: Why is dbg_assert bad? :D

christf: I do not appreciate navit being terminated when I'd rather find a route to wherever it is I am going ;)

The return 0 below catches the same cases and it returns a useful value that causes the outwards logic to do the right thing

christf:
this now contains a fix for the crash I observed last thursday.

jkoan (approved):

PR #1599, opened 2026-07-29
finalize circleCI migration

christf:
this migrates the spellchecker and sailfish build over to gha.


.github/workflows/build.yml
```diff
@@ -83,39 +83,48 @@ jobs:
...
+      image: debian:latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+      - name: Install doxygen and essentials
+        run: apt-get update && apt-get -y install doxygen ca-certificates git rsync
+      - name: Run doxygen
+        run: cd navit && doxygen
+      - name: Deploy to GitHub Pages
```
jkoan, line 99: My intention was to do it like in my example, so it doesn't pollute the git repo any more 

https://github.com/jkoan/fdroid-repo/blob/main/.github%2Fworkflows%2Fupdate_repo.yaml#L62-L82

.github/workflows/build.yml
```diff
@@ -83,39 +83,48 @@ jobs:
...
+        uses: peaceiris/actions-gh-pages@v4
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./doc/html
+          publish_branch: gh-pages
+          cname: doxygen.navit-project.org
+  build_sailfish:
+    runs-on: ubuntu-latest
```
jkoan, line 108: This doesn't use a docker image directly. Just noticed this, as my intention was to always use a docker image directly to avoid the "platform dependency" on the provided machines. But this is probably fine for now. Just add a FIXME so we can fix it later ;)

.github/workflows/spelling.yml
```diff
@@ -0,0 +1,30 @@
+name: Spelling Check
+
+on:
+  pull_request:
+    paths:
+      - 'docs/**'
+      - '*.md'
+      - '*.rst'
```
jkoan: Should this be double star?

.github/workflows/spelling.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/checkout@v7
+      - name: Install curl
+        run: apt-get update && apt-get -y install curl
+      - name: Install misspell
+        run: |
+          curl -L -o ./install-misspell.sh https://git.io/misspell
+          sh ./install-misspell.sh -b /usr/local/bin/
+      - name: Check docs folder for misspells
+        run: |
+          misspell docs/ > typos.txt
```
jkoan: In line 6-8 you specifically added all files which are not scanned here

CMakeLists.txt
```diff
@@ -241,6 +241,20 @@ if(PKG_CONFIG_FOUND)
...
+        # pkg_check_modules found the host libcurl, not the Android one.
+        # Require an explicit cross-compiled libcurl for the target ABI.
+        find_library(CURL_LIBRARY curl)
+        if(CURL_LIBRARY)
+            set(LIBCURL_FOUND TRUE)
+            set(LIBCURL_LIBRARIES ${CURL_LIBRARY})
+        else()
+            set(LIBCURL_FOUND FALSE)
+        endif()
+    endif()
```
jkoan: Why is this relevant in this PR?

contrib/sailfish/navit-sailfish.spec
```diff
@@ -15,15 +15,36 @@ License: GPL
...
 BuildRequires: gettext-devel
+%endif
 #BuildRequires: freetype-devel
 BuildRequires: zlib-devel
+%if 0%{?suse_version} >= 1500
+BuildRequires: libxslt
+%endif
+%if 0%{?fedora}
 BuildRequires: libxslt
+%endif
```
jkoan, line 47: Why add all the if's when the requirement is all the same?

jkoan (changes requested):

christf:
I am not sure why I see "build_all - Expected — Waiting for status to be reported"

contrib/sailfish/navit-sailfish.spec
christf: because these requirements are not part of every distribution. We build on ubuntu-latest. there many of these do not exist. The alternative is to remove them entirely, I think.

CMakeLists.txt
christf: Ah, I had this in one of my upcoming branches and needed to fix the android build after that. You are right, of course, I should just get rid of this commit.

.github/workflows/build.yml
christf: :-D

.github/workflows/spelling.yml
christf: frankly. I Wasn'T sure about the spelling workflow at all. maybe just remove it entirely?

jkoan: Agreed, just remove the whole workflow :)

contrib/sailfish/navit-sailfish.spec
jkoan: But sailfish is build inside a VM inside the sailfish sdk. So it shouldn't be a problem, right?

.github/workflows/spelling.yml
christf: removed

contrib/sailfish/navit-sailfish.spec
christf: somehow I couldn't simply add, for example espeak-devel on ubuntu-latest and that failed the build. Maybe I am doing something fundamentally wrong (or I need to to add one more layer of *inception* to this)

.github/workflows/build.yml
christf: I have to say, I do not even know if/where we deploy to gh pages. In any case, doxygen is as much sailfish as "fix libcurl for android". Do you agree to remove for now?

Or should we work this out? In this case I'D need to understand what you meant by "pollute the git repo" and how that other approach avoids it - which I currently do not.

Either way, this PR is not really urgent, I just felt that maybe we should finalize this because getting rid of old stuff feels good and reduces mental load.

christf: ah, you meant via scripts/update_doxygen.sh?

jkoan: Currently the deployment works by pushing html files to the branch pages branch on this repo. But GitHub added a new option. With it you can build and upload a artifact of the webroot which will be extracted on deploy. 

So build_doxygen.sh can be changed to only build a directory or tar archive like in my example. Then the artifact will be uploaded and another step in the ci will deploy the previously uploaded artifact

jkoan: > In any case, doxygen is as much sailfish as \"fix libcurl for android\". Do you agree to remove for now?

From the title of the PR this is about moving the remaining bits from CircleCI to GHA so for me doxygen is as much in skope as sailfish is. But libcurl would only in skope if it's a fix which is needed for the building to move to gha (like when there are different dependencys installed in one CI)

jkoan:
> I am not sure why I see "build_all - Expected — Waiting for status to be reported"

Because circleci is currently a mandatory check in github settings. This is not a blocker, i will remove this after merging this PR

jkoan:
You are removing the spellcheck without adding one to GitHub actions. Have I missed that we already have one in trunk or got it missing in the process?

christf:
argh. I apparently misunderstood. I actually thought you were being serious when you said, let's remove it.

jkoan:
> argh. I apparently misunderstood. I actually thought you were being serious when you said, let's remove it.

Sorry, my bad, I didn't remembered that we already discussed this, now I found or discussion about this. But let's discuss this again, what's your opinion on the spellcheck? I am fine with either. If the rest is ready I would also be fine merging the rest and leave a new (better) spellcheck as a Todo. What do you think?

christf:
> > argh. I apparently misunderstood. I actually thought you were being serious when you said, let's remove it.
> 
> Sorry, my bad, I didn't remembered that we already discussed this, now I found or discussion about this. But let's discuss this again, what's your opinion on the spellcheck? I am fine with either. If the rest is ready I would also be fine merging the rest and leave a new (better) spellcheck as a Todo. What do you think?

Judging by the amount of typos in the docs, it seems helpful to have one O:-)
I think, I pushed to see how the pipeline likes it while you commented - let me finish up, squash and then I think I am ready.

jkoan (approved):

jkoan:
There's currently a merge conflict. Except this all the changes look good 👍

PR #1600, opened 2026-07-29
Flatpak manifest added

### LINKED ISSUE #1468: Flatpak
Can you please package this as a flatpak for compatibility with most distros?

The packaging options provided do not, for instance, work on Alpine-based postmarketOS...

Thanks.

christf:
this provides a flatpak manifest. The flatpak can be built and run with

```
flatpak-builder --force-clean --user --install-deps-from=flathub --repo=repo --install builddir org.flatpak.navit.yml 
flatpak run org.flatpak.navit
```

- [ ] embed into one of the pipelines, so this falls out of a PR run and on every release
- [ ] test audio output
- [ ] test location via geoclue

this closes #1468

### DIFF
--- a/contrib/org.flatpak.navit.yml
+++ b/contrib/org.flatpak.navit.yml
@@ -1,0 +1,24 @@
+id: org.flatpak.navit
+runtime: org.gnome.Platform
+runtime-version: '48'
+sdk: org.gnome.Sdk
+command: navit
+finish-args:
+  - --socket=x11
+  - --socket=wayland
+  - --share=network
+  - --socket=pulseaudio
+  - --device=dri
+  - --filesystem=home
+modules:
+  - name: navit
+    buildsystem: cmake
+    builddir: true
+    config-opts:
+      - -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      - -DBUILD_MAPTOOL=OFF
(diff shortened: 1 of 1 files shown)


jkoan:
Thoughts:
- Which version of navit should be the definitive version to provide as a flatpak. Flatpaks are used on mobile Linux as well, so it needs to work well on mobile. But desktop as well. 
- does it already have access to location? I know there's a portal of sorts already, probably geoclue would be most appropriate for this, but I don't know 
- what should be the goal of the flatpak? A navit which works out of the box, or just the same as currently but as a flatpak

christf:
so far, I have only tried this on my desktop and not on my phone - indeed you are right, this absolutely needs to work too. And I reckon I am still lacking a lot of stuff. Let'S make a list of things to test (top post):

The goal I had in mind: Having an easy distribution mechanism to all distributions. This allows anyone to run a current navit on, say, debian stable.


Most projects I am aware of, there are two flatpaks:
* one for every release
* one on every merge

jkoan:
> Most projects I am aware of, there are two flatpaks:
> 
>     * one for every release
> 
>     * one on every merge

That would be the same as I am planning for the fdroid repo, ein release and one dev/nightly variant

PR #1603, opened 2026-08-03
Docs: Migrate Configuration(old) page from old wiki into new structure

christf:


docs/user/configuration/basic/layout.rst
```diff
@@ -72,5 +121,28 @@ To use a layer in multiple layouts, it can be referenced using the **ref** attri
...
+Fonts
+-----
+The default font used for text on the map is defined by the ``font``
+attribute of the ``layout`` tag. It is important to pick a font that your
+OS actually supports; on Linux systems ``fc-list`` lists the available
+fonts. When specifying a font, ignore the ``style=`` part and use just the
+font name.
+
 [[Category:Customizing]]
 [[Category:Configuration]]
```
jkoan: Can be removed, it's mediawiki syntax

docs/user/configuration/basic/vehicle.rst
jkoan: Here as well

docs/user/configuration/basic/layout.rst
```diff
@@ -32,6 +32,35 @@ Here the available options:
...
+      <polygon color="#c0c0c0" />
+  </itemgra>
+
+The poly\ **gon** color defines the color with which the polygon is filled;
+it only applies to polygons such as water, towns or woods. The poly\ **line**
+color defines the color with which lines are drawn. If the item is a line,
+such as a street, it is its color; if the item is a polygon, it is its
+border color.
+
 For infos about map icons, see [[Icons]]
```
jkoan: Here's another mediawiki reference. But the read the docs search can't find anything related to icons, so we don't seem to have any documentation for it :( so we can probably remove it to not confuse readers

jkoan (changes requested):

docs/user/configuration/basic/layout.rst
christf: there is user/configuration/Icons.rst or https://navit--1603.org.readthedocs.build/en/1603/user/configuration/Icons.html

Let me add it to the scope of this PR and then update the references.

christf:
Again, there is still fixup commits here, which I am pushing so you can see just the diff of what I changed.

docs/user/configuration/basic/layout.rst
jkoan: Thanks for including it. Strange that read the docs couldn't find it (or there was an error on my side). But now it's more cohesive 👍

jkoan (approved):

christf:
rebased to fix commit structure.

PR #1604, opened 2026-08-04
Refactor:maptool:Replace hand-written OSM parsers with libosmium

pgrandin:
> **Disclaimer: an AI model wrote this PR.** Claude Opus 5 (Anthropic) wrote the libosmium frontend, the manpage and documentation updates, the commit message and this description. Every number below comes from a real run on my hardware, not from an estimate by the model. Please review it as you review any other patch from a new contributor.

This change replaces the four hand-written OSM parsers of maptool (`osm_xml.c`, `osm_o5m.c`, `osm_protobuf.c` and `osm_protobufdb.c`) with one libosmium handler. The handler calls the same `osm_add_*` callbacks in `osm.c`. The code after the parse stage does not change. The commit message has the details.

The change adds 270 lines and deletes 2002 lines. It also deletes the protobuf-c dependency and the two bundled `.proto` files.

### What to look at

* `navit/maptool/osm_libosmium.cpp` is the complete new parser. It has about 180 lines.
* This change removes the options `-B` and `-O`. These options only used `osm_protobufdb.c`, which has no replacement.
* The options `-P` and `-M` now give the format for standard input only. With `-i`, maptool detects the format from the file name.
* maptool needs a C++14 compiler. `DISABLE_CXX` therefore has no effect if `BUILD_MAPTOOL` is on.
* The build prefers the system copies of libosmium and protozero. It downloads them only if they are absent. A distribution build and an offline build therefore do not need the network.

### Verification

You cannot compare maptool output byte by byte. The threaded phases of maptool put the items of a tile in a variable order. Two runs of the same binary give maps that differ by some hundred bytes.

I compared each tile as a multiset of item type, coordinates and attributes. This comparison ignores the order. I also compared every phase counter.

* France (Geofabrik, 5.05 GB input, 524 million nodes, 3.47 GB map): all 114,981 members are identical. All counters are identical. Both maps index the same five countries.
* Denver metro: all 330 members are identical.
* These input paths also give correct results: XML with `-i`, XML from standard input, PBF with `-i`, PBF from standard input, `.osm.gz` with `-i`, and `.osm.bz2` with `-i`.

The France extract is the important test. A bounding-box extract of one city has no complete `admin_level=2` relation. Such an extract does not test the country index, the boundary assembly, the multipolygons or the turn restrictions.

The parse time for France is 354 s before this change and 280 s after it. These values are the medians of three runs. The parse is about 8% of a full build, so the total build time changes very little. This change is about the deleted code, not about speed.

### Result in CI

All checks on this PR pass. `build_linux` installs `libosmium2-dev` 2.22.0 and `libprotozero-dev` 1.8.1 from Debian, reports `Enabled BUILD_MAPTOOL`, compiles `osm_libosmium.cpp` and builds the maptool target. The log has no download message, so the build uses the system headers and not the fallback. Note that the code also compiles against libosmium 2.22.0, although the fallback pins 2.20.0. The Android, Windows, WinCE and TomTom builds pass as well.

If a dependency is absent, cmake does not stop. It reports `Disabled BUILD_MAPTOOL ( libosmium needs zlib, expat, bzip2 and threads )`, which keeps the behavior of the old protobuf-c check.

I did not test real o5m data, because I had no o5m file.


docs/development/targets/Debian_dependencies.rst
```diff
@@ -31,7 +31,10 @@ Translations for the user interface
 Maptool
 -------
 
-**protobuf-c-compiler libprotobuf-c-dev**
+**libosmium2-dev libprotozero-dev**
+
+If these packages are not available, the build downloads them. Both libraries
```
christf, line 36: the build downloads the packages? How? Where?

christf:
I think the main topic for this one is going to be, whether we want the language switch to c++.

As to the content itself, I think it is great to reduce size of code base but I am not a big fan of duplicating an api between C and C++ . Maybe the common elements should be extracted from maptool.h and then included where needed to avoid drift. Also I think checking for un-closed resources could be worthwhile, I have not gone to that depth yet because the language topic is a potential killer.

jkoan:
> I am not a big fan of duplicating an api between C and C++

Can you elaborate on this? Do you mean the map_collect_data_osmium to make it callable from the rest of the c code? I personally don't see an issue with either this wrapper function or the fact that maptool would be cpp, i even link the idea to move as much as possible to cpp going forward

docs/development/targets/Debian_dependencies.rst
jkoan: Its here: https://github.com/navit-gps/navit/blob/8d1195a6cc979d71f0720efb2cbca0058a307c09/navit/maptool/CMakeLists.txt#L13-L28

man/maptool.1
```diff
@@ -2,19 +2,28 @@
 .SH NAME
 maptool \- parse osm textfile and convert to Navit binfile format
 .SH SYNOPSIS
-.B For OSM XML data:
-.B bzcat planet.osm.bz2 | maptool mymap.bin
-[\-h] [\-6] [\-a <level>] [\-c] -[\-d <connect string]
-[\-e <phase>] [\-i <file>] [\-k] [\-M] [\-N] [\-o] [\-r <file>] [\-s <phase>]
+.B maptool \-i planet.osm.pbf planet.bin
+[\-h] [\-6] [\-a <level>] [\-c] [\-d <connect string>]
+[\-e <phase>] [\-k] [\-M] [\-N] [\-o] [\-P] [\-r <file>] [\-s <phase>]
```
jkoan: Why remove `-i`?

navit/maptool/CMakeLists.txt
```diff
@@ -1,24 +1,40 @@
...
+		# not the libraries' own CMake projects.
+		include(FetchContent)
+		FetchContent_Declare(libosmium
+			URL https://github.com/osmcode/libosmium/archive/refs/tags/v2.20.0.tar.gz
+			SOURCE_SUBDIR non-existent)
+		FetchContent_Declare(protozero
+			URL https://github.com/mapbox/protozero/archive/refs/tags/v1.7.1.tar.gz
+			SOURCE_SUBDIR non-existent)
+		FetchContent_MakeAvailable(libosmium protozero)
+		include_directories(${libosmium_SOURCE_DIR}/include ${protozero_SOURCE_DIR}/include)
```
jkoan: I am not sure if i like to have that or not. Because it adds another point where versions need to be updated which should be provided by the distribution. On the other side, this i a header only lib that we can actually bring. Maybe therefore a "module" under support would be appropriate, so it fits with all the support libs

navit/maptool/CMakeLists.txt
```diff
@@ -1,24 +1,40 @@
...
-
-	if(NOT MSVC)
-		PROTOBUF_C_GENERATE_C (PROTO_SRCS PROTO_HDRS osmformat.proto)
-		PROTOBUF_C_GENERATE_C (PROTO_SRCS PROTO_HDRS fileformat.proto)
-		target_sources(maptool_core PRIVATE osm_protobuf.c osm_protobufdb.c
-			fileformat.pb-c.c osmformat.pb-c.c)
-		target_link_libraries(maptool_core ${PROTOBUF_C_LIBRARY})
-	endif(NOT MSVC)
+		itembin_buffer.c itembin_slicer.c misc.c osm.c osm_psql.c
+		osm_relations.c sourcesink.c tempfile.c tile.c zip.c osm_libosmium.cpp)
```
jkoan: at some point in time this was sorted :D

navit/maptool/osm_libosmium.cpp
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA  02110-1301, USA.
+ */
+
+/*
+ * libosmium-based OSM input frontend. Replaces the hand-rolled XML, o5m
+ * and protobuf parsers: reads any libosmium-supported format (osm, osm.gz,
```
jkoan: Comments should document the current state, not be a history lessen what was there some time in the past

navit/maptool/osm_libosmium.cpp
```diff
@@ -0,0 +1,174 @@
...
+void sig_alrm_end(void);
+extern int processed_nodes, processed_ways, processed_relations;
+
+int map_collect_data_osmium(const char *filename, const char *format, struct maptool_osm *osm);
+}
+
+namespace {
+
+/* matches BUFFER_SIZE in maptool.h; the old XML parser had the same limit */
+const int buffer_size = 1280;
```
jkoan: Again history lesson, and why not use the define from there to keep them in sync?

PR #1608, opened 2026-08-09
another round of stability / memory fixes

christf:
this ensures, maptool adheres to "keep tempfiles", and cleans up double free on -t.

And this prevents value overflows when routing.


christf:
sorry, that merge was an accident

jkoan (approved):

PR #1613, opened 2026-08-17
fix(graphics): suppress duplicate labels on screen

### LINKED ISSUE #1383: map rendering: labels duplicate
in navit, I just saw triple labels, that clutter the rendered map. I am not sure if it is the layout. Documenting it for now to hopefully come back to it later.

Image

christf:
A street name that is also carried by its mapped sidewalks, cycleways and footways was rendered once per way, cluttering the map with repetitions of the same text. The first of several identical labels that are close together on screen is now drawn once: a label is a duplicate if its rotated bounding box overlaps one already placed, or if its anchor is within half a label width (at least two line heights) of one already placed. A placement with a larger text size never suppresses a smaller duplicate of the same text. The same rotated footprint is used to cull off-screen labels and covers multiline labels drawn next to circles (most POI, town, house-number and highway-exit labels), so street and point labels share one definition of what a label covers.

The registry is per frame: it is reset when a frame begins (graphics_draw_mode draw_mode_begin) and freed with the graphics instance. Drawing is single-threaded by design, so no locking is needed.

this fixes #1383


christf:
reverting to draft because, the layout is an improvement but I am not happy with it yet.

christf:
nodup

navit/graphics.c
```diff
@@ -98,6 +98,10 @@ struct graphics {
     GHashTable *image_cache_hash;
     /* for dpi compensation */
     int dpi_factor;
+    /* Labels placed in the current frame, to suppress duplicate labels (label text -> list of placements).
+     * This is per-frame state: it is reset at the start of each frame (graphics_draw_mode begin) and freed in
```
jkoan, line 102: Why are the start instead of at the end? On low end hardware this ram could be used between frames for other things like traffic

navit/graphics.c
```diff
@@ -568,6 +572,8 @@ void graphics_font_destroy(struct graphics_font *gra_font) {
  * @returns nothing
  * @author David Tegze (02/2011)
  */
+static void graphics_labels_reset(struct graphics *gra);
```
jkoan: Here you're slipping the comment and the function below

navit/graphics.c
```diff
@@ -1012,6 +1019,8 @@ struct graphics_image *graphics_image_new(struct graphics *gra, char *path) {
  * @author Martin Schaller (04/2008)
  */
 void graphics_image_free(struct graphics *gra, struct graphics_image *img) {
+    (void)gra;
+    (void)img;
```
jkoan: Is this to shut up the compiler about unused?

jkoan (changes requested):

navit/graphics.c
christf: resetting at the end looks really annoying/scary for cases when rendering is aborted. I think this is the simpler implementation that should always work - and I didn't think labels would be consuming that much memory anyway. Do you think the added complexity is worth it for a few kb?

navit/graphics.c
christf: whoopsie! sorry for this.

christf:
pushed a fixup commit for easy review. will squash if that receives you blessing

jkoan (approved):

jkoan:
Ready to rebase

christf:
done.

PR #1617, opened 2026-08-26
fix Android Map Downloader

### LINKED ISSUE #1112: Map download at Planet Extractor not possible, almost same errors as in issue #1072

### LINKED ISSUE #1132: Can't download map using Navit menu on Android

### LINKED ISSUE #1311: Navit won't download maps on android

hoehnp:
Unlike the previous approach with curl, I went in this PR the simple way and updated the android java code to at least fix the issue for the so far only supported platform: android.


navit/android/src/org/navitproject/navit/NavitDownloadSelectMapActivity.java
```diff
@@ -49,13 +56,29 @@ public class NavitDownloadSelectMapActivity extends ExpandableListActivity {
...
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+        try {
+            updateGithubMetaData();
+        } catch (InterruptedException e) {
+            Log.e(TAG, "We failed to download the github api file.");
+            e.printStackTrace();
+            throw new RuntimeException("failed to download github api file.");
```
christf, line 71: usually I try to either handle an exception or rethrow it. logging and rethrowing makes for confusing traces. Not severe, but enough to make me comment ;)

navit/android/src/org/navitproject/navit/NavitDownloadSelectMapActivity.java
```diff
@@ -49,13 +56,29 @@ public class NavitDownloadSelectMapActivity extends ExpandableListActivity {
...
+        }
+
+        while (true) {
+            if (githubMetadata != "") break;
```
christf, line 75: I cannot try this code but this looks like busy waiting. maybe a sleep is a bit better here?

jkoan:
just to note it down, using the Github api directly has a big downside: 
Its rate limited.

So as a temporary solution this is fine, but the goal should be to define a own xml/json Format for https://github.com/navit-gps/gh-actions-mapserver/ and use that in the future

navit/android/src/org/navitproject/navit/NavitDownloadSelectMapActivity.java
jkoan: sleep wouldn't solve the Problem. When running into the github api limit the app will just block as the api will never respond with something valid

navit/android/src/org/navitproject/navit/NavitMapDownloader.java
```diff
@@ -679,13 +656,99 @@ private boolean downloadData(URLConnection c, long alreadyRead, long realSizeByt
         return success;
     }
 
-    private URL getDownloadURL() {
+    private String getLatestDate() {
+        URL url;
+        try {
+            url = new URL("https://api.github.com/repositories/384098365/releases/latest");
```
jkoan: there are 3 Locations where the github api is read, but only one request would be necessary as the one request to  https://api.github.com/repositories/384098365/releases/latest contains everything what's needed. in the response the assets tag contains all assets and there sizes. Also the json is never parsed as json but only as strings, this is really error prone

jkoan (changes requested):

navit/android/src/org/navitproject/navit/NavitDownloadSelectMapActivity.java
hoehnp: the problem here is that I need to do the downloading asynchronously, since otherwise the emulator throws an error. However, since the downloaded file is required for the functions afterwards, I added this code to ensure the file was downloaded .

hoehnp:
> just to note it down, using the Github api directly has a big downside: Its rate limited.
> 
> So as a temporary solution this is fine, but the goal should be to define a own xml/json Format for https://github.com/navit-gps/gh-actions-mapserver/ and use that in the future

certainly an own format could be better. But I would suggest to use the API for now to be able to fix the downloader for the time being.

navit/android/src/org/navitproject/navit/NavitMapDownloader.java
hoehnp: I will try to reduce the number of mentions. What I could do is to download the file once in the constructor of NavitMapDownloader and then use it internally.

I didnt not find any library in the Java Standard Library for JSON parsing.. I just saw that there is one in the Android API, I will give it a shot.

